PR #2, opened 2020-03-28
Add global styles + base for form page

notfelineit:
# Add Global Styles & Form
This PR:
- adds global styles based on [Juan's css file ](https://recidiviz.slack.com/archives/G010VACT8F2/p1585333687046500) 
- adds a base for the [form page 
](https://www.figma.com/proto/v43UQeEd1Xkb3w2NgxcWql/%F0%9F%9A%80-Recidiviz?node-id=288%3A4&viewport=183%2C343%2C0.09931680560112&scaling=scale-down-width)
- adds style examples to the `/test` page for each css style

## Styles
The styles are the following:
Screen Shot 2020-03-27 at 9 48 22 PM

We can update them as we build out the real pages! 😄

### DIFF
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -89,7 +89,7 @@ export const GlobalStyles = createGlobalStyle`
     height: 24px;
   }
 
-  .ic_share {
+  .ic-share {
     width: 14px;
     height: 14x;
   }
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -1,0 +1,101 @@
+import { createGlobalStyle } from "styled-components";
+
+export const GlobalStyles = createGlobalStyle`
+  @import url('https://fonts.googleapis.com/css?family=Poppins&display=swap');
+  @import url('https://fonts.googleapis.com/css?family=Libre+Baskerville&display=swap');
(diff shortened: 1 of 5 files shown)


src/styles.ts
```diff
@@ -0,0 +1,101 @@
+import { createGlobalStyle } from "styled-components";
```
szhu: can we name all our files .tsx? This avoids situations where we rename a file .ts -> .tsx after we realize we need to put jsx in it.

also, the linters that are set up are currently only checking .tsx files. alternatively we can just add .ts  patterns.

notfelineit: Sure thing!

src/styles.tsx
```diff
@@ -0,0 +1,101 @@
+import { createGlobalStyle } from "styled-components";
+
+export const GlobalStyles = createGlobalStyle`
+  @import url('https://fonts.googleapis.com/css?family=Poppins&display=swap');
+  @import url('https://fonts.googleapis.com/css?family=Libre+Baskerville&display=swap');
+  @import url('https://fonts.googleapis.com/css?family=Rubik&display=swap');
```
szhu: Creating a PR to remove this because we realized it didn't work, and because styled-components gives a warning about not doing this.

PR #3, opened 2020-03-28
Add form page and development scripts

notfelineit:
This PR:
- Adds two development scrips: `yarn dev` and `yarn clean` as shortcuts to running the app locally, and cleaning up builds
- Updates the form page under `/contribute`
- Adds `ContainerTint1` as a base for a page container w/ background color taken from the Figma mock


package.json
```diff
@@ -1,5 +1,9 @@
 {
   "private": true,
+  "scripts": {
+    "clean": "rm -rf dist && rm -rf .cache",
+    "dev": "bin/parcel serve index.html"
```
jessex: Let's update the readme to call out at least this new command, instead of the manual call to `bin/parcel`.

notfelineit: Will do! It's still in progress 😄 Was going to finish the form first, but maybe I'll just push this up now.

index.html
```diff
@@ -2,8 +2,13 @@
   <head>
     <title>TITLE GOES HERE</title>
   </head>
+
   <body>
     <div id="app"></div>
     <script src="src/entry-point/index.tsx"></script>
+    <link
```
szhu: nit: put stylesheets under <head>!

notfelineit: hey - please clarify by "under" 😄

szhu: omg… bad github formatting. I meant to write:

> nit: put stylesheets under `<head>`!

package.json
szhu: For `package.json` scripts, they automatically look inside `node_modules/.bin`, so we can just do:

```suggestion
    "dev": "parcel serve index.html"
```

src/page-form/components/TextInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+  placeholder: PropTypes.string,
+  value: PropTypes.string,
+  onChange: PropTypes.func,
+  defaultValue: PropTypes.string,
+  type: PropTypes.string,
+};
+
+TextInput.defaultProps = {
+  type: "text",
+};
```
szhu: I think PropTypes is largely redundant with TypeScript props interfaces -- we can probably just not use PropTypes?

As for defaultProps, we can do `type={props.type || "text"}` above. Or leave it, since I think having defaults can sometimes hide complexity in subtle ways.

szhu: Oh, I see that the current eslint config enforces it.

We can remove that rule from eslint! Doing so in https://github.com/Recidiviz/covid19-dashboard/pull/5#discussion_r399651406

Let me know if you actually wanted to continue using prop types.

szhu: Feel free to just 

```
git fetch && git checkout origin/szhu/news-feed .eslintrc.json
```

if you want it disabled before #5 is merged.

index.html
szhu: Just noticed something. I think between this `<link>` and the following in `styles.tsx`

```
  @import url('https://fonts.googleapis.com/css?family=Poppins&display=swap');
  @import url('https://fonts.googleapis.com/css?family=Libre+Baskerville&display=swap');
  @import url('https://fonts.googleapis.com/css?family=Rubik&display=swap');
```

we only need to use one of them!

notfelineit: I thought so too, which is why the previous PR was just shipped with the css imports - but I noticed it wasn't actually using the right fonts. After adding it here, the proper fonts are loading 😄

notfelineit:
Progress so far:
Screen Shot 2020-03-28 at 5 29 35 PM

package.json
```diff
@@ -1,6 +1,13 @@
 {
   "private": true,
-  "dependencies": {},
+  "scripts": {
+    "clean": "rm -rf dist && rm -rf .cache",
+    "dev": "bin/parcel serve index.html"
+  },
+  "dependencies": {
+    "prop-types": "^15.7.2",
```
szhu: This can be removed now!

package.json
```diff
@@ -1,6 +1,13 @@
...
+    "us-zcta-counties": "^0.0.2"
```
szhu: Can you move this to `devDependencies`? I don't have a strong understanding of dependencies vs devDependencies, but it looks like everything works even when everything is a devDependency, so I'm just trying to keep everything together for now. I'm also down to move everything to dependencies if that's better.

Also -- I've been using exact version numbers since we're not committing a lockfile. Maybe:

```suggestion
    "us-zcta-counties": "0.0.2"
```

src/page-form/components/TextInput.tsx
```diff
@@ -0,0 +1,50 @@
+import PropTypes from "prop-types";
```
szhu: This can be removed now! interesting that eslint didn't catch this 🤔

package.json
notfelineit: Whoo! Thanks.

package.json
notfelineit: Ah, it's needed at runtime and not just in development, so we should keep it in `dependencies`. I can update it to use the exact version :)

src/page-form/components/TextInput.tsx
notfelineit: Good catch!

PR #4, opened 2020-03-28
Add static rendering and window title support

szhu:
After this PR, running `yarn build` will create pages that can be directly deployed to GitHub Pages or other static hosts.

This is what its output looks like:
image

In particular:
- The initial view of pages is rendered statically
- Pages have titles now (both in dev and when built)
- On the static server, when visiting a page that's not `/`, refreshing the page no longer gives a 404.


src/entry-point-static/build.tsx
```diff
@@ -0,0 +1,24 @@
...
+import path from "path";
+
+import generatePageContent, { PageInfo } from "./generatePageContent";
+
+let pageInfos: PageInfo[] = [
+  { title: "Recidiviz COVID-19 Dashboard", location: "/" },
+  { title: "Test Page", location: "/test" },
+];
+
+console.log();
```
notfelineit: I think you forgot about this one?

szhu: thanks for the callout!

this line is intentional -- it prints the blank line highlighted here, just to create some space between the command outputs:

image

It does feel weird to put it here though. Moving to a more reasonable place:

In `package.json`:

```diff
    "scripts": {
-     "build": "parcel build index.html &&         parcel build --target node src/entry-point-static/build.tsx &&         node dist/build.js && rm -f dist/build.js dist/build.js.map"
+     "build": "parcel build index.html && echo && parcel build --target node src/entry-point-static/build.tsx && echo && node dist/build.js && rm -f dist/build.js dist/build.js.map"
    }
```

src/entry-point-static/build.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+let pageInfos: PageInfo[] = [
+  { title: "Recidiviz COVID-19 Dashboard", location: "/" },
+  { title: "Test Page", location: "/test" },
+];
+
+console.log("Rendering React statically into HTML files:");
+
+for (let pageInfo of pageInfos) {
+  let dirPath = path.join("dist", "." + pageInfo.location);
```
jessex: Small question: why the concatenation instead of just another comma to separate the period and `pageInfo.location` into separate args?

src/entry-point-static/generatePageContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+  let contentHtml = ReactDOMServer.renderToString(
+    styleSheet.collectStyles(
+      <StaticRouter location={location}>
+        <App />
+      </StaticRouter>,
+    ),
+  );
+  const stylesHtml = styleSheet.getStyleTags();
+  let template = fs.readFileSync("dist/index.html", "utf-8");
+  template = template.replace("TITLE GOES HERE", title);
```
jessex: It might just be me having never seen an app boostrapped quite like this before, but I didn't even realize that these three tokens being replaced here were tokens when I first saw them earlier in the PR. I assumed they were placeholders. It would be worth adding a comment next to them to ensure no one accidentally removes them and breaks this process in confusion.

src/entry-point-static/build.tsx
szhu: Thanks for the callout - that works!

```js
> const path = require("path")
> path.join("dist", "/test")
'dist/test'
```

I was thinking of Python's `os.path.join`, which behaves differently:

```py
>>> import path from os
>>> path.join("dist", "/test")
'/test'
```

src/entry-point-static/generatePageContent.tsx
szhu: Good point -- addressed!

index.html
```diff
@@ -1,10 +1,17 @@
+<!--
+  generatePageContent.tsx looks for the following strings in this file:
+  - "TITLE PLACEHOLDER"
+  - "<style></style>"
+  - "CONTENT PLACEHOLDER"
+  If you change any fo these, make sure to edit generatePageContent.tsx as well.
```
jessex: Nit: fo --> of

package.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {},
   "devDependencies": {
     "@hot-loader/react-dom": "16.13.0",
+    "@types/chalk": "2.2.0",
```
szhu: Oops, we didn't need this -- `chalk` comes with its own types.

szhu: Created #10 to remove this.

szhu:
Some of my later commits broke the static rendering system and we didn't notice:
- The templates stopped being substituted in at build time
- If we generate `/index.html` first, it breaks because we read `index.html` as a template for generating subsequent pages.

Fixed with this diff:

```diff
diff --git a/src/entry-point-static/build.tsx b/src/entry-point-static/build.tsx
index 6594f66..93f9dc0 100644
--- a/src/entry-point-static/build.tsx
+++ b/src/entry-point-static/build.tsx
@@ -12,13 +12,14 @@ let pageInfos: PageInfo[] = [
 
 console.log(chalk.blueBright("Rendering React statically into HTML files:"));
 
+let template = fs.readFileSync("dist/index.html", "utf-8");
 for (let pageInfo of pageInfos) {
   let dirPath = path.join("dist", pageInfo.location);
   let filePath = path.join(dirPath, "index.html");
 
   console.log(chalk.blueBright(`Writing: ${filePath}`));
 
-  let html = generatePageContent(pageInfo);
+  let html = generatePageContent(template, pageInfo);
   fs.mkdirSync(dirPath, { recursive: true });
   fs.writeFileSync(filePath, html);
 }
diff --git a/src/entry-point-static/generatePageContent.tsx b/src/entry-point-static/generatePageContent.tsx
index 5bddc52..5abcf89 100644
--- a/src/entry-point-static/generatePageContent.tsx
+++ b/src/entry-point-static/generatePageContent.tsx
@@ -13,7 +13,10 @@ export interface PageInfo {
   location: string;
 }
 
-export default function generatePageContent({ title, location }: PageInfo) {
+export default function generatePageContent(
+  template: string,
+  { title, location }: PageInfo,
+) {
   let styleSheet = new ServerStyleSheet();
   let contentHtml = ReactDOMServer.renderToString(
     styleSheet.collectStyles(
@@ -23,7 +26,6 @@ export default function generatePageContent({ title, location }: PageInfo) {
     ),
   );
   const stylesHtml = styleSheet.getStyleTags();
-  let template = fs.readFileSync("dist/index.html", "utf-8");
 
   [
     ["TITLE PLACEHOLDER", title],
@@ -38,7 +40,7 @@ export default function generatePageContent({ title, location }: PageInfo) {
       );
       process.exit(1);
     }
-    template.replace(toFind, toReplace);
+    template = template.replace(toFind, toReplace);
   });
   return template;
 }
```

This diff is part of #9 and will be fixed when that merges.

PR #5, opened 2020-03-28
Add NewsFeed component

szhu:
Figma:

> image

Actual (visible at `/test`):

> image

Also sets up:
- NewsFeedEvent interface
- Fake NewsFeedEvents
- An example of how we could consume TSV data

Not implemented yet:
- If no actions have been taken the past week, call out 2 things:
    - “[state] has taken no actions in the past week”
    -+ Add data → link to add data form for [state].
- There should be a When an action is “verified,” a source should be a link to a report, press release, or article so people can read more details about the specific action.


.eslintrc.json
```diff
@@ -45,6 +45,10 @@
     // very noisy with Prettier violations that get autocorrected anyway.
     "prettier/prettier": "off",
 
+    // We use TypeScript props interfaces, which is mostly redundent with prop
+    // types.
+    "react/prop-types": "off",
```
szhu: Turning this off bc I don't think we need it. See also https://github.com/Recidiviz/covid19-dashboard/pull/3#discussion_r399644074

szhu:
Modified it to pull from the real events, now that we have the TSV link. Now it looks like:

> image

.eslintrc.json
notfelineit: Thank you - I don't normally use them when using typescript, but saw that hey were enabled on eslint.

src/news-feed-events/NewsFeedEvent.tsx
```diff
@@ -0,0 +1,84 @@
...
+  "# OF PEOPLE RELEASED": string;
+  "# OF PEOPLE PREVENTED FROM ENTERING": string;
+  "SUMMARY RE-WRITE": string;
+}
+
+/**
+ * Convert a TSV spreadsheet into NewsFeedEvent objects.
+ */
+export function parseNewsFeedEventsFromTsv(tsv: string) {
+  let table = tsv.split("\n").map((line) => line.split("\t"));
```
jessex: I think something is simple is okay to get started, but I'd feel more comfortable using an actual csv/tsv library, which is going to have at least rudimentary support for escaping, character encodings, etc.

src/news-feed-events/NewsFeedEvent.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const Headers = table[0] as (keyof NewsFeedEvent)[];
+  return table.slice(1).map((row) => {
+    let record = {} as NewsFeedEvent;
+    for (let i = 0; i < Headers.length; i++) {
+      let header = Headers[i];
+      let cell = row[i];
+      record[header] = cell;
+      1;
```
jessex: Sorry for the ignorance: what is this line doing?

src/news-feed-events/NewsFeedEvent.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export const FakeNewsFeedEvents = parseNewsFeedEventsFromTsv(
+  FakeNewsFeedEventTsv,
+);
+
+// Real events
+
+// https://mnaoumov.wordpress.com/2018/02/05/download-google-drive-file-via-ajax-ignoring-cors/
+const NewsFeedEventsTsvUrl =
+  "https://cors-anywhere.herokuapp.com/https://docs.google.com/spreadsheets/d/e/2PACX-1vRccQtwTk0FiM3PxtBF5kF-O0RcdzOWyxZgolquqnNmReAEV7bmMSJE7PL_MT8HiQAEynhUJZrC0i6T/pub?gid=251892683&single=true&output=tsv";
```
jessex: Nit: especially if we are going to have more than one data source, but even if we only have one, can we extract their URLs out to a shared location somewhere so they can be shared with confidence and we won't have to worry about minor typos to and fro?

src/news-feed-events/NewsFeedEvent.tsx
szhu: No idea how that got there -- probably a typo. Thanks!

src/news-feed-events/NewsFeedEvent.tsx
szhu: This does appear brittle, didn't think about checking edge cases here.

I did a test of some edge cases just now and it looks like everything actually just works. Here's the sheet I made, which contains a tab and a newline within a cell, and some unicode characters:

image

The TSV we get looks like:

```
a<TAB><TAB>b
<TAB>c        d<TAB>
文<TAB>e f<TAB>🤠
<TAB>fdfsd<TAB>
```

The tab has been replaced by a few spaces, and the newline has been replaced with a single space. (In the future, we should probably switch to using CSV output and use a parser if we need to actually want the newlines to stay newlines.)

As for encodings, it looks like there are no problems. The TSV is UTF-8, and `fetch`'s `text()` uses UTF-8 as well:

https://developer.mozilla.org/en-US/docs/Web/API/Body/text

Based on this, I think the current implementation is fine as is. Open to feedback, of course!

src/news-feed-events/NewsFeedEvent.tsx
szhu: I agree! I don't know how it should be organized though, because I don't have a high-level overview of the data part of this project. Let me know what make the most sense. Alternatively, we can defer making this change until we have our second URL that we need, in which case the correct abstraction will become more obvious. What do you think?

src/news-feed-events/NewsFeedEvent.tsx
jessex: I appreciate this test! I think a more accurate test would be one that samples actual form input. In spite of their best intentions, I imagine the people pasting form input from one sheet into another are not going to look for things like, e.g. newlines or tabs in the form input, or quotes of all kinds. I'm fine to stick with the current approach for now, but it would be good to ping Serena on this and ask her if she can bound what kinds of form input formatting we might permit, or if everything is going to be just straight up lines of text.

src/news-feed-events/NewsFeedEvent.tsx
jessex: Sounds good!

src/news-feed-events/NewsFeedEvent.tsx
szhu: > In spite of their best intentions, I imagine the people pasting form input from one sheet into another are not going to look for things like, e.g. newlines or tabs in the form input, or quotes of all kinds.

Having these in the input should not cause issues! As tested above, I think this system can handle all kinds of characters.

> I think a more accurate test would be one that samples actual form input.

+1! I'll be careful over the next few days to see if anything breaks -- if it does, we can use a parser library to address it.

PR #8, opened 2020-03-30
Add overview page from v0

szhu:
This adds the overview page over from v1!

@macfarlandian please read these important notes:
- You can visit the page by running `bin/parcel serve index.html` and going to <http://localhost:1234/overview>
  - You can instead run `yarn dev` after  #3 is merged!
  - I just picked `/overview`, but I can change the URL if you have an opinion.
  - Once you're there, the page looks exactly the same as before!
      Screen Shot 2020-03-29 at 8 34 51 PM
- Here's where your files have gone:
   - `src/index.html` -> `src/page-overview/OverviewPage.tsx`
  - `src/style.css` -> Also `src/page-overview/OverviewPage.tsx`! (Check the top of the file)
  - The tailwind built-in styles -> `src/page-overview/assets/output.css_`
  - `src/*.js` -> `src/page-overview/assets/*.js_` (the nonstandard file extension tells Parcel to not preprocess the file)
  - `src/icons/*` -> `src/page-overview/assets/icons/*`
- Normally, in development mode, the page will instantly update, without even reloading, as soon as you save a file. Because the current JS isn't aware of the React rendering lifecycle, you'll instead get a blank page. Just reload the page to see your updates for now.
- To build the page for deployment, just run `yarn build` as usual. I tested it and I think it works perfectly.

Things that are hacky that we should eventually migrate off of:
- [Really hacky] Since we use react-router, we should migrate the JS to ~(1) not use global variables~ **update: this is done!** and (2) to be aware of the React lifecycle. Note: Currently your v0 overview page JS is run on all pages, not just the overview page!
- [Really hacky] We currently have two style systems (the styles imported from v0 by this PR and the styles added in #2) that might clash with each other. We should take a look at that. (Note: I am investigating this in a larger scale, having a conversation with Juan about what the design system of this website will be.)
- [Not a big deal] We should migrate the JS to be proper React code.
- [Not a big deal] There's a really big SVG included in the the HTML. Nothing's technically wrong, but we should move the SVG into another file and import it.


macfarlandian:
@szhu given that `output.css` is a generated file, is it necessary to track it?

szhu:
@macfarlandian Hmm… it looks like we can just a pre-compiled version.

Thanks! This PR's diff is way smaller now!

szhu:
@macfarlandian the js files are fixed! They are not complied through the Parcel pipeline rather than included verbatim. Saving JS files should no longer cause errors, but note that since they are still not aware of the React lifecycle, you'll need to reload the page to see most updates.

PR #9, opened 2020-03-30
Overview page visual polish

macfarlandian:
Screen Shot 2020-03-29 at 4 08 42 PM

This adds proper colors and typefaces to the Overview page to match the v0.5 mocks, along with some other styling to resolve conflicts with the v1 global styles, and also adds state tooltips that appear on map hover.


index.html
```diff
@@ -12,6 +12,10 @@
       rel="stylesheet"
       href="https://unpkg.com/tailwindcss@1.2.0/dist/tailwind.min.css"
     />
+    <link
+      href="https://fonts.googleapis.com/css?family=Poppins:300,400,600|Rubik&display=swap"
+      rel="stylesheet"
+    />
```
szhu: Note: This will be the third time we've tried to add support for fonts.

Attempt 1:
https://github.com/Recidiviz/covid19-dashboard/pull/2/files#diff-639894c3fb4e775b55e4069213b8fa69R4-R6

Attempt 2 (not merged yet):
https://github.com/Recidiviz/covid19-dashboard/pull/3/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR9-R12

I think this adding this is correct thought. What we should do:
1. Remove Attempt 1, since it's not working. We can do this at a later time.
2. Whoever merges the second of either this PR or the PR for Attempt 2 should make sure we're not adding the `<link>` element twice.

src/page-overview/assets/tooltip.js
```diff
@@ -0,0 +1,19 @@
...
+        .html(getStateName(event.target.id))
+        .css({
+          left: event.pageX + 15,
+          top: event.pageY + 15,
+        })
+        .removeClass("hidden");
+    })
+    .mouseout(function () {
+      $tooltip.html("").addClass("hidden");
+    });
```
szhu: Is this robust?

src/page-overview/assets/init.js
```diff
@@ -4,6 +4,7 @@ import {
   setCurrentState,
 } from "./data-forms";
 import { Slider } from "./sliders";
+import initTooltips from "./tooltip";
```
szhu: nit: Let's try to be consistent about import/exporting names!

- Can you convert the existing files to using default exports if you prefer it?
- The default export should have the same name as the file. (Note to self: I should add `eslint-plugin-filenames`'s `match-exported` rule.)

szhu: Created #10 to add the `match-exported` rule!

jessex:
I don't know if this is just me, @macfarlandian, but when I checked out your branch locally, I ran the app in two ways. First was by doing `yarn build` and then running the command to run the built files out of `/dist` as specified in the readme. I note that the map was all black except for the outlines, as well as the slider component.

Screen Shot 2020-03-30 at 10 23 31

When I simply run `parcel serve index.html`, however, everything looks as expected. Figured I'd let you know in case you haven't tried that yet locally.

index.html
macfarlandian: yeah I based this on @notfelineit 's work in #3 (had to add some additional weights though) so agreed

src/page-overview/assets/tooltip.js
macfarlandian: not sure what you mean by "robust" in this context? could you be more specific?

macfarlandian:
@jessex yeah I am also seeing some weirdness in the built version ... it was working at some point so I must have broken it. looking into it before merging

macfarlandian:
@jessex i enlisted sean to help figure this out and he pushed a fix to this branch, should be good now. merging and moving on!

jessex:
Excellent teamwork, friends

PR #13, opened 2020-03-31
update overview calculations with new model

macfarlandian:
In collaboration with Justine I have implemented a JavaScript version of our latest model for projecting hospital utilization by incarcerated populations; this now takes in a transsmission rate (R0) instead of an infection rate and outputs both a percentage and a number of days. There are some relatively small UI changes to account for that (sorry @szhu hopefully not too disruptive to the UI rewrite) but mostly just new calculation logic that runs under the hood.

Screen Shot 2020-03-30 at 6 42 13 PM

You can see that "transmission rate" wraps onto two lines, which we probably want to fix, but I left it alone here in light of what Sean is doing in parallel to rebuild the UI.

The data is still hardcoded into a local file for now, at Justine's recommendation, since she expects those numbers to be pretty stable for now. We may need to revisit that later but it seemed premature to try to wire it up to an external data source tonight.


src/infection-model/index.js
```diff
@@ -0,0 +1,122 @@
...
+  N,
+  R0,
+  dIncubation = 2,
+  dInfectious = 4.1,
+  dRecoveryMild = 9.9,
+  dHospitalLag = 2.9,
+  dHospitalRecovery = 22,
+  dHospitalFatality = 8.3,
+  pSevereCase = 0.26,
+  pFatalityRate = 0.026,
```
justkunz: For more context: These hard-coded values are what we are currently using to power the spreadsheet model currently except the "Normal Healthcare Fatality Rate", which is currently set to 1% on the variables tab but is actually closer to 2.6% [according to the CDC](https://www.cdc.gov/mmwr/volumes/69/wr/mm6912e2.htm). I've changed it in the working copy but not the live version yet.

src/infection-model/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+  const mildDelta = gamma * pMildCase * infectious - mild / dRecoveryMild;
+  const severeDelta = gamma * pSevereCase * infectious - severe / dHospitalLag;
+  const hospitalizedDelta =
+    severe / dHospitalLag - fatalitiesDelta - severeRecoveredDelta;
+
+  const infectiousDelta = alpha * exposed - gamma * infectious;
+
+  // The transition from susceptible -> exposed depend on the total infectious population
+  const totalInfectious = infectious;
```
justkunz: This may seem weird here (why assign one var to another var with no modification?) but it's a placeholder for when this method evolves to incorporate the age distribution.

src/infection-model/index.js
```diff
@@ -0,0 +1,122 @@
...
+  const mildRecovered = x[6];
+  const severeRecovered = x[7];
+  const fatalities = x[8];
+
+  // Compute the deltas from the bottom of the tree up, additions to the leaf nodes are subtracted from their parent node as people flow from the root to the leaves of the tree
+  const mildRecoveredData = mild / dRecoveryMild;
+  const fatalitiesDelta =
+    ((pFatalityRate / pSevereCase) * hospitalized) / dHospitalFatality;
+  const severeRecoveredDelta =
+    ((1 - pFatalityRate / pSevereCase) * hospitalized) / dHospitalRecovery;
```
jessex: Should it actually be `(((1 - pFatalityRate) / pSevereCase) * hospitalized) / dHospitalRecovery`? That is, instead of the math for the first component being `1 - (pFatalityRate / pSevereCase)` as it currently is, should it be `(1 - pFatalityRate) / pSevereCase`? 

Looking at it more closely, I think it actually _should_ be how you have it now, that the first component is supposed to be 1 - the first component from the previous calculation. But extra parentheses for grouping may broadcast intention more clearly? Take or leave that comment as you see fit.

src/infection-model/index.js
jessex: Might be worth adding a TODO in that case, so it's obvious where the extension point is.

src/page-overview/assets/data-forms.js
```diff
@@ -150,7 +145,7 @@ export function setCurrentState(stateCode) {
     incarceratedPopulation: pop,
     // define valid input range according to base number
     incarceratedPopulationMin: Math.round(pop * 0.5),
-    incarceratedPopulationMax: Math.round(pop * 1.5),
+    incarceratedPopulationMax: Math.round(pop * 1.1),
```
jessex: Settled on this range?

macfarlandian: yeah I confirmed this with Serena

src/infection-model/index.js
macfarlandian: when I add the extra parens Prettier auto-deletes them so ........... this is what it's gonna be

PR #16, opened 2020-04-01
Sitewide tech updates for v0 launch

szhu:
## Description of the change

**Preview link**: https://recidiviz-covid19-dashboard-1f79m8w5o.now.sh/

What's new
- After this PR, the site generated using `yarn build` generally looks presentable!

What's improved/fixed
- List of pages has been updated. Titles have been updated as well.
- You no longer have to edit two files when creating a new page. The list of pages used by dev and static are now in the same file.
- A previous merge accidentally message up `package.json` slightly. This is now fixed

What's still missing
- Technical
  - Homepage could still be more responsive. (The circle should always completely fit onto one page.)
  - The text/paragraph styling on the About and Get Involved pages looks a little off, I think
- Copy:
  - No intro on homepage
  - Copy on About and Get Involved are not finalized
  - The social media icons don't have links ye
  - The name of logo still seems a little cryptic
  - No link to Recidiviz?

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #17, opened 2020-04-01
Chart component for impact curves

macfarlandian:
## Description of the change

This builds an area chart from curve data using the [Semiotic](https://semiotic.nteract.io/) charting library, which after building this component definitely feels like a reasonable choice. We'll have to do another pass for final colors and type at some point but that shouldn't materially affect this PR so no point in waiting IMO.

Fed some sample data the chart looks like this: 

Screen Shot 2020-04-01 at 11 08 52 AM

(scripts got moved around in `package.json` in a way that might conflict with the #16 merge, i'll be sure to clean that up before merging assuming that one gets into master first)

## Type of change

### Development

### Code review


package.json
```diff
@@ -20,6 +21,7 @@
     "@typescript-eslint/parser": "2.25.0",
     "babel-plugin-styled-components": "1.10.7",
     "chalk": "3.0.0",
+    "d3": "^5.15.0",
```
szhu, line 24: Can we use exact dependencies versions? Just because we're not committing the lockfile.

```suggestion
    "d3": "5.15.0",
```

(This is a decision I made when I set up this repo that I think is good, but I could be wrong. We can talk about it!)

package.json
```diff
@@ -38,6 +40,7 @@
     "react-is": "16.13.1",
     "react-router-dom": "5.1.2",
     "resize-observer-polyfill": "^1.5.1",
+    "semiotic": "^1.20.5",
```
szhu, line 43: also here, exact dependency version

src/impact-dashboard/CurveChart.js
```diff
@@ -0,0 +1,120 @@
+import { curveCatmullRom, format } from "d3";
```
szhu, line 1: Can we use `.tsx` for new files going forward? If you want a quick intro to ts if you haven't used it before, just let me know!

macfarlandian: I tried, I spent some time with the introductory docs yesterday and mostly just wasted a bunch of time thrashing ... when we set this up you said I'd be able to write regular javascript, and I'm not so sure it's the best use of my time right now to stop development and try to learn enough to get by. maybe after we get v0 launched? can I just enable `allowJS` in `tsconfig` for now and move on?

szhu: We do allow js already!

JS is fine for now. The main thing I'd like to note is that TS makes it easier for other people (including you, at a later time) to understand your code. We can definitely postpone the conversion! Maybe later this week?

macfarlandian: ah, I think I got my local env into a weird state where it was rejecting some of my JS imports, but yes you're right it does work as is. I'm sure TypeScript is lovely, I have been meaning to learn it someday but I think today is not that day! Let's maybe talk about it on Friday, in the meantime pls approve if everything else outstanding has been addressed (I think it has?)

PR #18, opened 2020-04-02
React context for model input data

macfarlandian:
## Description of the change

This creates a shared datastore for both the input forms and the curve charts, initialized to match all the input fields in our latest mocks. We are probably going to want to add some more actions later (at the least, probably one to populate the store with state & county level data from an external data source, once we have one set up) but this should get us started for now.

Also I know some TypeScript now? 😎 

## Type of change

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -76,7 +76,7 @@ function useEpidemicModelDispatch() {
 
   if (context === undefined) {
     throw new Error(
-      "useEpidemicModelDispatch must be used within a EpidemicModelProvider",
+      "useEpidemicModelDispatch must be used within an EpidemicModelProvider",
     );
   }
 
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -47,7 +47,7 @@ function epidemicModelReducer(state: State, action: Action): State {
   }
 }
 
-function EpidemicModelProducer({ children }: EpidemicModelProviderProps) {
+function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
   const [state, dispatch] = React.useReducer(epidemicModelReducer, {
     stateCode: "US",
     // TODO: is this really where the initial data should come from?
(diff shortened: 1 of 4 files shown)


src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return context;
+}
+
+function useEpidemicModelDispatch() {
+  const context = React.useContext(EpidemicModelDispatchContext);
+
+  if (context === undefined) {
+    throw new Error(
+      "useEpidemicModelDispatch must be used within a EpidemicModelProvider",
```
jessex, line 79: Nit: within an EpidemicModelProvider

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -0,0 +1,90 @@
+import React from "react";
+
+const {
+  populationAndHospitalData,
+} = require("../page-overview/assets/dataSource") as any;
```
szhu, line 5: - Can you move the `dataSource` file up into a more common folder? I'm thinking into `infection-model`. Or can make another top-level folder called `data` or something like that.
- You can rename the file to `.tsx`. You won't have to change anything about the contents of the file, but also you'll now be able to `import` it and not use `any`. 

We don't have to do this in this PR, but I feel like it wouldn't be too hard for a change to make.

src/impact-dashboard/ImpactDashboard.tsx
```diff
@@ -0,0 +1,29 @@
...
+  border-right: 1px solid black;
+  flex: 1 0 auto;
+  width: 350px;
+`;
+
+const ChartsContainer = styled.div`
+  flex: 2 0 auto;
+  width: 350px;
+  margin: 0 15px;
+`;
```
szhu, line 18: Thank you for using flexbox! I think this is the easier and more maintainable responsiveness solution for components like this.

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: I agree but we need to do a whole new pass at the external datasources to pull in more stuff for this model so I left it alone for now. would rather keep this PR small

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: 🤓

PR #19, opened 2020-04-02
Fix Now preview deployments

szhu:
Added an empty file to keep Now from complaining.


jessex:
Nice! Thanks so much, @szhu. Do we need to do anything else here or do we expect this to work going forward? Feel free to close this if we're all set.

szhu:
Looks like this is caused by https://github.com/zeit/now/issues/3370

This will be fixed after this PR is merged @jessex. Can you hit approve?

PR #25, opened 2020-04-06
Read some model input data from spreadsheet

macfarlandian:
## Description of the change

This fetches some county-level incarceration, hospital, and confirmed count data from our model in Google Sheets and populates the `EpidemicModelContext` with it. Involved shuffling some stuff around in the type definitions but otherwise is pretty straightforward. We may want to define some accessor functions further down the road to avoid going deep into this data structure from different parts of the application but I haven't included that for now. 

## Type of change

### Development

### Code review


src/@types/d3/index.d.ts
```diff
@@ -0,0 +1,20 @@
+import { DSVRowAny, DSVRowString } from "d3";
```
jessex, line 1: I'm guessing we copied this in from somewhere and if so, we should attribute the source at the top. If I'm wrong feel free to resolve this comment!

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -52,12 +81,65 @@ function epidemicModelReducer(
...
+      "https://docs.google.com/spreadsheets/d/e/2PACX-1vSeEO7JySaN21_Cxa7ON_x" +
+        "UHDM-EEOFSMIjOAoLf6YOXBurMRXZYPFi7x_aOe-0awqDcL4KZTK1NhVI/pub?gid=" +
+        "1836987932&single=true&output=csv",
+    ).then(async (response) => {
+      let rawCSV = await response.text();
+      // the first line is not the header row so we need to strip it
+      rawCSV = rawCSV.substring(rawCSV.indexOf("\n") + 1);
+
+      const parsedArray = csvParse(rawCSV, (row) => {
+        const transformedRow = autoType(row);
```
szhu, line 103: I think it might be safer to force-parse cells in each column as a certain type rather than auto-interpret the value type. We don't want a stray character in a spreadsheet throwing a cryptic error much further down the line in the data processing pipeline. Open to thoughts though!

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -52,12 +81,65 @@ function epidemicModelReducer(
...
     usePopulationSubsets: false,
   });
 
+  // fetch from external datasource
+  React.useEffect(() => {
+    fetch(
+      "https://docs.google.com/spreadsheets/d/e/2PACX-1vSeEO7JySaN21_Cxa7ON_x" +
+        "UHDM-EEOFSMIjOAoLf6YOXBurMRXZYPFi7x_aOe-0awqDcL4KZTK1NhVI/pub?gid=" +
+        "1836987932&single=true&output=csv",
+    ).then(async (response) => {
```
szhu, line 97: nit: if you're going to use `.catch`, you can just do:

```tsx
let response = await fetch(...)
// all the code below can just go here
```

See here for an example of how I did it for news feed events:
https://github.com/Recidiviz/covid19-dashboard/blob/master/src/news-feed-events/NewsFeedEvent.tsx#L24-L37

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -52,12 +81,65 @@ function epidemicModelReducer(
...
+        .filter((row) => !!row.County);
+
+      const nestedStateCounty = rollup(
+        parsedArray,
+        // there will only ever be one row object per county
+        (v: object[]) => camelcaseKeys(v[0]),
+        (d: DSVRowAny) => d.State as string,
+        (d: DSVRowAny) => d.County as string,
+        // some wrong/outdated typedefs for d3 are making typescript sad
+        // but this should check out
```
szhu, line 126: Note to self: grep codebase for ` as ` in the future to figure out which ts things need to be fixed.

src/impact-dashboard/EpidemicModelContext.tsx
jessex, line 97: Do we need an explicit `error` callback in case the fetch fails?

src/entry-point-client/index.tsx
```diff
@@ -1,3 +1,5 @@
+import "regenerator-runtime/runtime";
```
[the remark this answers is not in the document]
macfarlandian: good questions. 
1. for some reason adding the `camelcase` dependency started triggering this error. Possibly some difference in the way it was transpiled? I suspect we probably did need it all along for older browser support but just haven't triggered an error thus far. 
2. including this here covers static pages as well because we load it from `index.html`

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: I couldn't do that because you apparently can't pass an async function to `useEffect` ... the alternative would be to move all this to an inner function and then immediately call the inner function which didn't seem obviously better than just doing this

src/@types/d3/index.d.ts
macfarlandian: no believe it or not I had to figure this shit out on my own 😒

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: yes probably, I was just cutting corners. Let me put in a quick "sorry everything failed" state as a fallback here

szhu: Oh I see what you mean, sorry I forgot about that. The idiom that I've learned to learned to use is:

```tsx
useEffect(() => {
  async function effect(){
    ...
  }
  effect()
})

```

Obviously, has some of the same drawbacks of this solution, but at least we're not special-casing the first line to have to be written some other kind of way.

macfarlandian: yeah i guess that makes sense, i'll change it

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: I think I actually agree but I am going to punt on this for now .... after this I am going to work on some more updates to the epidemic model, at which point I will know which fields from the spreadsheet we actually care about and will parse them more explicitly

macfarlandian:
@jessex we probably will want to iterate on a more sophisticated error state later but for now this will at least prevent an infinite "loading..." screen if something goes wrong

src/entry-point-client/index.tsx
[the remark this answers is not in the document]
szhu: oh ok both of those make sense, thanks for investigating these!

PR #30, opened 2020-04-07
Epidemic model updates

macfarlandian:
## Description of the change

This brings the web model up to feature completeness for initial launch. This includes full age-bracket data for both population and confirmed cases, and incorporation of facility occupancy rates and housing types. Outputs have been validated against the latest spreadsheet and are accurate to within 0.8%.

This does break the original map-and-slider UI; our plan is to remove it from the launch UI and bring it back later so I just did the bare minimum to get Typescript to stop yelling at me about it. A subsequent PR will clean that up so I don't think we should fuss about it for now. 

## Type of change

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -15,6 +15,7 @@ type CountyLevelData = Map<string, Map<string, CountyLevelRecord>>;
 type Action = { type: "update"; payload: EpidemicModelUpdate };
 type Dispatch = (action: Action) => void;
 
+export type RateOfSpread = "low" | "moderate" | "high";
 // any field that we can update via reducer should be here,
 // and should probably be optional
 interface ModelInputsUpdate {
@@ -35,7 +36,9 @@ interface ModelInputsUpdate {
   ageUnknownCases?: number;
   ageUnknownPopulation?: number;
   confirmedCases?: number;
-  rateOfSpreadFactor?: number;
+  facilityDormitoryPct?: number;
+  facilityOccupancyPct?: number;
+  rateOfSpreadFactor?: RateOfSpread;
   staffCases?: number;
   staffPopulation?: number;
   totalIncarcerated?: number;
@@ -43,7 +46,7 @@ interface ModelInputsUpdate {
 }
 // some fields are required for calculations, define them here
 interface EpidemicModelInputs extends ModelInputsUpdate {
-  rateOfSpreadFactor: number;
+  rateOfSpreadFactor: RateOfSpread;
   usePopulationSubsets: boolean;
 }
 
@@ -93,8 +96,10 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
     stateCode: "US Total",
     countyName: "Total",
     countyLevelDataLoading: true,
-    rateOfSpreadFactor: 3.7,
+    rateOfSpreadFactor: "high",
     usePopulationSubsets: false,
+    facilityOccupancyPct: 1,
+    facilityDormitoryPct: 0.15,
   });
 
   // fetch from external datasource
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -18,17 +18,26 @@ type Dispatch = (action: Action) => void;
 // any field that we can update via reducer should be here,
 // and should probably be optional
 interface ModelInputsUpdate {
-  age0?: number;
-  age20?: number;
-  age45?: number;
-  age55?: number;
-  age65?: number;
-  age75?: number;
-  age85?: number;
-  ageUnknown?: number;
+  age0Cases?: number;
+  age0Population?: number;
+  age20Cases?: number;
+  age20Population?: number;
+  age45Cases?: number;
+  age45Population?: number;
+  age55Cases?: number;
+  age55Population?: number;
+  age65Cases?: number;
+  age65Population?: number;
+  age75Cases?: number;
+  age75Population?: number;
+  age85Cases?: number;
+  age85Population?: number;
+  ageUnknownCases?: number;
+  ageUnknownPopulation?: number;
   confirmedCases?: number;
   rateOfSpreadFactor?: number;
-  staff?: number;
+  staffCases?: number;
+  staffPopulation?: number;
   totalIncarcerated?: number;
   usePopulationSubsets?: boolean;
 }
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -48,6 +48,8 @@ interface ModelInputsUpdate {
 interface EpidemicModelInputs extends ModelInputsUpdate {
   rateOfSpreadFactor: RateOfSpread;
   usePopulationSubsets: boolean;
+  facilityDormitoryPct: number;
+  facilityOccupancyPct: number;
 }
 
 interface MetadataUpdate {
--- a/src/infection-model/index.tsx
+++ b/src/infection-model/index.tsx
@@ -1,3 +1,5 @@
+import { zip } from "d3-array";
+
 import { EpidemicModelInputs } from "../impact-dashboard/EpidemicModelContext";
 import { getAllValues, getColView } from "./matrixUtils";
 import { ageGroupIndex, getCurveProjections, seirIndex } from "./seir";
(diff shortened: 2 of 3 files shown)


src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -84,8 +98,10 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
     stateCode: "US Total",
     countyName: "Total",
     countyLevelDataLoading: true,
-    rateOfSpreadFactor: 3.7,
+    rateOfSpreadFactor: "high",
     usePopulationSubsets: false,
+    facilityOccupancyPct: 1,
```
jessex, line 103: I haven't checked the calculation logic using these downstream yet, but at first glance it seems wrong for these two facility housing type percentages to sum to greater than 1. Is that an issue?

src/infection-model/index.tsx
```diff
@@ -1,69 +1,93 @@
...
-    totalIncarcerated: incarceratedPopulation,
-    rateOfSpreadFactor: R0,
-  });
-  const bedsRequiredPerDay = curves.hospitalized;
-  const peakRequirement = Math.max(...bedsRequiredPerDay);
-  const peakUtilization = peakRequirement / hospitalBedCapacity;
-  const peakDay = bedsRequiredPerDay.indexOf(peakRequirement) + 1;
-  return { peakUtilization, peakDay };
+export function estimatePeakHospitalUse() {
+  // TODO: this function is now broken by the latest model updates
```
jessex, line 89: Just for my own knowledge, briefly: why does this and the spinner get broken by the latest model updates?

src/impact-dashboard/EpidemicModelContext.tsx
jessex: Ah sorry I thought these were two dueling types of housing, but I see that they are orthogonal. One is overall occupancy, the other is the split between housing types. Don't mind me, but leaving this comment for any other reviewers who may have made the same initial mistake.

src/infection-model/seir.tsx
```diff
@@ -113,14 +127,34 @@ function simulateOneDay(inputs: SimulationInputs & SingleDayInputs) {
...
   } else {
-    exposedDelta =
-      Math.min(
-        susceptible,
-        (beta * totalInfectious * susceptible) / totalPopulation,
-      ) -
-      alpha * exposed;
-    susceptibleDelta =
-      (-beta * totalInfectious * susceptible) / totalPopulation;
+    if (simulateStaff) {
```
jessex, line 130: I don't get at first glance why the more complex calculation is gated by us not "simulating staff." What obvious point am I missing here?

src/infection-model/seir.tsx
```diff
@@ -159,6 +203,23 @@ function getCurveProjections(inputs: SimulationInputs & CurveProjectionInputs) {
...
+    (1 - facilityOccupancyPct) *
+      (rateOfSpreadCells - rateOfSpreadCellsAdjustment);
+  let rateOfSpreadDorms = R0Dorms[rateOfSpreadFactor];
+  const rateOfSpreadDormsAdjustment = 1.7; // magic constant
+  rateOfSpreadDorms =
+    rateOfSpreadDorms -
+    (1 - facilityOccupancyPct) *
+      (rateOfSpreadDorms - rateOfSpreadDormsAdjustment);
+
+  const ratioExposedToInfected = 0.487804878;
```
jessex, line 220: Can we comment about why this variable is set to this value?

src/infection-model/seir.tsx
```diff
@@ -136,18 +170,28 @@ function simulateOneDay(inputs: SimulationInputs & SingleDayInputs) {
...
+    facilityDormitoryPct,
+    facilityOccupancyPct,
     numDays,
-    initiallyInfected,
+    rateOfSpreadFactor,
   } = inputs;
 
-  // this number can't be zero;
-  // if it's missing, substitute a conservative assumption
-  initiallyInfected = initiallyInfected || 1;
```
jessex, line 149: At first glance, I don't see this fallback happening for the age group'd initially infected numbers. Is that no longer needed, or am I not seeing where it happens?

src/infection-model/seir.tsx
```diff
@@ -167,27 +228,22 @@ function getCurveProjections(inputs: SimulationInputs & CurveProjectionInputs) {
...
-  singleDayState.set(
-    ageGroupIndex.ageUnknown,
-    seirIndex.infectious,
-    initiallyInfected,
-  );
-  singleDayState.set(
-    ageGroupIndex.ageUnknown,
-    seirIndex.susceptible,
-    singleDayState.get(0, seirIndex.susceptible) - initiallyInfected,
+  // initially everyone is either susceptible or infected
```
justkunz, line 231: Nit: this comment should include also include "exposed" as a group initialized

src/infection-model/index.tsx
macfarlandian: the short answer is that the slider produces a single R0 value, whereas the new model now takes a qualitative "low/moderate/high" output and maps that to two separate R0 values for cells vs dorms.

src/infection-model/seir.tsx
macfarlandian: my understanding is that the effect of facility type on staff is not as strong as it is on the incarcerated, so the staff modeling leaves some of that out ... @justkunz would know better than I would though

src/infection-model/seir.tsx
macfarlandian: it shouldn't be needed anymore but there is still a gap in the overall data flow that makes sure this overall value is populated by _something_ ... so yeah you are not seeing where it happens because it's not happening yet 😎 . I think that is an issue for a separate PR though because it is more tightly coupled with the form UI and validation logic

src/infection-model/seir.tsx
```diff
@@ -196,29 +252,42 @@ function getCurveProjections(inputs: SimulationInputs & CurveProjectionInputs) {
...
       });
       setRowValues(singleDayState, rowIndex, projectionForAgeGroup);
     });
 
     // sum up each column to get the total daily projection for each SEIR bucket
     for (let colIndex = 0; colIndex < singleDayState.shape[1]; colIndex++) {
-      currentDayProjections.push(
-        sum(getAllValues(getColView(singleDayState, colIndex))),
+      const incarceratedValues = getAllValues(
+        getColView(singleDayState, colIndex),
```
justkunz, line 269: Does `incarceratedValues` include the SEIR values from the staff group?

macfarlandian: no, on the line directly below this I remove the staff counts from this array. (it's a little obscure because the array gets mutated in place, but the variable name reflects what the final state will be ... I can change that if you think it's needlessly confusing)

src/infection-model/index.tsx
jessex: Is there no reasonable way to consolidate R0 values? I'm sure you discussed this with @sychang already, so just trying to understand if there's any quick solution to keep this around for launch.

src/infection-model/seir.tsx
justkunz: The intuition here was that the rate of spread for the faculty does not change under different facility types. I think an argument could be made that faculty have completely different R0 values, but at this point I doubt there is data to show that conclusively.

src/infection-model/seir.tsx
jessex: Do you think that logic is a launch blocker?

src/infection-model/seir.tsx
```diff
@@ -57,6 +57,26 @@ enum ageGroupIndex {
...
+// days from end of infectious period to hospital admission
+const dHospitalLag = 2.9;
+// days from hospital admission to hospital release (non-fatality scenario)
+const dHospitalRecovery = 22;
+// days from hospital admission to deceased (fatality scenario)
+const dHospitalFatality = 8.3;
+// probability case will be severe enough for the hospital
+const pSevereCase = 0.26;
+// factor for inferring exposure based on confirmed cases
+const ratioExposedToInfected = dIncubation / dSymptoms;
```
jessex, line 78: This value (2 / 5.1) is different from what was hard-coded below... Was this a bug or another misunderstanding on my part?

src/infection-model/seir.tsx
jessex: Got it, that makes a lot of sense. Thanks!

src/infection-model/seir.tsx
justkunz: This is `dIncubation`/`dInfectious` and it's the ratio of total exposed to total infected pop.

src/infection-model/index.tsx
macfarlandian: based on my conversation with Serena it didn't really seem worth figuring that out before getting to the rest of the stuff on our punchlist ... there is definitely something to do there it's just less of a priority than other blockers

src/infection-model/seir.tsx
macfarlandian: yeah I have already discussed that with Serena also

src/infection-model/seir.tsx
macfarlandian: ugh good catch! sloppy work on my part, I put in the wrong denominator. stand by...

src/infection-model/seir.tsx
jessex: Cool, as long as another PR is on the way!

src/infection-model/seir.tsx
justkunz: Cool, I was 99% sure that I didn't understand the JS but wanted to double check. I had a version of the spreadsheet previously where I forgot to remove the staff from the incarcerated totals, so it's top of mind

PR #31, opened 2020-04-08
Add fatalities curve and output table

macfarlandian:
## Description of the change

This adds an additional curve to the chart for Fatalities and also adds a table below the chart that contains some of the same data and some additional metrics, broken out by incarcerated vs staff populations. 

Screen Shot 2020-04-07 at 7 50 39 PM

There is a bit of redundancy in the table code; I tried not to build every data structure and table segment completely by hand but I also didn't fuss over making the solution perfectly elegant. There's also a bit of inefficiency here in that we are calculating the same projections twice, which may or may not be desirable in the interest of keeping the components independent of one another. These are things we can revisit later if we decide it's worthwhile but I didn't think they were launch-blockers.

## Type of change

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/CurveChartContainer.tsx
+++ b/src/impact-dashboard/CurveChartContainer.tsx
@@ -1,4 +1,8 @@
-import { calculateCurves } from "../infection-model";
+import { zip } from "d3-array";
+
+import { calculateCurves, CurveData } from "../infection-model";
+import { getAllValues, getColView } from "../infection-model/matrixUtils";
+import { seirIndex } from "../infection-model/seir";
 import { MarkColors } from "./ChartArea";
 import CurveChart from "./CurveChart";
 import { useEpidemicModelState } from "./EpidemicModelContext";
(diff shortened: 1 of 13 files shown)


src/page-test/TestPage.tsx
```diff
@@ -75,7 +75,7 @@ const TestPage: React.FC = () => {
       <div>
         <h1> h1 </h1>
         <p> paragraph </p>
-        <body> body </body>
```
szhu, line 78: nice catch, thanks for fixing the warning

src/impact-dashboard/ImpactProjectionTable.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+const LabelCell = styled.td<{ italic?: boolean }>`
+  color: ${Colors.forest};
+  font-weight: normal;
+  text-align: left;
+  opacity: 0.8;
+  font-style: ${(props) => (props.italic ? "italic" : "inherit")};
+`;
+
+numeral.nullFormat("N/A");
```
szhu, line 80: Whoa, this seems kind of global. If code somewhere else calls `numeral.nullFormat("something else")`, then the behavior of the code below is nondeterministic.

I think the most proper way to do this is to move these formatting functions into a separate file, so it's clear that that's the only place `numeral` functions are being run.

Otherwise, we can just add a TODO here for now.

macfarlandian: yeah it does, doesn't it .... I have a workaround that is probably safer, i'll just replace this

src/impact-dashboard/CurveChartContainer.tsx
```diff
@@ -7,20 +11,32 @@ interface Props {
...
+  // merge and filter the curve data to only what we need for the chart
+  const curveData = {
+    exposed: combinePopulations(projectionData, seirIndex.exposed),
+    fatalities: combinePopulations(projectionData, seirIndex.fatalities),
+    hospitalized: combinePopulations(projectionData, seirIndex.hospitalized),
+    infectious: combinePopulations(projectionData, seirIndex.infectious),
+  };
 
   return modelData.countyLevelDataLoading ? (
     <div>Loading...</div>
```
jessex, line 35: Can we replace this with the actual `Loading` component now?

PR #37, opened 2020-04-08
Fixes type issues in the loading spinner.

jessex:
## Description of the change

This fixes a type issue that previously caused the `any` requirement. However, I originally sought to solve the issue that when running `yarn dev` the loading spinner is properly centered in the view port, but when running the files in `/dist` after `yarn build` the spinner is in the exact top-left corner. So while this change should definitely be merged, I'd also be interested in knowing if anyone has ideas about why that may be happening.

(As a test I changed the /verify page to just return `<Loading />` directly and it was centered correctly. So something about the way it is invoked in `App.tsx` after `yarn build` is causing the centering to not work.)

## Type of change

### Development

### Code review

### DIFF
--- a/src/design-system/Loading.tsx
+++ b/src/design-system/Loading.tsx
@@ -1,24 +1,24 @@
+import { css } from "@emotion/core";
 import BounceLoader from "react-spinners/BounceLoader";
 
 import { GlobalStyles } from "../styles";
 
-const override = {
-  display: "block",
-  margin: "0 auto",
-  position: "fixed",
-  top: "50%",
-  left: "50%",
-  transform: "translate(-50%, -50%)",
-};
+const override = css`
+  display: block;
+  margin: 0 auto;
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  -ms-transform: translateX(-50%) translateY(-50%);
+  -webkit-transform: translate(-50%, -50%);
+  transform: translate(-50%, -50%);
+`;
 
 const Loading: React.FC = () => {
   return (
     <>
       <GlobalStyles />
-      <div>
-        {/* BounceLoader's type definition might be wrong. */}
-        <BounceLoader css={override as any} size={60} color={"#005450"} />
-      </div>
+      <BounceLoader css={override} size={60} color={"#005450"} />
     </>
   );
 };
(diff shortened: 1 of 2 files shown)


jessex:
Yeah I don't think `as any` is at all involved with the issue--I am just fixing the need for that independently.

I did more testing and found that if I simply change `if (loading)` in `App.tsx` to `if (true)` and then do `yarn build` and run the app from `/dist`, it puts the spinner right in the center of the page. But when navigating around and the spinner gets briefly shown, it is rendered in the top-left corner. I wonder if it is simply loading so fast that the browser doesn't get a chance to move it into center somehow? That doesn't seem to make a lot of sense as an explanation, but the observation remains.

But one more thing I note: when it is `if (loading)` then the `yarn build` output has the following warnings:
```
The component styled.button with the id of "sc-fzolEj" has been created dynamically.
You may see this warning because you've called styled inside another component.
To resolve this only create new StyledComponents outside of any render method and function component.
The component styled.div with the id of "sc-fzqBkg" has been created dynamically.
You may see this warning because you've called styled inside another component.
To resolve this only create new StyledComponents outside of any render method and function component.
```

When it is simply `if (true)` and no other changes, those warnings are no longer in the `yarn build` output. I'm getting out of my depth on this subject matter rapidly so just reporting what I see in case this means anything to anyone else.

szhu:
@jessex thanks for that investigation! That actually does explain what's going on -- it has to do with how we generate the static stylesheets that we use. We render the page and collect all the styles that are present in the initial screen.

That's why placing the component inside stuff makes the styles not work, and why they advise doing this:

> To resolve this only create new StyledComponents outside of any render method and function component.

Now that you're doing this in the PR, there should be no more problems!

jessex:
Unfortunately, this did not resolve the issue @szhu. There are no more warnings in `yarn build` as is, but all of my previous observations about times when the spinner loads in the center of the page versus the top-left still remain true.

szhu:
Oh okay. I'm out of time to debug this at the moment (working on my feature) but down to investigate tomorrow! I support whatever is needed to just make it work for now.

jessex:
The spinner works it just isn't centered in deployed environments. I have spent a couple of hours on this myself and don't think this can be considered launch blocking, though it would be good to figure out as a fast follow. I'm going to merge this for now and file an issue to work on proper centering.

PR #41, opened 2020-04-08
Clean up model data fetching from spreadsheet

macfarlandian:
## Description of the change

This adds a more explicit interface for only the columns we actually need from the model spreadsheet. It also adds an estimate of initial cases (using the same methodology as the Excel model) to seed the model with.

After the initial update, it now selects all the necessary generic inputs to plot the entire US in the chart and table, so that chart should now load properly and display something meaningful after the page loads. @szhu you are probably already aware that we are going to have to do something similar every time a new state or county is selected from the input forms (this is still a loose end in the data flow from my perspective but may already be part of what you're doing? let me know if this is something I can help with in the morning).

(also realized that I forgot to replace a dummy value in the table data with the real data from the context; that's fixed here too)

## Type of change

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -131,6 +131,8 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
             const totalIncarceratedPopulation = numeral(
               row["Total Incarcerated Population"],
             ).value();
+            const estimatedTotalCases =
+              (numeral(row.cases).value() || 0) * (1 / caseReportingRate);
             return {
               county: row.County || "",
               state: row.State || "",
@@ -139,7 +141,7 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
               estimatedIncarceratedCases:
                 (totalIncarceratedPopulation /
                   numeral(row["Total Population"]).value()) *
-                (1 / caseReportingRate),
+                estimatedTotalCases,
             };
           },
         )
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -1,5 +1,4 @@
-import camelcaseKeys from "camelcase-keys";
-import { autoType, csvParse, DSVRowAny } from "d3";
+import { csvParse, DSVRowAny } from "d3";
 import { rollup } from "d3-array";
 import numeral from "numeral";
 import React from "react";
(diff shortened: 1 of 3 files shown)


src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -120,29 +124,36 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
...
-          });
-
-          return transformedRow;
-        })
+        const parsedArray = csvParse(
+          rawCSV,
+          (row): CountyLevelRecord => {
+            // we only need a few columns, which we will explicitly format
+            const totalIncarceratedPopulation = numeral(
+              row["Total Incarcerated Population"],
```
jessex, line 132: Is there any chance this is ever null or empty? Will `numeral` auto convert this to 0?

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -120,29 +124,36 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
...
+            ).value();
+            const estimatedTotalCases =
+              (numeral(row.cases).value() || 0) * (1 / caseReportingRate);
+            return {
+              county: row.County || "",
+              state: row.State || "",
+              hospitalBeds: numeral(row["Hospital Beds"]).value(),
```
jessex, line 139: Same question here as above, since we're just indexing straight in with no fallback.

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -120,29 +124,36 @@ function EpidemicModelProvider({ children }: EpidemicModelProviderProps) {
...
+              totalIncarceratedPopulation,
+              estimatedIncarceratedCases:
+                (totalIncarceratedPopulation /
+                  numeral(row["Total Population"]).value()) *
```
jessex, line 143: Aaaaand same question here.

macfarlandian:
one thing to note about this PR is that it doesn't distinguish between jail and prison populations, it always uses the combined total — which, I have realized, is a little different from how the excel sheet currently works and may also not fit with the form UI. I'm not 100% sure if that's a problem or not, might be something for @sychang to weigh in on also

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: no, numeral will output `null` if the value is missing. looks like I'm not handling this consistently throughout this function — the `@types` definition for `numeral().value()` appears to actually be wrong, which is part of why I didn't catch this. I'll take another pass at these and make sure there are sensible fallback values for every `numeral`

macfarlandian:
update on the jail vs prison thing per separate conversation: this should be OK for now, the jail vs prison distinction hasn't made it to the public model yet anyway so it's not a regression. it'll change eventually though

PR #43, opened 2020-04-08
Add useQueryParams hook

notfelineit:
## Description of the change
This PR adds a `useQueryParams` hook that:
- Is configurable to replace or push to history using `setAction`
- Will prioritize any present URL query params as the default values
- Can also be passed default values 
- Updates the URL query params when values change
- Changes values with URL query params change

### How to Use
V1 of this hook looks like this:
```
const [values, setValues, setAction] = useQueryParams({ defaultVal1: 1, defaultVal2: 2 })

// Change from replacing (default) to pushing to history:
setAction('push')

// Change from pushing to replacing history:
setAction('replace')

// Set arbitrary form values:
setValues({ ...values, defaultVal2: 3 })

// If you want to narrow the set of valid query parameters to check, pass in a set when initializing the hook:
const validKeys = ['defaultVal1', 'defaultVal2']
const [values, setValues, setAction] = useQueryParams({ defaultVal1: 1, defaultVal2: 2 }, validKeys)
```

## Type of change

### Development

Added some testing buttons to the `/test-page` where you can try out this hook on a JSON that looks like:
```
{ val1: 1, val2: 2 }
```
Increment either of these values from buttons in the UI, or pass values for them as query params in the URL.

You can also change between replacing and pushing to history via buttons on the TestPage. 

#### Passing in initial values from URL
Screen Shot 2020-04-08 at 1 47 29 AM

#### Passing in initial values from state
urlfromstate

#### Replacing history vs. Pushing to history
replacehistory2
This gif demonstrates switching between replacing and pushing to history (notice the URL changing)

### Code review


src/hooks/useQueryParams.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    if (!isEqual(currentQuery, values)) {
+      location.search = queryString.stringify(values);
+      if (action == "replace") {
+        history.replace(location);
+      }
+      if (action == "push") {
+        history.push(location);
+      }
+    }
```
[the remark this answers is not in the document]
szhu: Actually, maybe it makes sense to return the following:
- `pushState`
- `replaceState`

Where `replaceState` will get called when the user edits the form values, and `pushState` will get  called when the user clicks copy link.

edit: okay that's actually kind of the same as `createNewState(action, values)`, either approach works

src/page-test/TestPage.tsx
```diff
@@ -83,6 +86,25 @@ const TestPage: React.FC = () => {
...
+        </button>
+        <button
+          onClick={() =>
+            setValues({ ...values, val2: (values.val2 as number) + 1 })
+          }
+        >
+          Val2: {values.val2}
+        </button>
+        <button onClick={() => setAction("push")}>Push History</button>
+        <button onClick={() => setAction("replace")}>Replace History</button>
```
szhu, line 106: On the form, _every_ form value update should replace history, not just when the user clicks a dedicated button. (aka, we'd need to call `setValues(...)`, then `setAction("replace")`, then `setAction("push")` some time later instead of just each `setValues(...)` above.)

notfelineit:
We might want to debounce this functionality, it could get slow if it updates the URL on every keystroke.

src/hooks/useQueryParams.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }, [values]);
+
+  // If the query params in the url changes, also set the values
+  useEffect(() => {
+    const nextValues = queryString.parse(location.search) as QueryParams;
+
+    if (!isEqual(values, nextValues)) {
+      setValues(nextValues);
+    }
+  }, [history.location.search]);
```
szhu, line 55: nit: As noted in the warning, the correct dependency is `location.search`. Actually, if we have `useHistory`, maybe we don't need to `useLocation` at all since `location` is always available inside `history`?

src/hooks/useQueryParams.tsx
```diff
@@ -0,0 +1,60 @@
...
+): [
+  QueryParams,
+  Dispatch<SetStateAction<QueryParams>>,
+  Dispatch<SetStateAction<HistoryAction>>,
+] => {
+  const history = useHistory();
+  const location = useLocation();
+
+  const defaultQueryParams = queryString.parse(location.search) as QueryParams;
+  const hasExistingURLQueryParams = Object.keys(defaultQueryParams).length > 0;
```
szhu, line 25: Note: Sometimes a custom query param will get added without the user's consent. For example, Facebook adds `fbclid` or something like that to links you click that take you away from Facebook. We'll probably need to let the hook user pass in a custom way to detect whether the existing query params actually count as "has existing".

package.json
```diff
@@ -20,9 +20,11 @@
     "@types/camelcase-keys": "5.1.1",
     "@types/d3": "5.7.2",
     "@types/d3-array": "2.0.0",
+    "@types/lodash": "^4.14.149",
```
szhu, line 23: nit: Exact types for all of these?

package.json
```diff
@@ -46,13 +48,15 @@
...
+    "lodash": "^4.17.15",
     "ndarray": "1.0.19",
     "npm-run-all": "4.1.5",
-    "numeral": "2.0.6",
+    "numeral": "^2.0.6",
     "parcel": "1.12.4",
     "prettier": "2.0.2",
-    "react": "16.13.1",
-    "react-dom": "16.13.1",
+    "query-string": "^6.12.0",
```
szhu, line 57: Note: I've heard comparisons of the `query-string` and `qs` packages before. I think the last time I checked, I ended up using `qs` because it supports nesting but `query-string` doesn't? For example, I think you can pass `{a: {b: 'c'}}` to `qs` but not `query-string`.

Anyway, not a big deal if you can't look into it, maybe we don't need that functionality. And we can always swap it out later pretty easily.

src/hooks/useQueryParams.tsx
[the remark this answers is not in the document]
macfarlandian: I second this, but I don't think it's a showstopper for the initial release. Wouldn't be too hard to go back and change later since we're only planning to use it in one or two places.

src/hooks/useQueryParams.tsx
```diff
@@ -0,0 +1,60 @@
...
+    if (!isEqual(currentQuery, values)) {
+      location.search = queryString.stringify(values);
+      if (action == "replace") {
+        history.replace(location);
+      }
+      if (action == "push") {
+        history.push(location);
+      }
+    }
+  }, [values]);
```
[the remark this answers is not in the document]
notfelineit: I don't have the warning - would you mind just pasting it for visibilitiy?

src/hooks/useQueryParams.tsx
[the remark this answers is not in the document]
notfelineit: Those are all good points, and things I considered. Since you are the client of this hook, if that design makes more sense, I'll update it to return two kinds of `setValues`.

src/page-test/TestPage.tsx
notfelineit: Yeah, the default behavior is that every form value update replaces the history. The buttons here merely provide an explicit way of testing the hook behavior.

notfelineit: Originally designed that way because we'd be `replacing` more than we'd be `pushing` (from our convo yesterday)

src/hooks/useQueryParams.tsx
notfelineit: Location is indeed always inside history, I just didn't want to change history.location  history is mutable while useLocation() is not 😄 

Please post the warning!

package.json
notfelineit: Sounds good - will keep as query-string for now, but when you need nested query strings, please feel free to change this! 😄

macfarlandian:
going to merge this so we can work on integrating it with the form. Thanks for the great work @notfelineit !

PR #53, opened 2020-04-09
Persist model inputs in URL

macfarlandian:
## Description of the change

This uses @notfelineit 's new hook to persist form context data to the URL for sharing, bookmarking, etc. It wound up being way more of a beast than I expected but here is where it's at right now:

- updates to the location and facility inputs trigger a history state replacement (meaning no back/forward navigation)
- upon the initial page load, any existing URL parameters are incorporated into the existing state

I initially hoped to preserve back/forward navigation, but I got too deep in the weeds and ultimately abandoned that for now. I think it's still possible but maybe can follow later (unless someone has a quick solution that has evaded my pudding brain today).

There were some challenges that I worked around for now: having the `useQueryState` maintain the same reference across renders would make things cleaner, as would more properly reconciling the query type definitions with the model. You will see that I cheated on both of these; it works as far as I can tell but definitely there are some possible edge case bugs. 

Also, there seem to be lots of annoying whitespace changes in this diff! Fair warning that you might want to turn those off while reviewing.

## Type of change

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -1,7 +1,11 @@
 import { csvParse, DSVRowAny } from "d3";
 import { rollup } from "d3-array";
+import { mapValues, pick } from "lodash";
 import numeral from "numeral";
-import React from "react";
+import React, { useEffect } from "react";
+import { useHistory } from "react-router-dom";
+
+import useQueryParams from "../hooks/useQueryParams";
 
 type CountyLevelRecord = {
   hospitalBeds: number;
@@ -13,16 +17,7 @@ type CountyLevelRecord = {
 
 type CountyLevelData = Map<string, Map<string, CountyLevelRecord>>;
 
-type Action =
-  | { type: "update"; payload: EpidemicModelUpdate }
-  | {
-      type: "reset";
-      payload: {
-        dataSource?: CountyLevelData;
-        stateCode?: string;
-        countyName?: string;
-      };
-    };
+type Action = { type: "update"; payload: EpidemicModelUpdate };
 type Dispatch = (action: Action) => void;
 
 export enum RateOfSpread {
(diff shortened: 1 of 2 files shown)


src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -101,13 +137,34 @@ interface ResetPayload {
...
+    countyLevelDataLoading: false,
+    totalIncarcerated: dataSource.get(stateCode)?.get(countyName)
+      ?.totalIncarceratedPopulation,
+    hospitalBeds: dataSource.get(stateCode)?.get(countyName)?.hospitalBeds,
+    ageUnknownPopulation: dataSource.get(stateCode)?.get(countyName)
+      ?.totalIncarceratedPopulation,
+    ageUnknownCases: dataSource.get(stateCode)?.get(countyName)
+      ?.estimatedIncarceratedCases,
+    confirmedCases: dataSource.get(stateCode)?.get(countyName)
+      ?.estimatedIncarceratedCases,
```
jessex, line 158: I know this was already the case, but should this actually be _estimated_ here? The same value as for `ageUnknownCases`?

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -148,93 +186,166 @@ function epidemicModelReducer(
...
+      }
+      return Object.assign({}, state, updates);
   }
 }
 
+function sanitizeQueryParams(rawQueryParams: QueryParams) {
+  return mapValues(pick(rawQueryParams, urlParamKeys), (value) => {
+    // most of these are numbers but some are strings
+    const n = numeral(value).value();
+    return n != null ? n : "" + value;
```
jessex, line 216: Plus sign concatenation in JS always freaks me out because I'm never sure what I'm gonna get. Is this canonical, or should we do `str(value)`?

src/impact-dashboard/ImpactDashboard.tsx
```diff
@@ -54,21 +54,7 @@ function useModel() {
     dispatch({ type: "update", payload: update });
   }
 
-  function resetModel(stateCode?: string, countyName?: string) {
```
jessex, line 57: I don't understand at first glance why `resetModel` is no longer required?

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: "confirmed cases" may be kind of a misnomer; this is also what the spreadsheet does. we adjust the number of confirmed cases we get from the NY Times dataset by some underreporting factor and use that as the default. it _is_ confusing, maybe we ought to just rename it?

src/impact-dashboard/ImpactDashboard.tsx
macfarlandian: the logic of when to use it was too convoluted and distributed too much between here and the UI ... basically the work it did was factored out into two new functions, `getResetBase` and `getLocaleData`, that can be used independently of one another to achieve the same result. now the `update` action decides when it wants to reset the state, rather than having the input UI make that decision. (we don't want the input UI to have to know about the URL business, is how this all came about)

src/impact-dashboard/EpidemicModelContext.tsx
jessex: Got it, thanks for the context. Let's file that away as a tech debt task. I'd rather not make any modeling changes at this precise instance in time, to minimize regression risk.

src/impact-dashboard/ImpactDashboard.tsx
jessex: Nice separation of concerns. Makes sense.

PR #68, opened 2020-04-10
Detect IE and display a message

AhmedMugtaba:
Added a way to detect IE and display a message to users, using a package called react-device-detect. I tested it using safari user agent simulation and it works fine. would love to know if this would work on a real IE. cheers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "prettier": "2.0.2",
     "query-string": "6.12.0",
     "react": "^16.0.0",
+    "react-device-detect": "^1.11.14",
     "react-dom": "^16.0.0",
     "react-hot-loader": "4.12.20",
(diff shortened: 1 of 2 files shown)


jessex:
@AhmedMugtaba, @szhu: unfortunate news, but Sean's intuition was right. This didn't work running on IE11 in my Parallels VM. We get the same IE11 syntax error in the bundled `entry-point-client-*.js` file. This is the same reason my previous attempt with `react-render-in-browser` didn't work: the JS simply won't be executed due to the syntax error happening on initial page load. Only the statically rendered HTML shows, hence the partial form and the infinite spinner that IE11 users see.

@szhu: would you mind talking to @AhmedMugtaba about the non-JS solutions for this problem that might work for us, if you two haven't synced up already?

PR #82, opened 2020-04-10
Add "unsupported browser" page and functionality

### LINKED ISSUE #72: Add banner indicating current lack of IE support
If people try to land on the app from IE 11, display a message that asks people to switch browsers or email covid@recidiviz.org if this is a significant concern.

The solution will likely need to avoid using JS, since the broken IE11 experience is due to IE11 not understanding the bundled JS in spite of all of our best intentions.

### LINKED ISSUE #80: Provide IE11 support
The current experience for IE11 users is broken. The landing page shows a partially rendered form behind an infinite loading spinner (though this is likely to change to an infinite spinner over a blank background after a recent change), and the console shows an error message indicating a syntax error in the bundled JS file.

After many hours investigating the issue between @szhu and @jessex, we had to punt on support for now. #72 is adding a banner, ideally as a temporary measure, indicating that IE11 is not currently supported. 

More notes to come on the investigation and attempts made so far...

szhu:
## Description of the change

Show the following page if the user is using a browser that runs into an error running the app JS:

image

Note that the "unsupported browser" page works because it is rendered statically.

Fixes #72
Closes #68 

## Type of change

### Development

### Code review

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,6 +55,15 @@
     <script src="https://code.jquery.com/jquery-3.4.1.min.js"></script>
     <script src="https://d3js.org/d3.v5.min.js"></script>
     <script src="src/entry-point-client/index.tsx"></script>
+
+    <!--
+      Redirect if src/entry-point-client/index.tsx doesn't load properly.
+      The variable `window.appLoaded` is set at the bottom of index.tsx.
+     -->
+    <script>
+      if (location.pathname === "/" && !window.appLoaded)
+        location.href = "/unsupported-browser";
+    </script>
     <link
       href="https://fonts.googleapis.com/css?family=Libre+Baskerville|Poppins|Rubik:300,400&display=swap"
       rel="stylesheet"
--- a/src/page-unsupported-browser/UnsupportedBrowserPage.tsx
+++ b/src/page-unsupported-browser/UnsupportedBrowserPage.tsx
@@ -12,19 +12,21 @@ const UnsupportedBrowserPage: React.FC = () => (
         <main className="py-8 sm:py-16 sm:py-24">
           <div className="max-w-3xl mx-auto px-6 sm:px-8 lg:px-10 flex flex-col justify-start">
             <p className="leading-7 text-base sm:text-lg my-6">
-              Sorry, you're using a legacy browser that that site does not
-              support.
+              We’re sorry, legacy browsers such as Internet Explorer are not
+              currently supported by the COVID-19 model. Please try loading the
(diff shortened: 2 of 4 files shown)


index.html
```diff
@@ -55,6 +55,15 @@
     <script src="https://code.jquery.com/jquery-3.4.1.min.js"></script>
     <script src="https://d3js.org/d3.v5.min.js"></script>
     <script src="src/entry-point-client/index.tsx"></script>
+
+    <!--
+      Redirect if src/entry-point-client/index.tsx doesn't load properly.
+      The variable `window.appLoaded` is set at the bottom of index.tsx.
+     -->
+    <script>
+      if (location.pathname === "/" && !window.appLoaded)
```
szhu, line 64: If the synchronous part of the app JS crashes for any reason, this would show the "unsupported browser" page. There would be some false positives in case we just write buggy code and it crashes the app -- but at least they'll be sent to a link where they can contact us.

If we want to be less aggressive in redirecting, we can:
- Only test for the syntax feature we know we need: `try { eval("()=>1") } catch { redirect }`
- Place setting `window.appLoaded` right after the imports in `index.tsx`. Then, if synchronous part of the React code causes any issues, we won't redirect (but of course the user will be left with a broken page).

Let me know what you think!

jessex: Yeah I think this is great, but I'm going to revise the copy to give us a bit of a messaging out if this is seen in error. Thanks!

src/page-unsupported-browser/UnsupportedBrowserPage.tsx
```diff
@@ -0,0 +1,36 @@
+import styled from "styled-components";
+
+import SiteHeader from "../site-header/SiteHeader";
+
+const GetInvolvedPageDiv = styled.div``;
+
+const UnsupportedBrowserPage: React.FC = () => (
```
jessex, line 7: Fantastic stuff, @szhu! Here is the finalized copy for the page, then we're good to go:




We're sorry, legacy browsers such as Internet Explorer are not currently supported by the COVID-19 model. Please try loading the website in a modern browser such as Microsoft Edge, Google Chrome, Firefox, or Safari.

If you think you are seeing this message in error, or if you can't switch browsers, please email us at <a href="mailto:covid@recidiviz.org?Subject=COVID%20Model%20Browser%20Support" className="font-semibold">covid@recidiviz.org</a> so we can provide support.

szhu: ah, forgot about Safari!

image

PR #96, opened 2020-04-14
Add R0 selector to input forms

### LINKED ISSUE #95: Add user input for R0

macfarlandian:
## Description of the change

This adds a dropdown for the user to input a rate of spread factor. The model calculations already accept this value, it just never quite made it into the UI, so no calc updates were required. 

Screen Shot 2020-04-13 at 6 08 41 PM

In the process I also made some style tweaks to the input elements for consistency, most notably adding a caret to `InputSelect` as a visual indicator that it's a dropown.

## Type of change

## Related issues

Closes #95 

### Development

### Code review

### DIFF
--- a/src/design-system/Input.tsx
+++ b/src/design-system/Input.tsx
@@ -1,6 +1,8 @@
 import { useEffect, useState } from "react";
 import styled from "styled-components";
 
+import Colors from "./Colors";
+
 export interface InputLabelProps {
   labelAbove?: string;
   labelHelp?: string;
@@ -38,17 +40,15 @@ export function useInputValue<T>(props: InputValueProps<T>) {
 export type InputBaseProps<T> = InputLabelProps & InputValueProps<T>;
 
 export const StyledInput = styled.input`
-  background: #e0e4e4;
+  background: ${Colors.gray};
   border-radius: 2px;
   border: none;
   box-shadow: none;
   box-sizing: border-box;
-  color: #00413e;
+  color: ${Colors.forest};
   flex: 1 1 auto;
-  font-family: "Rubik", sans-serif;
-  font-size: 16px;
+  font: 16px/1.2 "Rubik", sans-serif;
   height: 48px;
-  line-height: 1.2;
   margin-top: 8px;
   outline: 0 solid transparent;
   padding: 0 16px;
--- a/src/design-system/Input.tsx
+++ b/src/design-system/Input.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useState } from "react";
+import styled from "styled-components";
 
 export interface InputLabelProps {
   labelAbove?: string;
@@ -35,3 +36,21 @@ export function useInputValue<T>(props: InputValueProps<T>) {
 }
 
 export type InputBaseProps<T> = InputLabelProps & InputValueProps<T>;
+
+export const StyledInput = styled.input`
+  background: #e0e4e4;
+  border-radius: 2px;
+  border: none;
+  box-shadow: none;
+  box-sizing: border-box;
+  color: #00413e;
+  flex: 1 1 auto;
+  font-family: "Rubik", sans-serif;
+  font-size: 16px;
+  height: 48px;
+  line-height: 1.2;
+  margin-top: 8px;
+  outline: 0 solid transparent;
+  padding: 0 16px;
+  width: 100%;
+`;
--- a/src/design-system/InputSelect.tsx
+++ b/src/design-system/InputSelect.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import styled from "styled-components";
 
+import { StyledInput } from "./Input";
 import TextLabel from "./TextLabel";
 
 interface Props {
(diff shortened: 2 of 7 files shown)


szhu:
Where is the Now bot and the GitHub actions bot?

src/design-system/Input.tsx
```diff
@@ -35,3 +36,21 @@ export function useInputValue<T>(props: InputValueProps<T>) {
...
+export const StyledInput = styled.input`
+  background: #e0e4e4;
+  border-radius: 2px;
+  border: none;
+  box-shadow: none;
+  box-sizing: border-box;
+  color: #00413e;
+  flex: 1 1 auto;
+  font-family: "Rubik", sans-serif;
+  font-size: 16px;
```
szhu, line 49: To reduce the number of lines here, consider using the `font` shorthand property.

You can do:

```css
  font: 16px/1.2 "Rubik", sans-serif;
```

This would replace both the two lines here and `line-height` below. The thing I like the most about this property is that in encourages you to set everything font-related one line. This makes it easier to reason about how a font would look, since it's all grouped in the same place, and makes it harder to accidentally omit font-related properties.

Let me know if you think this is helpful. If you don't think so, no need to use it!

src/design-system/InputSelect.tsx
```diff
@@ -20,32 +22,46 @@ const SelectContainer = styled.div`
...
-  width: 100%;
-  border: none;
-  padding: 18px;
-  background: #e0e4e4;
-  outline: 0 solid transparent;
-  margin-top: 8px;
-  border-radius: 2px;
-  font-size: 16px;
-  color: #00413e;
+const caretSize = 5;
```
szhu, line 25: Thanks for including this as a constant so that there are less magic numbers below.

- nit: Consider capitalizing this, since it is a constant. Lowercase is fine if you think this could become variable sometime in the future.
- ~Can you briefly explain what the value of `caretSize` represents and place it in a comment above?~ Nevermind, I just read the description and now I know what the caret is, I guess that's pretty obvious…

src/impact-dashboard/MitigationInformation.tsx
```diff
@@ -19,6 +24,14 @@ type RowProps = PlannedRelease & {
   updateRelease: (opts: ReleaseUpdate) => void;
 };
 
+const rateOfSpreadDisplayText = {
```
szhu, line 27: You don't have to do this, but I tend to add a type for these things as follows:

```suggestion
const rateOfSpreadDisplayText: {[key: RateOfSpread]: string} = {
```

This makes it so that if you ever add an enum value to `RateOfSpread`, you'll immediately get a ts error here telling you that you didn't account for the new value.

Also consider capitalizing since it's a constant.

src/impact-dashboard/MitigationInformation.tsx
```diff
@@ -97,13 +118,38 @@ const MitigationInformation: React.FC = () => {
...
+        spread) in the model.
+      </Description>
+      <FormGrid>
+        <FormGridRow>
+          <FormGridCell>
+            <InputSelect
+              label="Rate of spread"
+              onChange={(e) =>
+                updateModel({
+                  rateOfSpreadFactor: e.target.value as RateOfSpread,
```
szhu, line 132: This is a lot of nesting! You don't have to do it in this PR, but consider reducing nesting at some point. An easy way to do this is to pull out the InputSelect as follows:

```tsx
  let rateOfSpreadInput = <InputSelect ... />

  ...

  return (
    <Container>
      <Description>
        Select the most likely rate of spread. This will change the R0 (rate of
        spread) in the model.
      </Description>
      <FormGrid>
        <FormGridRow>
          <FormGridCell>
            {rateOfSpreadInput}
```

src/design-system/Input.tsx
```diff
@@ -35,3 +36,21 @@ export function useInputValue<T>(props: InputValueProps<T>) {
 }
 
 export type InputBaseProps<T> = InputLabelProps & InputValueProps<T>;
+
+export const StyledInput = styled.input`
+  background: #e0e4e4;
```
szhu, line 41: Can you use a color from Colors.tsx or add it there if it doesn't already exist?

src/design-system/Input.tsx
```diff
@@ -35,3 +36,21 @@ export function useInputValue<T>(props: InputValueProps<T>) {
 }
 
 export type InputBaseProps<T> = InputLabelProps & InputValueProps<T>;
+
+export const StyledInput = styled.input`
```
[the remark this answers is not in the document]
macfarlandian: yeah that's the idea ... there were some weird differences between the different kinds of inputs that would reveal themselves when I started tweaking the styles (around padding and line height and things of that nature). As seen in `InputSelect` you can even apply these styles to other element types such as `<select>` by using the StyledComponent's `as` prop.

src/design-system/InputSelect.tsx
macfarlandian: I don't know whether we have really settled on a style for uppercasing constants (nothing linted at least, that I'm aware of) but if pressed I would favor [this rule](https://github.com/airbnb/javascript/#naming--uppercase), which advises against capitalizing here as well as in `MitigationInformation.tsx`

Issue #98, opened 2020-04-14
Add React components for ON/OFF Toggles - MultiFacility page


daschi:
**What needs to be done? Why does it need to be done?**
React component for ON/OFF toggles #93 
https://www.figma.com/proto/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=735%3A130&viewport=403%2C338%2C0.09717194736003876&scaling=min-zoom

jessex:
This was accidentally not auto-closed by #101. Closing manually.

Issue #93, opened 2020-04-13
Multi-facility support: Develop React components


cawarren:
**Is your feature request related to a problem? Please describe.**
The full application shown here will be assembled later this week:
https://www.figma.com/proto/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=735%3A130&viewport=403%2C338%2C0.09717194736003876&scaling=min-zoom

In order to expedite development, we'd like to get most of the small component widgets developed.

**Describe the solution you'd like**
React components for the following elements from the mocks:
- ON/OFF toggles
- On-hover ( ? ) tooltips
- ~~Modal dialog with title, corner X, the 1, 2, 3 steps, and buttons~~ No longer in mocks
- ~~Menus (in the library view)~~ Pushed to next release (multi-scenario)
- The dark-gray 'promo' box
- If time, the facility rows

**Describe alternatives you've considered**
The SWE developing the full design could build these de novo, but this would likely push out the anticipated date of completion.

**Additional context**
@jessex Please take a look and revise this bug if needed. We can also hang FRs for each react component off of this if needed.

cawarren:
Updated mocks (most of the updates are minor, but there are a couple that aren't):
https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=910%3A130

If you have any questions, feel free to ping me.

jessex:
Everything in the multi-facility release has been completed! Items for the later multi-scenario release are deferred to newer issues. Closing this out.

PR #101, opened 2020-04-14
Add toggle component for  multi-facility page

daschi:
## Description of the change
 
> Adds a `InputToggle` component that takes `toggled` and `onChange` props. 

This is my first PR and my first time working with styled-components and TypeScript, so I appreciate any and all feedback! Thank you for including me in this project, I'm looking forward to working together. 

The first commit is from @jovergaag and adds a multi-facility page so that we can stage the components. 

## Type of change

## Related issues

Closes [#98]
Part of #93 

### Development

### Code review

### DIFF
--- a/src/design-system/InputToggle.tsx
+++ b/src/design-system/InputToggle.tsx
@@ -1,0 +1,80 @@
+import React from "react";
+import styled from "styled-components";
+
+import Colors from "./Colors";
+
+interface Props {
+  toggled?: boolean;
+  onChange: (e: React.ChangeEvent) => void;
+}
+
+const ToggleContainer = styled.div`
+  display: flex;
+  flex-direction: row nowrap;
+`;
+
+const Label = styled.label<{ toggled?: boolean }>`
+  align-self: center;
+  color: ${(props) => (props.toggled ? Colors.teal : Colors.darkTeal)};
+  font-size: 14px;
+  font-family: "Poppins", sans-serif;
+  min-width: 50px;
+  padding: 0 1em;
+`;
+
+const ToggleInput = styled.input`
+  height: 0;
+  visibility: hidden;
+  width: 0;
+`;
+
+const ToggleButtonContainer = styled.label<{ toggled?: boolean }>`
+  align-items: center;
+  background: ${(props) => (props.toggled ? Colors.teal : Colors.darkTeal)};
+  border-radius: 20px;
+  cursor: pointer;
+  display: flex;
+  height: 26px;
+  justify-content: space-between;
+  position: relative;
+  transition: background-color 0.2s;
+  width: 48px;
+`;
+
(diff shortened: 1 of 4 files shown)


src/entry-point-client/PageList.tsx
```diff
@@ -2,6 +2,7 @@ import GetInvolvedPage from "../page-get-involved/GetInvolvedPage";
 import OverviewPage from "../page-overview/OverviewPage";
 import UnsupportedBrowserPage from "../page-unsupported-browser/UnsupportedBrowserPage";
 import VerificationNeeded from "../page-verification-needed/VerificationNeeded";
+import MultiFacilityPage from "../page-multi-facility/MultiFacilityPage";
```
jessex, line 5: Hey, @daschi! I commented the same on Jen's PR that unfortunately our README didn't make explicit that you should run linting locally before pushing up commits. You can use `yarn lint-all-check` and/or `yarn lint-all-fix`. This issue is in Jen's commit and it looks like you don't have any linting issues in your own stuff, but just wanted to let you know!

src/design-system/InputToggle.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const InputToggle: React.FC<Props> = (props) => {
+  const { toggled, onChange } = props;
+  const label = toggled ? "On" : "Off";
+
+  return (
+    <ToggleContainer>
+      <Label toggled={toggled}>{label}</Label>
+      <ToggleInput
+        id="ToggleInput"
```
jessex, line 66: I'm not 100% sure about this (not super experienced on the frontend), but if this becomes an id on the rendered HTML element and there are many of these components on the page, then they'd all share the id and we may have issues with any code that tries to identify a given input in the document by id. Do we need this id field or can we just omit it?

macfarlandian: good callout ... it looks like we do need it, @daschi i would recommend just augmenting it with `lodash.uniqueId` since we already have that package as a dependency

daschi: Oh yes, great point--I'll add that!

src/design-system/InputToggle.tsx
```diff
@@ -57,18 +58,19 @@ const ToggleButton = styled.span<{ toggled?: boolean }>`
 
 const InputToggle: React.FC<Props> = (props) => {
   const { toggled, onChange } = props;
+  const inputId = uniqueId("ToggleInput_");
```
macfarlandian, line 61: hm so ... this will generate a new ID on every render, right? I don't think it will break anything but it could cause unnecessary DOM updates. probably not a huge deal but maybe better to store the unique ID once in a state hook and then use that instead?

daschi: Definitely, that makes sense. I can update this now to use a state hook.

PR #104, opened 2020-04-15
Add Modal components for MultiFacilityPage

### LINKED ISSUE #102: Add Modal dialog for MultiFacilityPage
Issue #93 

Add a React component for a Modal dialog that has a title, corner X, and 1/2/3 steps and buttons.

daschi:
## Description of the change

> Adds a Modal component with a ModalTrigger and a ModalDialog. Also adds `@types/svg.d.ts` for importing svgs. 

I'm opening this PR as a WIP, I'll continue adding the following components unless someone else comments that they are working on it!

TODO: 
https://www.figma.com/proto/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=735%3A545&viewport=403%2C338%2C0.09717194736003876&scaling=min-zoom

- [x] Add Step components for the 1 / 2 / 3 steps at the bottom of the Modal 
- [x] Add the Next button in the bottom right corner
- [x] Add a ref to the modal dialog so that only click events on the background trigger the modal to close

I am assuming that the actual modal content with the form submission is not part of this task, but let me know if it should be!

Thanks!

## Type of change

## Related issues
Related to #93 
Closes #102

### Development

### Code review

### DIFF
--- a/src/design-system/InputButton.tsx
+++ b/src/design-system/InputButton.tsx
@@ -8,8 +8,7 @@ const StyledButton = styled.button<Props>`
   color: white;
   font-family: "Poppins", sans-serif;
   height: 48px;
-  width: ${(props) =>
-    props.styles && props.styles.width ? props.styles.width : "200px"};
+  width: ${(props) => props.styles?.width || "200px"};
   outline: none;
 `;
 
--- a/src/design-system/InputButton.tsx
+++ b/src/design-system/InputButton.tsx
@@ -1,20 +1,16 @@
 import React from "react";
-import styled, { InterpolationFunction } from "styled-components";
+import styled from "styled-components";
 
-interface StyleProps {
-  styles: React.CSSProperties;
-}
-
-const StyledButton = styled.button<StyleProps>`
+const StyledButton = styled.button<Props>`
   background: #00615c;
   font-size: 16px;
   border-radius: 12px;
   color: white;
   font-family: "Poppins", sans-serif;
   height: 48px;
-  width: 200px;
+  width: ${(props) =>
+    props.styles && props.styles.width ? props.styles.width : "200px"};
   outline: none;
-  ${(props: StyleProps) => props.styles as InterpolationFunction<StyleProps>};
 `;
 
 interface Props {
@@ -25,7 +21,7 @@ interface Props {
 
 const InputButton: React.FC<Props> = (props) => {
   return (
-    <StyledButton styles={props.styles || {}} onClick={props.onClick}>
+    <StyledButton styles={props.styles} onClick={props.onClick}>
       {props.label}
     </StyledButton>
   );
--- a/src/design-system/InputButton.tsx
+++ b/src/design-system/InputButton.tsx
@@ -1,7 +1,11 @@
 import React from "react";
-import styled from "styled-components";
+import styled, { InterpolationFunction } from "styled-components";
 
-const StyledButton = styled.button`
+interface StyleProps {
+  styles: React.CSSProperties;
+}
+
+const StyledButton = styled.button<StyleProps>`
   background: #00615c;
   font-size: 16px;
   border-radius: 12px;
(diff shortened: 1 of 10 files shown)


jessex:
Thanks for the progress so far, @daschi! The basic modal functionality and styling looks good so far.

src/design-system/InputButton.tsx
```diff
@@ -10,15 +14,21 @@ const StyledButton = styled.button`
   height: 48px;
   width: 200px;
   outline: none;
+  ${(props: StyleProps) => props.styles as InterpolationFunction<StyleProps>};
```
macfarlandian, line 17: this is clever but I also think I don't quite understand what's going on here. there's a bunch of assertions being made, how sure are you that this is type-safe?

src/design-system/Modal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  color: ${Colors.green};
+  cursor: pointer;
+  font-family: 'Libre Baskerville', serif;
+  font-size: 32px;
+  line-height: 32px
+  letter-spacing: -0.03em
+`;
+
+interface Props {
+  modalTitle?: string;
```
macfarlandian, line 19: you could export this prop from the child component that cares about it so that you don't have to define it twice and keep the two in sync as they change

src/design-system/Modal.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useState } from "react";
+import styled from "styled-components";
+
+import Colors from "./Colors";
+import ModalDialog from "./ModalDialog";
+
+const ModalContainer = styled.div``;
+
+const ModalTrigger = styled.h1`
```
macfarlandian, line 9: for accessibility should this maybe be a `button` instead?

src/design-system/ModalDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+  height: 20px;
+  padding: 0 1em;
+  float: right;
+`;
+
+interface Props {
+  numSteps?: number;
+  title?: string;
+  open?: boolean;
+  onClick: (e: React.MouseEvent<HTMLElement>) => void;
```
macfarlandian, line 57: I think maybe a more descriptive name like `closeModal` or whatever would help here

src/design-system/Modal.tsx
daschi: I'm not exactly sure what that would look like, when you have time maybe we could have a quick chat about it?

src/design-system/InputButton.tsx
daschi: I am not sure! I'm still learning about Typescript 😄

I was trying to find a way to make this button's style more extendable without having to define every single property with a ternary expression, and when I first implemented this without the `as InterpolationFunction<StyleProps` I saw the following error:

```
Argument of type '(props: StyleProps) => React.CSSProperties' 
is not assignable to parameter of type 'Interpolation
 <ThemedStyledProps
 <Pick<DetailedHTMLProps
 <ButtonHTMLAttributes<HTMLButtonElement>, ...
```

Have you seen an error similar to that before? I assume it might mean that the interpolated `props` in this styled-component expects to be one of those types listed, instead of `React.CSSProperties`, but I wasn't sure. I researched until I found someone doing something similar and noticed that `as InterpolationFunction<Props>` example. 

Is there another way you've found to extend styled-components styles? Would it be preferable to include the style like this?

```
const StyledButton = styled.button<StyleProps>`
  background: #00615c;
  font-size: 16px;
  border-radius: 12px;
  color: white;
  font-family: "Poppins", sans-serif;
  height: 48px;
  width: ${props => props.width ? props.width : "200px"};
  outline: none;
`;
```

src/design-system/InputButton.tsx
```diff
@@ -1,24 +1,30 @@
...
+const StyledButton = styled.button<Props>`
   background: #00615c;
   font-size: 16px;
   border-radius: 12px;
   color: white;
   font-family: "Poppins", sans-serif;
   height: 48px;
-  width: 200px;
+  width: ${(props) =>
+    props.styles && props.styles.width ? props.styles.width : "200px"};
```
macfarlandian, line 12: I _think_ there is some neat TypeScript shorthand you can use here to bail out of `props.styles.width` if `styles` is undefined: `props.styles?.width` should evaluate to `undefined` in that case (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

src/design-system/Modal.tsx
```diff
@@ -0,0 +1,43 @@
...
+  color: ${Colors.green};
+  cursor: pointer;
+  font-family: 'Libre Baskerville', serif;
+  font-size: 32px;
+  line-height: 32px
+  letter-spacing: -0.03em
+`;
+
+interface Props {
+  modalTitle?: TitleProps["title"];
```
macfarlandian, line 20: does `TitleProps.title` work here?

daschi: It doesn't, I originally tried that and the linter suggested using `TitleProps["title"]`

src/design-system/InputButton.tsx
daschi: cool!

PR #107, opened 2020-04-16
Model output inspection page

macfarlandian:
## Description of the change

This adds a new auth-protected, unlisted page that lets a user provide the same inputs as on the regular model page but displays much more detailed output: all of the projected curves as well as a (huge, scrollable) table of the full 3D projection array, including population inputs and an automatic validity check. This is intended to allow detailed inspection similar to what can be done in the model spreadsheet; the aim is that we can use it during ongoing model feature development to tune our changes now that we no longer can rely on working spreadsheet or Python implementations as a backstop. 

For the sake of expediency there is some copypasta from the `impact-dashboard` directory, particularly with the layout and chart components; it's not really important that this stays up to date with most of the latest changes there (if anything it is probably preferable that it doesn't) so I didn't bother trying to refactor the copied items for reuse. The notable thing that is shared is the set of user input forms, so that at least this page will automatically stay up to date as those are expanded. It makes this PR larger but it should be safe to gloss over all that stuff and focus on the changes in `infection-model` and the new `ModelInspectionTable` stuff.

There is a fair amount of change to the model code to expose all the additional data required here, but the interface to the visualizations on the overview page remains unchanged. In a way it's cleaner because the data is now always in a single 3D array until the point at which it is displayed, rather than taking the form of various intermediate data structures.

The new table features a fixed header row and column; seemed necessary due to the size, but required some new libraries to make it work reasonably. 

Screenshot_2020-04-15 Inspect epidemic model output • COVID-19 Dashboard

## Type of change

## Related issues

none

### Development

### Code review


macfarlandian:
@jessex i don't have access to our Zeit Now dashboard so I can't see why these deployments are failing ... no build errors on my end

macfarlandian:
nevermind I realized it was the same issue you were talking about earlier

jessex:
@macfarlandian, you should have an invite to now.sh now.

src/infection-model/seir.tsx
```diff
@@ -271,31 +284,24 @@ function getCurveProjections(inputs: SimulationInputs & CurveProjectionInputs) {
...
-    currentDayStaffProjections: initialStaffProjections,
-  } = getTotalsBreakdown(singleDayState);
-
-  const dailyIncarceratedProjections = [...initialIncarceratedProjections];
-  const dailyStaffProjections = [...initialStaffProjections];
+  updateProjectionDay(0, singleDayState);
 
-  let day = 0;
-  while (day < numDays) {
+  let day = 1;
```
jessex, line 303: There are _a lot_ of change-by-one edits throughout this PR, which makes me wonder what validation we've done to ensure no regression? In particular, I know you just fixed what was basically a glorified off-by-one issue with the model this week so I want to make sure we didn't regress.

macfarlandian: yes ironically some of that is because I realized I actually missed some off-by-one errors in that previous PR ... I am going to do another round of manual comparison with the current `master` version with a clear head today to double-check all this again

src/infection-model/seir.tsx
```diff
@@ -225,6 +216,27 @@ function getCurveProjections(inputs: SimulationInputs & CurveProjectionInputs) {
     rateOfSpreadFactor,
   } = inputs;
 
+  // 3d array. D1 = SEIR compartment. D2 = day. D3 = age bracket
+  const projectionGrid = ndarray(
+    new Array(seirIndexList.length * numDays + 1 * ageGroupIndex.__length).fill(
```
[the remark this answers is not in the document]
macfarlandian: yeah good catch ... luckily it didn't really affect anything though, because all these values are supposed to get overwritten anyway. this is just a backstop to prevent any `undefined`s from sneaking into the arithmetic

macfarlandian:
@jessex ok so two things: 

- in terms of verification, I compared the overview page on this branch with current master and verified that the values in the curves and the table are the same for various inputs
- I took the issue of off-by-one changes to heart and decided to reduce some complexity by having the final projection array contain exactly `numDays` days of data, rather than `numDays + 1` to include the current day. I don't think it really helps this PR be less of a beast (sorry) but it does further reduce the surface area for future off-by-one errors.

PR #23, opened 2020-04-03
Add legend and clean up chart

macfarlandian:
## Description of the change

Creates components to represent the chart legend and tooltip. Tweaks type and color styles to better match the mocks.

Screen Shot 2020-04-06 at 9 50 22 AM

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #110, opened 2020-04-16
Ahmed/develop

AhmedMugtaba:
## Description of the change

Updates the main output graph legend to allow toggling specific curves. 

Screen Shot 2020-04-16 at 6 00 57 PM

Ongoing things to resolve:
- CurveData values are not calculated on page load.

Screen Shot 2020-04-16 at 4 26 15 PM

Screen Shot 2020-04-16 at 3 54 07 PM

## Type of change

- [x] New feature (interactive legend. when legend button cliked curveData will toggle certain metric like exposed, hospitalized, etc)

## Related issues

Closes #23 

### Development

### Code review


src/hooks/useToggle.tsx
```diff
@@ -0,0 +1,13 @@
...
+  const [state, setState] = useState(true) as any;
+
+  const toggle = () => {
+    setState(!state);
+  };
+
+  return [state, toggle];
+}
+
+export default useToggle;
```
macfarlandian, line 13: you don't seem to actually be using this anywhere? if not needed please delete

src/impact-dashboard/ChartArea.tsx
```diff
@@ -25,22 +36,74 @@ const LegendContainer = styled.div`
...
+      filteredGroupStatus.reduce(
+        (data, key) => Object.assign(data, { [key]: curveData[key] }),
+        {},
+      ),
+    );
+  };
+
+  useEffect(() => {
+    updateDisplayData();
+  }, [groupStatus]);
```
macfarlandian, line 93: as we discussed earlier, it would be preferable for the data processing to remain inside `CurveChartContainer` for now. Why can't this logic run inside `CurveChartContainer` if you were to pass it `groupStatus` as a prop instead of the entire `curveData` structure?

src/impact-dashboard/CurveChartLegend.tsx
```diff
@@ -15,19 +15,49 @@ const LegendItem = styled.div`
...
     <LegendWrapper>
-      <LegendItem color={markColors.exposed}>exposed</LegendItem>
-      <LegendItem color={markColors.infectious}>infectious</LegendItem>
-      <LegendItem color={markColors.hospitalized}>hospitalized</LegendItem>
-      <LegendItem color={markColors.fatalities}>fatalities</LegendItem>
+      <LegendItem
+        onClick={() => toggleGroup("exposed")}
+        color={markColors.exposed}
+      >
+        {groupStatus["exposed"] ? "exposed" : "exposed"}
```
macfarlandian, line 41: this logic (and similar statements for the other items) isn't doing anything, you can just render the string directly

index.html
```diff
@@ -1,10 +1,10 @@
-<!--
+<!-- 
   generatePageContent.tsx looks for the following strings in this file:
   - "TITLE PLACEHOLDER"
   - "<style></style>"
   - "CONTENT PLACEHOLDER"
-  If you change any of these, make sure to edit generatePageContent.tsx as well.
- -->
+  If you change any of these, make sure to edit generatePageContent.tsx as well. -->
+
```
macfarlandian, line 7: this looks like junk, I would just revert this

src/page-overview/OverviewPage.tsx
```diff
@@ -1,3 +1,4 @@
+import { isIE } from "react-device-detect";
```
macfarlandian, line 1: I think it would be better if you just reverted the IE detection commit to get rid of these changes? that way we're not looking at two different features here

jessex:
Hey @AhmedMugtaba that PR description is a bit hard to follow. Would you mind taking another pass at updating it so it's focused on a specific task and references a specific issue? Feel free to take a look at other PRs in this repository to get a sense for what's expected here.

src/page-overview/OverviewPage.tsx
jessex: Agreed. For the record, @AhmedMugtaba, we want each PR to be focused on a single task, and we don't want to include work that we've already decided we're no longer pursuing. You should create a new branch each time you begin work on a new feature or task, as opposed to one large branch that you work on across all of your features.

src/impact-dashboard/CurveChartLegend.tsx
jessex: What's the idea here, @AhmedMugtaba? Why have a ternary operator that has the same value on both sides?

src/impact-dashboard/CurveChart.tsx
```diff
@@ -137,6 +137,7 @@ const CurveChart: React.FC<CurveChartProps> = ({
   curveData,
   hospitalBeds,
   markColors,
+  // toggled,
```
jessex, line 140: We should remove this before merging, unless we actually need it.

src/impact-dashboard/ChartArea.tsx
```diff
@@ -25,22 +36,74 @@ const LegendContainer = styled.div`
   flex: 0 0 auto;
 `;
 
-const ChartArea: React.FC = () => {
+function combinePopulations(data: CurveData, columnIndex: number) {
```
jessex, line 39: This point might be moot if you follow @macfarlandian's advice to move the calculation logic back to where it was previously. But if it does get moved out for some reason, we should try to not have two identical definitions of this function in two separate places.

src/impact-dashboard/ChartArea.tsx
AhmedMugtaba: I think it's better for the parent component to hold the data, makes it a lot easier to share data between different children, instead of sibling communication, I put this data initially on it's own context. but as per our discussion, that was overkill. but you're are totally right it may seem like not needed at the moment but it will things a lot easier going forward.

src/impact-dashboard/CurveChartLegend.tsx
AhmedMugtaba: This was meant to be a place holder to implement Active/Inactive functionality on legend when the user toggled, will update today.

index.html
AhmedMugtaba: Agree, will update.

src/page-overview/OverviewPage.tsx
AhmedMugtaba: Got it, will update.

src/impact-dashboard/ChartArea.tsx
AhmedMugtaba: this function is not needed here, just forgot to delete it, but I think it's better for the parent component to hold the data, makes it a lot easier to share data between different children, instead of sibling communication, I put this data initially on it's own context. but as per my discussion with Ian, it may seem like not needed at the moment but it will make things a lot easier going forward.

src/impact-dashboard/CurveChart.tsx
AhmedMugtaba: Agree, will update.

src/hooks/useToggle.tsx
AhmedMugtaba: this left here as a handy toggle hook, in case we needed to use it in other parts of the App, but will delete and update again when needed.

src/impact-dashboard/ChartArea.tsx
macfarlandian: ok but, the curve data isn't being shared between siblings. the only thing that's being shared is the toggle status. The chart should own the curve data because it is specific to what the chart displays. 

There is an argument to be made that the full projection data should be shared to avoid doing those calculations twice — in fact I left some TODOs to that effect so we could think about coming back to clean that up later — but that's not what you're doing here. IMO this is just unnecessarily creating a tight coupling between the chart component and its parent.

AhmedMugtaba: Alright, will update.

macfarlandian:
hey @AhmedMugtaba this is very close! I think I have figured out why the curves are not working on initial page load; I'm going to just merge this into a temporary branch and take a stab at fixing it while you move on to your next ticket.

PR #113, opened 2020-04-16
Multi-Facility Page Layout Markup

sprad:
"Merging" the work I've already done with frontend the work that is in progress.

I had a bit of a head start on doing some o the layout and styles for the multi-facility page.  Now that @daschi and @jovergaag are helping out as well, I wanted to make sure that I got the work I had already completed "merged" in so that we don't step on each others' toes.

Screen Shot 2020-04-16 at 3 57 28 PM

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

Closes []

### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #117, opened 2020-04-16
Adds the Ability to Read and Visualize Model Data from Cloud Firestore

### LINKED ISSUE #114: Read-only Multi-Facility Page Projections
Update the multi-facility page to include projection charts that read their data from Cloud Firestore

sprad:
## Description of the change

* Adding in a mechanism to query Firestore to grab a user's baseline scenario document and its underlying facilities.
* Passing each of the facility's model inputs to our projection charting code.

Screen Shot 2020-04-16 at 5 51 31 PM

## Type of change

## Related issues

Closes #114

### Development

### Code review

### DIFF
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -6,4 +6,10 @@ service cloud.firestore {
       allow read, write: if request.auth.uid == userId
     }
   }
+
+  match /databases/{database}/documents {
+    match /scenarios/{documents=**} {
+      allow read, write: if request.auth.uid != null;
+    }
+  }
 }
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -150,7 +150,7 @@ export const getFacilities = async (): Promise<Array<Facility> | null> => {
     if (!baselineScenario) return null;
 
     const facilitiesResults = await baselineScenario
-      .collection("facilities")
+      .collection(facilitiesCollectionId)
       .get();
 
     const facilities = facilitiesResults.docs.map(
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -8,12 +8,15 @@ import createAuth0Client from "@auth0/auth0-spa-js";
 
 import config from "../auth/auth_config.json";
 import { EpidemicModelPersistent } from "../impact-dashboard/EpidemicModelContext";
+import { Facility } from "../page-multi-facility/types";
 import { prepareForStorage, prepareFromStorage } from "./utils";
 
 // As long as there is just one Auth0 config, this endpoint will work with any environment (local, prod, etc.).
 const tokenExchangeEndpoint =
   "https://us-central1-c19-backend.cloudfunctions.net/getFirebaseToken";
 const modelInputsCollectionId = "model_inputs";
+const scenariosCollectionId = "scenarios";
+const facilitiesCollectionId = "facilities";
 
 // Note: None of these are secrets.
 let firebaseConfig = {
@@ -58,6 +61,10 @@ const authenticate = async () => {
   await firebase.auth().signInWithCustomToken(customToken);
 };
 
+const currrentUserId = () => {
+  return (firebase.auth().currentUser || {}).uid;
+};
+
 const getInputModelsDocRef = async () => {
   await authenticate();
 
@@ -66,9 +73,7 @@ const getInputModelsDocRef = async () => {
   }
 
   const db = firebase.firestore();
-  return db
-    .collection(modelInputsCollectionId)
-    .doc((firebase.auth().currentUser || {}).uid);
+  return db.collection(modelInputsCollectionId).doc(currrentUserId());
 };
 
 // TODO: Guard against the possibility of autosaves completing out of order.
@@ -107,6 +112,55 @@ export const getSavedState = async (): Promise<EpidemicModelPersistent | null> =
     console.error(
       "Encountered error while attempting to retrieve saved state:",
     );
+    console.error(error);
+
+    return null;
+  }
+};
+
+const getBaselineScenario = async () => {
+  await authenticate();
+
+  if (!firebase.auth().currentUser) {
+    throw new Error("Firebase user unexpectedly not set");
+  }
+
+  const db = firebase.firestore();
+  const query = db
+    .collection(scenariosCollectionId)
+    .where("baseline", "==", true);
+
+  const results = await query.get();
+  const userId = currrentUserId();
+
+  const baselineScenario = results.docs.find((doc) => {
+    const scenario = doc.data();
+    return userId && scenario.roles[userId] == "owner";
+  });
+
+  if (!baselineScenario) return null;
+
+  return baselineScenario.ref;
+};
+
+export const getFacilities = async (): Promise<Array<Facility> | null> => {
+  try {
+    const baselineScenario = await getBaselineScenario();
+
+    if (!baselineScenario) return null;
+
+    const facilitiesResults = await baselineScenario
+      .collection("facilities")
+      .get();
+
+    const facilities = facilitiesResults.docs.map(
+      (doc) => doc.data() as Facility,
+    );
+
+    return facilities;
+  } catch (error) {
+    console.error("Encountered error while attempting to retrieve facilities:");
+
     console.error(error);
 
     return null;
--- a/src/impact-dashboard/CurveChartContainer.tsx
+++ b/src/impact-dashboard/CurveChartContainer.tsx
@@ -26,7 +26,11 @@ function combinePopulations(data: CurveData, columnIndex: number) {
     getAllValues(getColView(data.staff, columnIndex)),
   ).map(([incarcerated, staff]) => incarcerated + staff);
 }
-const CurveChartContainer: React.FC<Props> = ({ markColors, groupStatus }) => {
+const CurveChartContainer: React.FC<Props> = ({
+  markColors,
+  groupStatus,
+  chartHeight,
+}) => {
   const modelData = useEpidemicModelState();
   const [curveData, updateCurveData] = useState({} as ChartData);
   const [curveDataFiltered, setCurveDataFiltered] = useState({} as ChartData);
@@ -60,6 +64,7 @@ const CurveChartContainer: React.FC<Props> = ({ markColors, groupStatus }) => {
     <Loading />
   ) : (
     <CurveChart
+      chartHeight={chartHeight}
       curveData={curveDataFiltered}
       hospitalBeds={modelData.hospitalBeds}
       markColors={markColors}
--- a/src/impact-dashboard/CurveChartContainer.tsx
+++ b/src/impact-dashboard/CurveChartContainer.tsx
@@ -10,6 +10,7 @@ import CurveChart from "./CurveChart";
 import { useEpidemicModelState } from "./EpidemicModelContext";
 
 interface Props {
+  chartHeight?: number;
   markColors: MarkColors;
   groupStatus: Record<string, any>;
 }
--- a/src/page-multi-facility/FacilityRow.tsx
+++ b/src/page-multi-facility/FacilityRow.tsx
@@ -1,23 +1,24 @@
-import styled from "styled-components";
+import React, { useContext } from "react";
 
-import Colors from "../design-system/Colors";
-import Loading from "../design-system/Loading";
+import { MarkColors as markColors } from "../design-system/Colors";
 import CurveChartContainer from "../impact-dashboard/CurveChartContainer";
-import { Facility } from "./types";
+import { EpidemicModelProvider } from "../impact-dashboard/EpidemicModelContext";
+import { FacilityContext } from "./FacilityContext";
 
-const markColors = {
-  exposed: Colors.green,
-  fatalities: Colors.black,
-  hospitalized: Colors.lightBlue,
-  hospitalBeds: Colors.red,
-  infectious: Colors.red,
+const groupStatus = {
+  exposed: true,
+  fatalities: true,
+  hospitalized: true,
+  infectious: true,
 };
 
-interface Props {
-  facility: Facility;
-}
+const FacilityRow: React.FC = () => {
+  const facility = useContext(FacilityContext);
 
-const FacilityRow: React.FC<Props> = ({ facility }) => {
+  if (!facility) {
+    throw new Error("Facility must be provided to the FacilityContext");
+  }
+
   const { name, modelInputs } = facility;
 
   const confirmedCases = modelInputs.confirmedCases;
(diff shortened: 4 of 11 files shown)


sprad:
I'm going to apologetically (cowardly?) caveat this PR with the statement that I'm primarily a backend developer who has never used TypeScript or Cloud Firestore and whose only React experience pre-dates Hooks.  That said, I hope I didn't do anything too egregious.

To get things to work for this PR, I more or less poked holes into the `CurveChart` and `CurveChartContainer` such that I could pass them relevant properties to get the visualization to display.  To be clear, this PR can not be merged into master as it stands unless we copy these files into the page-multi-facility directory and revert the changes to the original files in the "impact-dashboard" directory.  My understanding is that the current "Impact Dashboard" is going to be replaced entirely with the multi-facility page so copying these files actually shouldn't be a huge deal in the long-term.

I think the biggest thing I need help with understanding is the best approach to state management and passing data around.  At present, we mainly use the `EpidemicModelContext` and its Provider to share data between components such that when something in the form changes, the projection charts respond accordingly.  This will still need to happen on the multi-facility page, but it will need to occur at the facility level.  Does this lead to creating something like a FacilityContext that wraps each of the rows displayed on the right side of the screen (see attached screenshot)?

backend/firestore.rules
```diff
@@ -6,4 +6,10 @@ service cloud.firestore {
       allow read, write: if request.auth.uid == userId
     }
   }
+
+  match /databases/{database}/documents {
+    match /scenarios/{documents=**} {
+      allow read, write: if request.auth.uid != null;
+    }
+  }
```
sprad, line 14: At the moment, anyone who's authenticated can access the scenarios.  This should be tightened up for sure, but probably shouldn't be a launch blocker.

src/database/index.tsx
```diff
@@ -8,12 +8,15 @@ import createAuth0Client from "@auth0/auth0-spa-js";
 
 import config from "../auth/auth_config.json";
 import { EpidemicModelPersistent } from "../impact-dashboard/EpidemicModelContext";
+import { Facility } from "../page-multi-facility/types";
```
sprad, line 11: For the time being, I just continued my work in the folder that was already in progress (i.e. "page-multi-facility"), but this folder should likely be renamed in the future to something like "multi-facility-impact-dashboard" so that it doesn't appear that backend code depends on frontend code.

src/database/index.tsx
```diff
@@ -112,3 +117,52 @@ export const getSavedState = async (): Promise<EpidemicModelPersistent | null> =
...
+  const query = db
+    .collection(scenariosCollectionId)
+    .where("baseline", "==", true);
+
+  const results = await query.get();
+  const userId = currrentUserId();
+
+  const baselineScenario = results.docs.find((doc) => {
+    const scenario = doc.data();
+    return userId && scenario.roles[userId] == "owner";
```
sprad, line 138: Ideally there'd be a way to include this filter in the actual query.

src/impact-dashboard/CurveChartContainer.tsx
```diff
@@ -6,10 +6,13 @@ import { getAllValues, getColView } from "../infection-model/matrixUtils";
...
 import { MarkColors } from "./ChartArea";
 import CurveChart from "./CurveChart";
-import { useEpidemicModelState } from "./EpidemicModelContext";
+import { EpidemicModelState } from "./EpidemicModelContext";
 
 interface Props {
+  chartHeight?: number;
   markColors: MarkColors;
+  // This isn't actually optional, but disabling it for PR purposes
+  modelData?: EpidemicModelState;
```
sprad, line 15: Passing `modelData` in as a property at the facility level now that it's no longer "global" for the entire page.

src/impact-dashboard/CurveChartContainer.tsx
```diff
@@ -20,8 +23,12 @@ function combinePopulations(data: CurveData, columnIndex: number) {
   ).map(([incarcerated, staff]) => incarcerated + staff);
 }
 
-const CurveChartContainer: React.FC<Props> = ({ markColors }) => {
-  const modelData = useEpidemicModelState();
+const CurveChartContainer: React.FC<Props> = ({ chartHeight, markColors, modelData }) => {
+  // Temporarily appeasing the linter
```
sprad, line 27: Obviously this needs to go, but first we have to figure out how to pass state around.  See the conversation here:  https://github.com/Recidiviz/covid19-dashboard/pull/117#issuecomment-614934105

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -0,0 +1,60 @@
+import styled from "styled-components";
+import Colors from "../design-system/Colors";
+import Loading from "../design-system/Loading";
+
+import CurveChartContainer from "../impact-dashboard/CurveChartContainer";
+
+import { Facility } from "./types"
+
+const markColors = {
```
sprad, line 9: Need to de-dup this

src/page-multi-facility/MultiFacilityPage.tsx
```diff
@@ -16,6 +23,21 @@ const MultiFacilityImpactDashboard = styled.main.attrs({
 })``;
 
 const MultiFacilityPage: React.FC = () => {
+  const [facilities, setFacilities] =
+    useState({ data: [] as Array<Facility>, loading: true });
```
sprad, line 27: I dig types and all, but I feel like I'm doing something wrong in this code given the verbosity of all of these `as Array<Facility>`

backend/firestore.rules
jessex: Let's just make sure to file issues for anything we know had to be done but defer until after the next release.

src/database/index.tsx
jessex: Making sure I understand: we're querying for _all_ baseline scenarios across all users, and then filtering for ones where the current user has a role and that role is owner. Is that right? `find` returns the first match, right? I wonder if we should perform some assertion that there are not multiple baseline scenarios that match this condition set.

src/database/index.tsx
```diff
@@ -112,3 +117,52 @@ export const getSavedState = async (): Promise<EpidemicModelPersistent | null> =
     return null;
   }
 };
+
+const getBaselineScenario = async () => {
```
jessex, line 121: I see this PR is marked as closing #114 but it looks like it only has support for a single scenario. Is that coming in a separate issue or did we move multiple scenarios out of this release?

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Props {
+  facility: Facility;
+}
+
+const FacilityRow: React.FC<Props> = ({ facility }) => {
+  const {
+    name,
+    model_inputs
+  } = facility
```
jessex, line 25: A few quick nits in this file: semi-colon here and on line 27, and missing newline at the end of the file.

src/page-multi-facility/MultiFacilityPage.tsx
```diff
@@ -1,10 +1,17 @@
...
+import Loading from "../design-system/Loading";
 import SiteHeader from "../site-header/SiteHeader";
+
+import { getFacilities } from "../database";
+
+import FacilityRow from "./FacilityRow";
 import ProjectionsHeader from "./ProjectionsHeader";
 import ScenarioSidebar from "./ScenarioSidebar";
 
+import { Facility  } from "./types"
```
jessex, line 13: Nit: extra space before closing curly

src/impact-dashboard/CurveChartContainer.tsx
jessex: I actually think your proposal/wondering in the last paragraph of that comment sounds right, and I think you're on the right path as-is. Having a `FacilityContext` wrapping each facility on the page, each containing an `EpidemicModelContext` that maintains its own state that is being persisted to the backend via the `FacilityContext` feels clean. There would be a list of `FacilityContexts` in the page, ordered by whatever our ordering scheme is (facility name?), and when we go through the flow to add a new facility, we would initialize a new `FacilityContext` with a new `EpidemicModelContext` that has default parameters auto-populated.

Does that sound reasonable to you, @sprad and @macfarlandian?

src/impact-dashboard/CurveChartContainer.tsx
macfarlandian: I think when this shakes out we will still want one `EpidemicModelContext` per facility (and one chart per facility) so in the end you shouldn't have to do this and can revert to the previous data fetching strategy.

src/impact-dashboard/CurveChartContainer.tsx
macfarlandian: I think that's on the right track ... one thing that comes to mind is that `EpidemicModelContext` will have to stop managing the URL state, because when multiples are on the page they will start fighting over it.

src/page-multi-facility/FacilityRow.tsx
sprad: @jessex - Also, even though we propose `model_inputs` to be under_scored instead of camelCased in the data model design document, I'd like to suggest we use camelCase everywhere to more closely match JSON and JS naming conventions

src/page-multi-facility/MultiFacilityPage.tsx
jessex: :eagle:

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+              Last update: on March 25, 2020
+            </div>
+            <div className="mr-8">
+              <a className="px-1" href="#">Delete</a>
+              <a className="px-1" href="#">Edit</a>
+              <a className="px-1" href="#">Share</a>
+            </div>
+          </div>
+        </div>
+        <div className="w-3/5">
```
macfarlandian, line 52: generally speaking we are trying to move away from using these Tailwind css utilities in favor of `styled-components` but that's not a launch blocker IMO. depends on your appetite for writing a bunch of CSS here

src/page-multi-facility/FacilityRow.tsx
jessex: I'm a okay with that. I vaguely thought @itsjon or someone else had chosen that naming convention for some annoying reason but it's not coming to me right now. If we can go to camelCase, let's do it.

src/page-multi-facility/MultiFacilityPage.tsx
macfarlandian: this one is annoying but correct as far as I can tell. you should probably be either guarding against `null` on L31 or accepting `null` here rather than asserting that `getFacilities()` will always return `Array<Facility>`, since we know that to be incorrect.

To save characters you could also just `export type Facilities = Facility[]` from your facility types file?

src/page-multi-facility/FacilityRow.tsx
macfarlandian: I recommend you install a Prettier extension in your text editor of choice, if there is one, unless you enjoy getting comments such as these 😉

src/impact-dashboard/CurveChartContainer.tsx
sprad: That's actually great news.  I'll try to wire things up accordingly

src/page-multi-facility/FacilityRow.tsx
sprad: I crave attention...

src/page-multi-facility/FacilityRow.tsx
sprad: Not much of an appetite at the moment, but as an FYI, I did find a clever(?) way to use Tailwind utilities inside of styled-components which may help us ease into a full-on transition to styled components in the future:  https://github.com/Recidiviz/covid19-dashboard/pull/113/files#r409815530

src/page-multi-facility/MultiFacilityPage.tsx
sprad: `export type Facilities = Facility[]` is a neat trick ✅

src/page-multi-facility/FacilityRow.tsx
macfarlandian: oh yeah that's cool

src/impact-dashboard/CurveChartContainer.tsx
sprad: Ok, this approach makes sense to me.  I don't know a ton about Contexts/Providers, etc., but I'll start diving in...

src/page-multi-facility/FacilityRow.tsx
jessex: Yeah that kind of refactor is definitely not release-blocking. We can file issues for tech debt tasks if we feel strongly about something.

src/impact-dashboard/CurveChartContainer.tsx
jessex: @macfarlandian great point. I think with this change we probably _have_ to get rid of the URL parameter updates. It's possible that generating a URL with those parameters that can be shared would still be a valid use case, but I'm not sure if @cawarren has spoken on this topic.

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -19,42 +18,43 @@ interface Props {
...
-              {name}
-            </div>
+            <div className="w-1/4 text-red-600 font-bold">{confirmedCases}</div>
+            <div className="w-3/4 font-bold">{name}</div>
           </div>
           <div className="text-xs text-gray-500 pb-4 flex flex-row justify-between">
-            <div>
-              Last update: on March 25, 2020
-            </div>
+            <div>Last update: on March 25, 2020</div>
```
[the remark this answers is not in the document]
sprad: This is meant to represent when the facility was last updated.  The scenarios also have an updated date associated with them as seen in the [mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=910%3A130)

sprad:
@jessex - Ok, this latest [commit](https://github.com/Recidiviz/covid19-dashboard/pull/117/commits/3dd06f51971fbaf785a054412982f6cc1a0d596b) gets us to what I believe is a mergeable state.

I'm defining "meragable" to mean that the existing code all still works and the new multi-facility page loads as expected.

That said, I'm not overly crazy about the way I ended up re-using the `EpidemicModelContext` (i.e. by passing in a `facilityModel`), but this approach allows us to move forward without having to do major surgery right now.  However, I do _think_ we're going to create a new, more global `Context` that wraps the entire page and is responsible for loading the Google Doc's locale information.  Right now, that is bundled in with the current `EpidemicModelContext` which feels off in the context of supporting multiple facilities.

src/page-multi-facility/MultiFacilityPage.tsx
```diff
@@ -27,7 +51,17 @@ const MultiFacilityPage: React.FC = () => {
             <div className="flex flex-col flex-1 pb-6 pl-8">
               <AddFacilityModal />
               <ProjectionsHeader />
-              <div>Projections will go here</div>
+              {facilities.loading ? (
+                <Loading />
+              ) : (
+                facilities.data.map((facility, index) => {
```
macfarlandian, line 57: there's still the possibility that `facilities.data` will be null here .... I recommend using `facilities.data?.map` as a guard against that (it should just return null instead of proceeding to try and call `.map()`)

sprad: Good catch!  Fixed it locally.  We'll squash my commits and merge now.

PR #133, opened 2020-04-17
Separate locale data from EpidemicModelContext

### LINKED ISSUE #122: Create a new LocaleContext to Serve Locale Data Globally
At the moment, loading locale data is intertwined within the EpidemicModelContext.  Since we are moving to a world where we support multiple facilities, we will need to move the logic that loads locale data from the Google Spreadsheet out of the EpidemicModelContext and into its own LocaleContext (or whatever name makes the most sense).  This new LocaleContext will need to wrap the whole Multi-facility Dashboard page such that locale data is available the facility add/edit forms.

macfarlandian:
## Description of the change

This untangles the loading of locale data (from the external spreadsheet) from the provisioning of an `EpidemicModelContext` and moves it to a new `LocaleDataContext` that I have just made globally available for now because multiple pages need it. Marshaling the required data to display a single facility now requires three roughly hierarchical contexts to be present: `LocaleDataContext`, `FacilityContext`, and `EpidemicModelContext`.  

`EpidemicModelContext` is much cleaner now because it no longer has to deal with the loading state of the locale data — it requires a source of locale data to be passed in when provisioned. It also does not bother with auto-saving or URL state management anymore, because we are no longer doing those things.

In terms of regression, I made sure that the existing Overview page as well as the Model Inspection page still work with this new data loading strategy. Not having any actual facility data and not being sure how to populate it, I haven't actually tested that `FacilityRow` behaves as expected now.

Something not strictly in scope but related enough that I decided to fix it (rather than reimplement it with the same mistake) was the conflation of "confirmed cases" with "estimated cases among the incarcerated", which @cawarren pointed out to me earlier this week. I assumed that the "cases" value we display for a facility should be the number of user-input cases at the facility, not the number of confirmed cases state- or county-wide, so I updated that at the same time as well. 

## Type of change

## Related issues

Closes #122 

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/CurveChartContainer.tsx
+++ b/src/impact-dashboard/CurveChartContainer.tsx
@@ -1,4 +1,5 @@
 import { zip } from "d3-array";
+import isEmpty from "lodash/isEmpty";
 import { useEffect, useState } from "react";
 
 import Loading from "../design-system/Loading";
@@ -48,6 +49,11 @@ const CurveChartContainer: React.FC<Props> = ({
   }, [modelData]);
 
   useEffect(() => {
+    if (isEmpty(curveData)) {
+      setCurveDataFiltered({});
+      return;
+    }
+
     let filteredGroupStatus = Object.keys(groupStatus).filter(
       (groupName) => groupStatus[groupName],
     );
@@ -60,7 +66,7 @@ const CurveChartContainer: React.FC<Props> = ({
     );
   }, [groupStatus, curveData]);
 
-  return modelData.countyLevelDataLoading ? (
+  return !curveDataFiltered ? (
     <Loading />
   ) : (
     <CurveChart
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -1,22 +1,6 @@
-import { csvParse, DSVRowAny } from "d3";
-import { rollup } from "d3-array";
-import parseJSON from "date-fns/parseJSON";
-import { mapValues, pick, size } from "lodash";
-import numeral from "numeral";
-import React, { useEffect } from "react";
+import React from "react";
 
-import { getSavedState, saveState } from "../database";
-import useQueryParams, { QueryParams } from "../hooks/useQueryParams";
-
-type CountyLevelRecord = {
-  hospitalBeds: number;
-  totalIncarceratedPopulation: number;
-  estimatedIncarceratedCases: number;
-  county: string;
-  state: string;
-};
-
-type CountyLevelData = Map<string, Map<string, CountyLevelRecord>>;
+import { LocaleData } from "../locale-data-context";
 
 export type PlannedRelease = { date?: Date; count?: number };
 export type PlannedReleases = PlannedRelease[];
@@ -80,10 +64,6 @@ interface MetadataPersistent {
 }
 
 interface MetadataUpdate extends MetadataPersistent {
-  countyLevelData?: CountyLevelData;
-  countyLevelDataLoading?: boolean;
-  countyLevelDataFailed?: boolean;
-  stateInitialized?: boolean;
   // this is not user input so don't store it
   hospitalBeds?: number;
 }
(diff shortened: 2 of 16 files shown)


src/impact-dashboard/CurveChartContainer.tsx
```diff
@@ -60,7 +66,7 @@ const CurveChartContainer: React.FC<Props> = ({
     );
   }, [groupStatus, curveData]);
 
-  return modelData.countyLevelDataLoading ? (
+  return !curveDataFiltered ? (
```
sprad, line 69: Much better 🙌 .  Using `countyLevelDataLoading` as a proxy for data loading in the chart was bothering me now that data loading is at the facility level.

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -1,22 +1,8 @@
...
-import parseJSON from "date-fns/parseJSON";
-import { mapValues, pick, size } from "lodash";
-import numeral from "numeral";
-import React, { useEffect } from "react";
+import { sum } from "d3";
+import { pick } from "lodash";
+import React from "react";
 
-import { getSavedState, saveState } from "../database";
-import useQueryParams, { QueryParams } from "../hooks/useQueryParams";
```
sprad, line 9: We can probably delete this `/hooks/useQueryParams` file as well

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -201,216 +188,44 @@ function epidemicModelReducer(
...
-          const nestedStateCounty = rollup(
-            parsedArray,
-            // there will only ever be one row object per county
-            (v: object[]) => v[0],
-            (d: DSVRowAny) => d.state as string,
-            (d: DSVRowAny) => d.county as string,
-            // some wrong/outdated typedefs for d3 are making typescript sad
-            // but this should check out
-          ) as CountyLevelData;
-
```
sprad, line 346: The first `useEffect` call and lines 347 - 413 were all deeply confusing to me so I'm really glad to see they're gone 🥳

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -201,216 +188,44 @@ function epidemicModelReducer(
...
-  } = useQueryParams({});
+  const resetBase = getResetBase();
+  const initialState = {
+    ...resetBase,
+    ...getLocaleData(
+      localeDataSource,
+      resetBase.stateCode,
+      resetBase.countyName,
+    ),
+    // any passed data (e.g. fetched from database) takes precedence
```
sprad, line 222: I dig it 👍

src/impact-dashboard/ImpactDashboard.tsx
```diff
@@ -62,7 +62,7 @@ const HorizontalDivider = styled.hr`
 `;
 
 const ImpactDashboard: React.FC = () => {
-  const { countyLevelDataFailed } = useEpidemicModelState();
+  const { failed: countyLevelDataFailed } = useLocaleDataState();
```
sprad, line 65: Sorry, newbie question:  What does it mean when we assign a key like `failed` here?  I don't see it being used anywhere so it's not clear to me why it's needed/useful?

src/locale-data-context/LocaleDataContext.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { rollup } from "d3-array";
+import numeral from "numeral";
+import React from "react";
+
+type LocaleRecord = {
+  county: string;
+  estimatedIncarceratedCases: number;
+  hospitalBeds: number;
+  reportedCases: number;
+  state: string;
```
sprad, line 11: Even though I don't really like it because we use proper state names, should use `stateCode` here to be consistent with how we use this naming elsewhere and so that we don't confuse it with React's concept of `state`?

src/impact-dashboard/ImpactDashboard.tsx
macfarlandian: sorry I didn't see these comments earlier for some reason. this is just renaming the variable from `failed` to `countyLevelDataFailed`, since in the context of this component `failed` seemed like too generic a variable name

src/locale-data-context/LocaleDataContext.tsx
macfarlandian: yeah good call out. what i really want to do is go through and change it to `stateName` everywhere but I have not gotten around to that. we can come back to that later though, i'll file an issue

src/impact-dashboard/ImpactDashboard.tsx
sprad: Cool, cool.  Thanks for the explanation

PR #136, opened 2020-04-18
Start wiring up the scenario sidebar

### LINKED ISSUE #131: Wire the Scenario Sidebar to Read From and Write To Firestore
Screen Shot 2020-04-17 at 3 07 19 PM

Populate the Scenario Sidebar with information from the Cloud Firestore.  You can get information about the baseline scenario from Firestore by using the following code:

```javascript
const baselineScenarioRef = await getBaselineScenarioRef();
const result = await baselineScenarioRef.get();
const scenario = result.data();
```

Note `getBaselineScenarioRef` can be imported from our [db code](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/index.tsx#L135).

For this mockup you'll need the `name`, `updatedAt`, `dailyReports`, and `dataSharing` attributes found on the `scenario` object.

As a final step, when the user toggles the `dailyReports`, and `dataSharing` toggles we should save these updates to Firestore by calling the `saveScenario` method that can also be imported from our [db code](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/index.tsx#L210).

```javavscript
  saveScenario({
    dataSharing: true,
  });
```

daschi:
## Description of the change

> Pass in the baseline scenario to the Scenario Sidebar component. I'm opening this as a WIP and having turned off the linter requirements so that it can be picked up by someone else later if needed. 

TODO: 

- [ ] Update the typescript types for the component props
- [x] Add `dailyReports` and `dataSharing` to the `scenario` model 
- [x] Add a data formatter for the `updatedAt` data 
- [x] Save the daily reports & data sharing booleans to the scenario

Question - What data should go in the description below the header?

## Type of change

## Related issues

Closes #131

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -223,7 +223,7 @@ export const saveScenario = async (scenario: {}): Promise<void> => {
       updatedAt: currrentTimestamp(),
     });
 
-    baselineScenarioRef.update(payload);
+    return await baselineScenarioRef.update(payload);
   } catch (error) {
     console.error("Encountered an error while saving the scenario:");
     console.error(error);
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -192,6 +192,8 @@ export const createBaselineScenario = async () => {
     baselineScenarioRef = await db.collection(scenariosCollectionId).add({
       name: "Baseline Scenario",
       baseline: true,
+      dataSharing: false,
+      dailyReports: false,
       roles: {
         [userId]: "owner",
       },
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -194,6 +194,8 @@ export const createBaselineScenario = async () => {
       baseline: true,
       dataSharing: false,
       dailyReports: false,
+      description:
+        "Welcome to your new scenario. To get started, add in facility data on the right-hand side of the page. Your initial scenario is also your 'Baseline' - meaning this is where you should keep real-world numbers about the current state of your facilities, their cases, and mitigation steps.",
       roles: {
         [userId]: "owner",
       },
--- a/src/page-multi-facility/MultiFacilityImpactDashboard.tsx
+++ b/src/page-multi-facility/MultiFacilityImpactDashboard.tsx
@@ -20,7 +20,11 @@ const MultiFacilityImpactDashboardContainer = styled.main.attrs({
   `,
 })``;
 
-const MultiFacilityImpactDashboard: React.FC = () => {
+interface MultiFacilityImpactDashboardType = {
+  baselineScenario?: object;
+}
+
+const MultiFacilityImpactDashboard<MultiFacilityImpactDashboardType>: React.FC = ({ baselineScenario }) => {
   const { data: localeDataSource } = useLocaleDataState();
 
   const [facilities, setFacilities] = useState({
@@ -28,7 +32,20 @@ const MultiFacilityImpactDashboard: React.FC = () => {
     loading: true,
   });
 
+  const [scenario, setScenario] = useState({
+    data: {},
+    loading: true
+  })
+
   useEffect(() => {
+    async function fetchScenario () {
+      const result = await baselineScenario.data.get();
+      setScenario({
+        data: result?.data(),
+        loading: false
+      })
+    }
+
     async function fetchFacilities() {
       const facilitiesData = await getFacilities();
 
@@ -39,12 +56,14 @@ const MultiFacilityImpactDashboard: React.FC = () => {
         });
       }
     }
+
+    fetchScenario();
     fetchFacilities();
   }, []);
 
   return (
     <MultiFacilityImpactDashboardContainer>
-      <ScenarioSidebar />
+      <ScenarioSidebar scenario={scenario?.data} />
       <div className="flex flex-col flex-1 pb-6 pl-8">
         <AddFacilityModal />
         <ProjectionsHeader />
--- a/src/page-multi-facility/MultiFacilityImpactDashboard.tsx
+++ b/src/page-multi-facility/MultiFacilityImpactDashboard.tsx
@@ -8,6 +8,7 @@ import { useLocaleDataState } from "../locale-data-context";
 import AddFacilityModal from "./AddFacilityModal";
 import { FacilityContext } from "./FacilityContext";
 import FacilityRow from "./FacilityRow";
+import { ScenarioType } from "./MultiFacilityPage";
 import ProjectionsHeader from "./ProjectionsHeader";
 import ScenarioSidebar from "./ScenarioSidebar";
 import { Facilities } from "./types";
@@ -20,11 +21,13 @@ const MultiFacilityImpactDashboardContainer = styled.main.attrs({
   `,
 })``;
 
-interface MultiFacilityImpactDashboardType = {
-  baselineScenario?: object;
+interface Props {
+  baselineScenario?: ScenarioType;
 }
 
-const MultiFacilityImpactDashboard<MultiFacilityImpactDashboardType>: React.FC = ({ baselineScenario }) => {
+const MultiFacilityImpactDashboard: React.FC<Props> = ({
+  baselineScenario,
+}) => {
   const { data: localeDataSource } = useLocaleDataState();
 
   const [facilities, setFacilities] = useState({
@@ -33,19 +36,22 @@ const MultiFacilityImpactDashboard<MultiFacilityImpactDashboardType>: React.FC =
   });
 
   const [scenario, setScenario] = useState({
-    data: {},
-    loading: true
-  })
+    data: null,
+    loading: true,
+  });
 
   useEffect(() => {
-    async function fetchScenario () {
-      const result = await baselineScenario.data.get();
+    async function fetchScenario() {
+      const result = await baselineScenario?.data?.get();
       setScenario({
         data: result?.data(),
-        loading: false
-      })
+        loading: false,
+      });
     }
 
+    fetchScenario();
+  }, []);
+  useEffect(() => {
     async function fetchFacilities() {
       const facilitiesData = await getFacilities();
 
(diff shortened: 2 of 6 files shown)


src/database/index.tsx
```diff
@@ -192,6 +192,8 @@ export const createBaselineScenario = async () => {
     baselineScenarioRef = await db.collection(scenariosCollectionId).add({
       name: "Baseline Scenario",
       baseline: true,
+      dataSharing: false,
+      dailyReports: false,
```
sprad, line 196: Thanks for adding these.  I've manually updated the existing scenarios in Firebase to set these values

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -1,11 +1,18 @@
 import PromoBoxWithButton from "../design-system/PromoBoxWithButton";
 import ToggleRow from "./ToggleRow";
+import { saveScenario } from "../database";
+
+interface ScenarioSidebar = {
+  scenario: Object;
```
sprad, line 6: Feel free to add a proper `Scenario` type in this file:  https://github.com/Recidiviz/covid19-dashboard/blob/master/src/page-multi-facility/types.tsx

sprad: Will look something like this:

```javascript
export type Scenario = {
  name: string;
  baseline: boolean;
  dataSharing?: boolean;
  dailyReports?: boolean;
  roles: Map<String, String>;
  facilities?: Facilities;
};
```

sprad: Will also include  `createdAt` and  `updatedAt` as required fields.  They're stored as `firebase.firestore.FieldValue.serverTimestamp();`, but I'm not sure what that translates to in the TypeScript/JavaScript Datetime world

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
```diff
@@ -20,15 +20,32 @@ const MultiFacilityImpactDashboardContainer = styled.main.attrs({
...
     data: [] as Facilities,
     loading: true,
   });
 
+  const [scenario, setScenario] = useState({
+    data: {},
+    loading: true
+  })
+
   useEffect(() => {
```
sprad, line 40: I don't know enough about React, but is each piece of managed state supposed to get its own `useEffect` or is it common to nest multiple calls under a single `useEffect` like we're doing here?

src/page-multi-facility/MultiFacilityPage.tsx
```diff
@@ -10,19 +10,25 @@ import MultiFacilityImpactDashboard from "./MultiFacilityImpactDashboard";
 
 const MultiFacilityPageDiv = styled.div``;
 
+type ScenarioType = {
+  data: object | null;
+  lodaing: boolean;
```
sprad, line 15: Nit:  "lodaing" -> "loading"

src/page-multi-facility/ScenarioSidebar.tsx
daschi: Okay I can add these!

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
jessex: I think the way to think of it is as discrete operations. If you have two things you want to happen independently of each other, they should be their own hooks. If you have two things that you want to happen in tandem or in a particular order, they may live in the same hook.

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -15,15 +22,17 @@ const ScenarioSidebar: React.FC = () => {
           </p>
         </div>
         <div>
-          <p className="text-xs text-gray-500">Last Update: March 25, 2020</p>
+          <p className="text-xs text-gray-500">Last Update:</p>
```
sprad, line 25: I'm not 100% sure how Firestore returns date times so you'll likely need to do some formatting here.  The `DateFormats` file may be of service:  https://github.com/Recidiviz/covid19-dashboard/blob/94e5cafa080eeebc63aadd032f1a94643f8a403a/src/design-system/DateFormats.tsx

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
jessex: In this case, I think maybe the thinking was that loading one or the other but not both might be a poor experience so we should place it within a single hook?

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -15,15 +22,17 @@ const ScenarioSidebar: React.FC = () => {
...
         </div>
         <div className="mt-5 mb-5 border-b border-gray-300" />
         <div>
           <ToggleRow
+            data={dailyReports}
             label="Daily Reports"
             labelHelp="Tooltip help Lorem ipsum dolor sit amet, consectetur adipiscing elit"
```
sprad, line 32: We should get final copy for all of the "lorem ipsum" text in the side bar as well

jessex: I'm going to take a pass through and update all copy this weekend!

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
daschi: Ah okay--I actually just updated it to be 2 hooks before I saw this comment. I think will mean the facilities fetching will not depend on the scenario fetching to complete first, so I think they would happen independently from each other. If we want to ensure both are loaded before loading the content maybe that would be more of a conditional rendering?

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -1,11 +1,21 @@
...
+import { Scenario } from "./types";
+
+interface Props {
+  scenario?: Scenario | null;
+}
+
+const ScenarioSidebar: React.FC<Props> = (props) => {
+  // Typescript error that these properties do not exist on type Scenario
+  // Have I defined the type wrong?
+  const { name, dailyReports, dataSharing } = props.scenario;
```
daschi, line 13: I'm still having issues here with the typescript error saying that `Scenario` type does not have these properties.

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -12,6 +13,15 @@ const ScenarioSidebar: React.FC<Props> = (props) => {
   const { scenario } = props;
   const updatedAtDate = Number(scenario?.updatedAt.toDate());
 
+  const handleToggle = (toggleValue: keyof Scenario) => {
+    const value = scenario?.[toggleValue];
+    // TODO: I think the default value is still set to false for both dataSharing and dailyReports
```
jessex, line 18: Correct: for new scenarios these should default to / start out as false.

jessex:
@daschi, just noticed your question:

>Question - What data should go in the description below the header?

We need to add a description field to the scenario model and make that text editable from the frontend. If you're already in this realm and are comfortable making that addition then great! Otherwise, it can be deferred to another PR.

src/database/index.tsx
sprad, line 196: @daschi - We can add a `description` field and default text here. 
@jessex - Do we have copy we can include as a default description?

jessex: Yep! You can use this:
> Welcome to your new scenario. To get started, add in facility data on the right-hand side of the page. Your initial scenario is also your 'Baseline' - meaning this is where you should keep real-world numbers about the current state of your facilities, their cases, and mitigation steps.

I know this was also used in the fresh start modal but I think that may have been a misunderstanding. I'm going to make sure all copy is where it should be later this weekend, so you can use the above for now.

sprad: Whoops, accidentally hit Resolve.  Meant to add a comment to let you all know that I added a `description` to the existing Scenarios in firestore with some generic "Default description" copy.

jessex: Nice! So I think that means we should be unblocked to add description text reads and updates in the sidebar, @daschi.

src/page-multi-facility/types.tsx
```diff
@@ -6,3 +6,19 @@ export type Facility = {
...
+  name: string;
+  baseline: boolean;
+  dataSharing?: boolean;
+  dailyReports?: boolean;
+  roles: Map<string, string>;
+  createdAt: TimestampType;
+  updatedAt: TimestampType;
+};
+
+type TimestampType = {
```
sprad, line 20: I'd probably just call this `TimeStamp` unless there is a technical reason to add on the `Type` suffix.

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
jessex: I think it's likely a non-issue as currently constructed, but it should become apparent quickly in final testing of so. Let's leave it as is!

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -1,29 +1,60 @@
...
-            This text describes unique qualities for this data model. Taking
-            notes here will be useful when you have multiple data model
-            experiments in your library?
-          </p>
+          <InputTextArea
+            label="Description"
+            value={scenario?.description}
+            placeholder=""
+            onChange={(event) => {
+              // Question - Should we have a Save button to save the description?
```
jessex, line 32: I don't see a save button in the mocks, so I think it should just persist the description change automatically when finished.

itsjon:
@jessex thank you! For now I'm deferring most of the linting issues unless they're build-breaking.

daschi:
Happy to see it was merged, thanks @itsjon and @jessex!

PR #138, opened 2020-04-18
Add the Ability to Save Facilities

### LINKED ISSUE #127: Add the Ability to Save Facilities
Add the ability to save facilities as they are added or edited in the front end.  Note:  We can cheat for now and simply attach the Facility to the Baseline Scenario's `facilities` sub-collection.

sprad:
* Adding the ability to save and update a facility to Firestore.

## Type of change

## Related issues

Closes #127

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -5,9 +5,13 @@ import "firebase/auth";
 import "firebase/firestore";
 
 import createAuth0Client from "@auth0/auth0-spa-js";
+import { pick } from "lodash";
 
 import config from "../auth/auth_config.json";
-import { EpidemicModelPersistent } from "../impact-dashboard/EpidemicModelContext";
+import {
+  EpidemicModelPersistent,
+  persistedKeys,
+} from "../impact-dashboard/EpidemicModelContext";
 import { Facility } from "../page-multi-facility/types";
 import { prepareForStorage, prepareFromStorage } from "./utils";
 
@@ -85,6 +89,27 @@ const getDb = async () => {
   return firebase.firestore();
 };
 
+const buildCreatePayload = (entity: any) => {
+  const timestamp = currrentTimestamp();
+  return Object.assign({}, entity, {
+    createdAt: timestamp,
+    updatedAt: timestamp,
+  });
+};
+
+const buildUpdatePayload = (entity: any) => {
+  const timestamp = currrentTimestamp();
+  const payload = Object.assign({}, entity, {
+    updatedAt: timestamp,
+  });
+
+  // We don't want to actually store the id as a field. It should
+  // only be used as a Firestore document reference.
+  delete payload.id;
+
+  return payload;
+};
+
 const getInputModelsDocRef = async () => {
   const db = await getDb();
   return db.collection(modelInputsCollectionId).doc(currrentUserId());
@@ -152,30 +177,6 @@ export const getBaselineScenarioRef = async () => {
   return baselineScenario.ref;
 };
 
-export const getFacilities = async (): Promise<Array<Facility> | null> => {
-  try {
-    const baselineScenarioRef = await getBaselineScenarioRef();
-
-    if (!baselineScenarioRef) return null;
-
-    const facilitiesResults = await baselineScenarioRef
-      .collection(facilitiesCollectionId)
-      .get();
-
-    const facilities = facilitiesResults.docs.map(
-      (doc) => doc.data() as Facility,
-    );
-
-    return facilities;
-  } catch (error) {
-    console.error("Encountered error while attempting to retrieve facilities:");
-
-    console.error(error);
-
-    return null;
-  }
-};
-
 export const createBaselineScenario = async () => {
   try {
     let baselineScenarioRef = await getBaselineScenarioRef();
@@ -185,19 +186,20 @@ export const createBaselineScenario = async () => {
     }
 
     const userId = currrentUserId();
-    const timestamp = currrentTimestamp();
 
     const db = await getDb();
 
-    baselineScenarioRef = await db.collection(scenariosCollectionId).add({
+    const payload = buildCreatePayload({
       name: "Baseline Scenario",
       baseline: true,
       roles: {
         [userId]: "owner",
       },
-      createdAt: timestamp,
-      updatedAt: timestamp,
     });
+
+    baselineScenarioRef = await db
+      .collection(scenariosCollectionId)
+      .add(payload);
 
     return baselineScenarioRef;
   } catch (error) {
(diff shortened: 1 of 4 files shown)


sprad:
This is a WIP PR, but the code is functional in that it will save data to CloudStore.  The main things I want to address before this gets merged in are:

1.  Handling some sort of data validation before shipping things off to be persisted.  As it stands, this basically just lets you push anything into CloudStore.
2.  Doing a lot of testing to make sure that our JavaScript objects get serialized and deserialized property when they are coming to and from CS.  This is particularly true for nested attributes like `plannedReleases` and complex data types like Date/Times.

src/database/index.tsx
```diff
@@ -85,6 +85,27 @@ const getDb = async () => {
...
+};
+
+const buildUpdatePayload = (entity: any) => {
+  const timestamp = currrentTimestamp();
+  const payload = Object.assign({}, entity, {
+    updatedAt: timestamp,
+  });
+
+  // We don't want to actually store the id as a field. It should
+  // only be used as a CloudStore document reference.
```
jessex, line 103: I keep seeing CloudStore references in this PR. Do we mean _Firestore_ or is this a term I'm just not familiar with in the stack?

src/database/index.tsx
```diff
@@ -162,9 +183,11 @@ export const getFacilities = async (): Promise<Array<Facility> | null> => {
...
       .get();
 
-    const facilities = facilitiesResults.docs.map(
-      (doc) => doc.data() as Facility,
-    );
+    const facilities = facilitiesResults.docs.map((doc) => {
+      const facility = doc.data() as Facility;
+      facility.id = doc.id;
+      return facility;
+    });
```
sprad, line 190: Adding the CloudStore docRef `id` onto each facility here so that we know which to persist when an individual Facility is selected and needs to be updated

src/database/index.tsx
sprad: The full name is "Cloud Firestore", but we should just refer to it as "Firestore".  "Cloudstore" was an invention of my sleepy brain

src/database/index.tsx
```diff
@@ -152,30 +177,6 @@ export const getBaselineScenarioRef = async () => {
   return baselineScenario.ref;
 };
 
-export const getFacilities = async (): Promise<Array<Facility> | null> => {
```
sprad, line 155: Ignore:  I just moved this such that the `scenario` and `facility` related persistence methods were grouped together

src/database/index.tsx
```diff
@@ -217,13 +219,165 @@ export const saveScenario = async (scenario: {}): Promise<void> => {
...
+  try {
+    // Cheating for launch expediency.
+    // See: https://github.com/Recidiviz/covid19-dashboard/issues/129
+    const baselineScenarioRef = await getBaselineScenarioRef();
+
+    if (!baselineScenarioRef) return null;
+
+    const facilitiesResults = await baselineScenarioRef
+      .collection(facilitiesCollectionId)
+      .orderBy("name")
```
sprad, line 241: Added an `orderBy` here so that results are returned consistently ⬆️

src/database/index.tsx
```diff
@@ -217,13 +219,165 @@ export const saveScenario = async (scenario: {}): Promise<void> => {
...
+      .get();
+
+    const facilities = facilitiesResults.docs.map((doc) => {
+      const facility = doc.data() as Facility;
+      facility.id = doc.id;
```
sprad, line 246: Also tacking on the `docRef` id so that we can use it later for updates and deletions

src/database/index.tsx
```diff
@@ -217,13 +219,165 @@ export const saveScenario = async (scenario: {}): Promise<void> => {
...
+    return facilities;
+  } catch (error) {
+    console.error("Encountered error while attempting to retrieve facilities:");
+
+    console.error(error);
+
+    return null;
+  }
+};
+
```
sprad, line 259: Everything below this comment is new code

src/database/testcases.tsx
```diff
@@ -0,0 +1,196 @@
+import { deleteFacility, saveFacility } from "./index";
```
sprad, line 1: Note:  This file isn't meant to be merged in along with this PR, but it is probably the most important thing to review.  The code above is all relatively straightforward (I hope 🤞).  This is where I spent most of my time with this PR, creating test cases to give us confidence that things work as expected.  Please review the scenarios I tested and feel free to comment with more and I'll try them out locally.

src/database/testcases.tsx
```diff
@@ -0,0 +1,196 @@
...
+    age75Population: 25,
+    age65Population: 25,
+    age55Population: 25,
+    age45Population: 25,
+    age20Population: 25,
+    age0Population: 25,
+  },
+});
+
+// Verified to work as expected,  ✅
```
sprad, line 51: And `localeDataSource` was not persisted as desired

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -82,8 +78,8 @@ export type EpidemicModelPersistent = ModelInputsPersistent &
 // TODO: is there a smarter way to get these values?
 export const persistedKeys: Array<keyof EpidemicModelPersistent> = [
   "countyName",
-  "facilityName",
   "stateCode",
+  "hospitalBeds",
```
sprad, line 82: Turns out we do want to persist `hospitalBeds`.  If we don't, the app seems to be loading the "US Total" numbers and breaks the charts as follows:

Screen Shot 2020-04-18 at 10 11 26 AM

sprad:
Ok, this is no longer WIP and is ready for a proper review.  Even though we won't be committing it, the most important file to look at is actually `testcases.tsx.  It should give you a sense of what working with these newly added functions looks like.

Also, note that when I hack up the `ImpactDashboardContainer` component locally to use the newly added `getFacility` function and pass the returned `Facility`'s `modelInputs` into the [EpidemicModelProvider](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/impact-dashboard/ImpactDashboardContainer.tsx#L12), with one small exception (Confirmed Cases is off), things Just Work ™️  🎉 

Screen Shot 2020-04-18 at 11 49 58 AM

Note:  I had to zoom way out to capture the entire screen so that's why the proportions may look off.

Note 2:  I'd guess we could fix the confirmed cases issue by storing this information along with the model inputs

src/database/testcases.tsx
```diff
@@ -0,0 +1,196 @@
...
+    localeDataSource: {},
+    totalIncarcerated: 480,
+    ageUnknownPopulation: 305,
+    ageUnknownCases: 3,
+    confirmedCases: 451,
+    staffCases: 3,
+    staffPopulation: 37,
+    plannedReleases: [
+      {
+        date: "2020-04-24T04:00:00.000Z",
```
sprad, line 26: This is how these `date` values come in from the form.  They are stored as a `string` in Firestore which I'm not sure I'm crazy about, but as you can see from the screenshot attached with this PR's opening comment, the "Planned Releases" fields are updated properly.

src/database/index.tsx
```diff
@@ -217,13 +219,165 @@ export const saveScenario = async (scenario: {}): Promise<void> => {
...
+ *
+ * Instead, when saving model inputs, you'll need to pass in the full
+ * list of attributes that you are interested in saving as defined by
+ * the type EpidemicModelPersistent found in the EpidemicModelContext.
+ *
+ * This might seem annoying but in practice isn't that big of a deal
+ * because the submission from the web form will include all of the model
+ * input fields that the user wishes to store.
+ *
+ * Note: It is fine leave model input values out of modelInputs field
```
jessex, line 324: Nit: "It is fine to leave..." and "...that those _fields_..."

src/database/testcases.tsx
jessex: This is fantastic! I wouldn't delete this, actually. Even though it's not structured to run as unit tests, I would still move it into a `src/database/__tests__` directory and maybe add a comment at the top that this is a file for operating manual tests.

src/database/testcases.tsx
```diff
@@ -0,0 +1,196 @@
+import { deleteFacility, saveFacility } from "./index";
+
+// 1. Full Form Submission - New Facility
+//
+// Note: some fileds get submitted along with the form submission
```
jessex, line 5: Nit: fields

src/database/testcases.tsx
jessex: Maybe add this comment explicitly in your code here, if we follow my suggestion to leave this in the repository as test code.

src/database/testcases.tsx
```diff
@@ -0,0 +1,196 @@
...
+    age65Population: 26,
+    age55Population: 26,
+    age45Population: 26,
+    age20Population: 26,
+    age0Population: 1234,
+  },
+});
+
+// Verified to work as expected ✅
+// Note this techincally updates all of the fieldss
```
jessex, line 175: Nit: fields

src/impact-dashboard/EpidemicModelContext.tsx
jessex: CC'ing @macfarlandian just in case this is surprising to him for any reason.

sprad:
@jessex - Ok, all nits have been addressed.  For what it's worth, adding `confirmedCases` to the `modelInputs` did allow it to be properly displayed in the form.  My suspicion is that this merging code doesn't "reset" the confirmed case count with the state and county values provided by the facility's model input:  https://github.com/Recidiviz/covid19-dashboard/blob/master/src/impact-dashboard/EpidemicModelContext.tsx#L215-L224

One drawback of this is that if the value is updated in the google spreadsheet it will be overwritten by whatever we've saved in Firestone.  Unclear if that's problematic or not since presumably, the cases counts at the age level will be updated regularly which will then be rolled up and stored in firebase giving us the most up to date counts as provided by our users.

jessex:
@sprad that's the desired behavior actually. @cawarren confirmed that once a model is persisted, updated value from the Google Sheet should _not_ override what has been saved. The saved model becomes its own source of truth, unless you intentionally override it with default values that you find from experimenting with the sheet.

src/impact-dashboard/EpidemicModelContext.tsx
macfarlandian: FWIW I think storing hospital beds is a mistake. The value is not user-editable, but it is fetched from the google sheet as "locale data". so the value we would store in firestore is at best redundant and at worst out of date. If the facility is associated with a state and county but they aren't being properly selected by this view to populate this data, then that is an issue we should fix

macfarlandian:
While I'm objecting to data-related things @jessex and @sprad , I also don't think we should be storing "confirmed cases" ... we have been doing a poor job of using that value so far which has led to some confusion:

- The value we were using was actually an estimate of cases among the incarcerated, derived from the number of confirmed reported cases, which we put in our spreadsheet after fetching it from the NY Times dataset.
- We were putting this estimated value in the field labeled "confirmed cases," which was a misnomer and which @cawarren pointed out to me as a mistake at some point earlier this week. I fixed that in #122, populating that field with the actual number of reported cases among the general population in the currently selected locality. That field is user editable but we don't actually use it for any calculations, which is confusing. I don't think our users are going to be a more reliable source of state- or county-wide infection data than the official sources, right? So really this probably shouldn't be an editable field at all IMO. If the user has confirmed case numbers for the specific facility, they should be put in one of the age bracket fields or age unknown.
- We still use the estimate to seed "age unknown cases" in the model, which is fine but once the user sets any age bracket values for a facility we should definitely prefer that, as you've already said above.

Anyway that is a long-winded way of reiterating that we shouldn't store "confirmed cases" in our database because we don't use it for anything and it can, like the hospital beds figure, at best be redundant and confusing and at worst be out of date and incorrect.

Probably the same bug that was causing the hospital data to be wrong is also causing this. We should fix that! I can look into it.

PR #139, opened 2020-04-18
Renaming and Reorganizing to Match Higher Level Needs

sprad:
* The current "unnamed" facility page matched what was provided by the mock ups, but was really more of a use case specific name.  Doing a quick rename and reorganization here so that these pages match their more global intention and need.

## Type of change

## Related issues

NO-REF

### Development

### Code review


sprad:
@jessex - a quick renaming and reorganization to more closely match our needs rather than a specific use case.

src/entry-point-client/PageList.tsx
```diff
@@ -49,16 +49,16 @@ const PageList: PageInfo[] = [
     contents: <MultiFacilityPage />,
   },
   {
-    path: "/inspect-model",
-    title: getPageTitle("Inspect epidemic model output"),
+    path: "/multi-facility/facility",
```
sprad, line 52: Is there a way to set a dynamic parameter here (i.e. the facility's id) such that we can pass it as a prop to the `FacilityPage` component.  When present, it would load that facility's data.  When not present, it would load the blank form.  Not sure if that's the way things are handled in the React world or not.

jessex: I've never done that in React myself, instead relying on props to determine state. You could check if react-router has anything to that effect but I don't think, from my still non-expert perspective, that that's canonical.

sprad: Cool, cool.  Won't sweat it for now and will figure out a way to pass along the info I need

PR #140, opened 2020-04-18
properly initialize model context with locale data and saved model

macfarlandian:
## Description of the change

Following up on my comments on #138 I have fixed the issue with locale data ... as @sprad suspected there was a bug in the reset logic. Fixing it means we can remove `confirmedCases` and `hospitalBeds` from the saved model objects and join that saved data with the proper external locale data.

Now that the form is intended for a specific facility only, I don't really think that seeding the incarcerated population inputs with state- or county-wide data is such a sensible default anymore, so I just took that out. There is now just a single function to get all the default values for a locale and ensure that all required context fields are populated (which includes some metadata fields that are not locale-specific).

## Type of change

## Related issues

none

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/EpidemicModelContext.tsx
+++ b/src/impact-dashboard/EpidemicModelContext.tsx
@@ -129,35 +129,30 @@ interface ResetPayload {
   countyName?: string;
 }
 
-function getLocaleData(
+function getLocaleDefaults(
   dataSource: LocaleData,
-  stateCode: string,
-  countyName: string,
+  stateCode = "US Total",
+  countyName = "Total",
 ) {
   return {
-    confirmedCases:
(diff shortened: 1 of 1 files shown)


macfarlandian:
yes true ... the latest mocks show empty states with no default values for anything, but that is quite a change from the existing behavior so some clarification would help

cawarren:
Agree with Ian's rationale, let's remove auto population in this version.

We will have some new data that makes this relevant again in about a week, but for now no concerns at all / let's proceed with this behavior. Thanks for checking.

macfarlandian:
there was still some unnecessary complexity in the reset logic left over from the old URL management feature that I was not happy with. Now there is just a single function that gets all the default values for a locale, so it should be more straightforward to change those defaults in the future as needed.

PR #157, opened 2020-04-20
Support editing of scenario name and facilities names from the scenario details page

itsjon:
## Description of the change

Not the neatest, but about there. There are a few things that should be addressed after this PR:

- Bug: When a facility's name is edited from the scenario detail page and then the user clicks through to the facility detail page, the old name is shown on that page (though the changes are in fact persisted)
- Usability: There aren't yet any indications pre-mouseover that the scenario name and facility names are editable from the scenario details page.

## Type of change

### Development

### Code review


macfarlandian:
The two outstanding issues mentioned above should be addressed now. The pencil icon appears on hover like so: 

Screen Shot 2020-04-20 at 1 56 38 PM

I also opened #163 to address the increasing gnarliness of facility state management that contributed to the bug. Probably something we should try to get to soon because it's only going to get more complicated from here.

macfarlandian:
oh i totally missed the bit about the scenario name, sorry! I haven't touched that at all. let me take a look.

I think because we are updating the state with each keystroke, having the facilities actively re-sorted by name could be pretty disruptive while a person is typing. (waiting until a person is done typing would probably require some re-work in the state management as well as the input components themselves so if we can live with it as-is for now I think we'd be better off deferring that.)

macfarlandian:
@jessex the bit about the facility name wrapping is indeed not simple to fix in the current layout. If we can live with that for now it's another issue I could file for later followup.

macfarlandian:
also, verified no CRUD regressions with facilities

itsjon:
Thank you for fully baking this half-bake, @macfarlandian!!

src/design-system/InputText.tsx
```diff
@@ -16,20 +16,36 @@ const VDiv = styled.div`
...
+  headerStyle?: boolean;
+  onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void;
+  onKeyDown?: (event: React.KeyboardEvent) => void;
+  focus?: boolean;
 }
 
 const InputText: React.FC<Props> = (props) => {
   let inputValue = useInputValue(props);
 
+  const nameInput = useRef() as React.MutableRefObject<HTMLInputElement>;
```
szhu, line 28: @itsjon It's kind of hard to discover, but the way to do this without casting is

```suggestion
  const nameInput = useRef<HTMLInputElement>(null);
```

Issue #129, opened 2020-04-17
Unwind Persistence from Assuming only the Baseline Scenario


sprad:
**Update 4/19/2020:**  At the moment, almost all of the methods in the `index.tsx` database file reference and key off of a `baselineScenario`.  This will have to change if/when we want to support the Multi-Scenario use case.  `saveScenario` is mentioned below, but this really applies to almost all of the persistence methods.  Search for this issue's URL in the code to find references.

**Original Text:** 

The `saveScenario` method that persists a scenario's updates to Cloud Firestore just assumes we are saving the baseline scenario.  This was fine for the initial launch, but we want to support the ability to save any scenario.  As such, this method should take in a Scenario's ID and only update that particular scenario.

Additionally, we'll want to include some validation so that a user can't set the `dailyReports` and `dataSharing` flags on any scenario other than the baseline.

PR #165, opened 2020-04-20
Add logic for promo states on scenario sidebar

### LINKED ISSUE #159: Promo behavior on the sidebar of the Scenario details page
The promo on the sidebar of the scenario details page is supposed to have more sophisticated behavior, e.g. permanent dismissal and only displaying in certain  states. We haven’t done anything of the sort yet so we can do this post-release. 

Behavior: 
- Only display 0-1 promos at a time
- Permanent dismissal ( per session )
- Display in certain states (states: )

Questions:
**What are the states this should display for?**
**Which promo takes priority in these cases?**

1. Both toggles are OFF, on baseline scenario page, user has > 3 facilities
2. Both toggles are OFF, on baseline scenario page, user has < 3 facilities
3. One of the toggles is OFF, user has < 3 facilities

**Promo 1:** Turn on 'Daily Reports' to receive briefings based on the data in this scenario, prepared by Recidiviz and CSG.
- Only display if daily reports is OFF and on baseline scenario

**Promo 2:** Turn on 'Data Sharing' to provide your baseline data to public researchers, to help improve models of disease spread in prisons in the future.
- Only display if sharing toggle is OFF and on baseline scenario

**Promo 3:** Add additional facilities to see the impact across your entire system. 
- Only display once per session per scenario (save to scenario model?)
- Only display if user has < 3 facilities 

**Additional context**
See the doc for updated copy & behavior:
https://docs.google.com/spreadsheets/d/1bnBsQXbtrRogruVkDPiAketsHUrv3H26ibw0ydh94D8/edit#gid=1594361286

daschi:
## Description of the change

> Adds some logic to the `ScenarioSidebar` for when to display different text in the promo box. 
> Adds a boolean property `showPromo` on the scenario model so that a user can dismiss it permanently. 
> Adds `numFacilities` prop to `ScenarioSidebar` for knowing when to display promo text
> Adds `enabled` and `onDismiss` props for the `PromoBoxWithButton` component

## Type of change

## Related issues

Closes #159

### Development

### Code review


daschi:
Oh okay, thanks for clarifying that! I will update it to dismiss per promo per scenario. @sprad - I may need your help again later today to update the database again.

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
```diff
@@ -64,8 +64,23 @@ const MultiFacilityImpactDashboard: React.FC<Props> = ({
...
+      if (!scenario.hasOwnProperty("promoStatuses")) {
+        scenario = {
+          ...scenario,
+          promoStatuses: {
+            dataSharing: true,
+            dailyReports: true,
+            addFacilities: true,
+          },
+        };
+        await updateScenario(scenario);
```
daschi, line 79: Let me know if placing this here makes sense!

jessex: Would love to hear from @macfarlandian or @sprad on this one. The main thing was that we wanted to ensure we didn't have to manually migrate all existing scenarios to have this new `promoStatuses` object field, without which the app would break. Is there a better way to accomplish this that you're aware of?

macfarlandian: I am inclined to say that this should be somewhere in the `database` modules instead, so that the UI component just receives the data structure it requires and doesn't have to patch it.

daschi: I was thinking that as well, currently we are getting the `scenario` here from the already fetched `baselineScenario` with `baselineScenario.data.get().data()`. Would you recommend adding this to the `getBaselineScenarioRef` function in the database module, or should I add another function that is `getBaselineScenario` that would then return the scenario directly instead of the `ref` and update it with the missing property?

macfarlandian: unless @sprad disagrees, I think a new function that consumes the `ref`, updates it as needed, and just returns the scenario object is the way to go. Good idea! I'm for moving all database knowledge out of the UI components

daschi: @macfarlandian Ok! This is ready for another look.

sprad: To specifically answer the question in this PR, I agree that this code belongs in the DB module, but not in the `getBaselineScenarioRef` method.

To comment more globally though, we will want to come up with some sort of data migration process or tooling.  Otherwise, we won't really ever be able to trust our data and our code will have to be "smart" enough to work with every version and permutation of the data model throughout time.  As the code base grows, this will become untenable.

You can already see this happening with this PR.  We used to store these promo flags directly on the scenario document, but now we're storing them in a separate `promoStatuses` object which makes sense.  But, because we didn't get rid of the old flags, we're having to do stuff like this:  https://github.com/Recidiviz/covid19-dashboard/blob/097d4eb94c442cc1f7819389706a4c942454e71e/src/page-multi-facility/ScenarioSidebar.tsx#L51-L57 It's going to be very confusing in the future if someone is looking at the data for a scenario and sees both  `dailyReports` and `promoStatuses.dailyReports`.   Which should they trust?  What if their values differ?

Additionally, we're technically still doing a migration here it's just that we're doing it in slowly in production application code whenever a user who doesn't meet these conditions logs in and views this page.  At the moment, we only have about ~30 scenarios in our DB.  Once all of the users are up to date with this new `promoStatuses` model this code will no longer serve a purpose.  And unless we go back and regularly check if everyone's documents have been updated, we'll have to leave this no-op code around indefinitely.

sprad: This is the "this code" I'm referring to in the last paragraph:

https://github.com/Recidiviz/covid19-dashboard/blob/097d4eb94c442cc1f7819389706a4c942454e71e/src/database/index.tsx#L204-L213

And perhaps even the entire new `getBaselineScenario` function, unless we find a use case for it later.

jessex: @sprad, I'll file an issue for managing database migrations to be completed at some point in the near to medium future. All of those points are well met!

However, I think you maybe misunderstood the code a tiny bit, or else I did. There is no `dailyReports` field on the scenario object, just a `promoStatuses` field that includes its own `dailyReports` field. The `dailyReports` that you are referring to, if I understand, is just on the client side. Let me know if I'm wrong on that.

Otherwise, for now, let's just focus on an approach that works without too much cruft. We'll remove any outdated fields from the data model, like the original attempt with the `showPromo` boolean, and try to have the code for ensuring we have default `promoStatuses` isolated to as small of a location as possible.

sprad: @jessex - Unfortunately, we do already have both `dailyReports` and `dataSharing` fields on existing scenarios in our DB:

Screen Shot 2020-04-22 at 3 25 40 PM

Additionally, we're still setting them when we create a new baseline scenario, which we will need to stop doing (cc @daschi ):

https://github.com/Recidiviz/covid19-dashboard/blob/097d4eb94c442cc1f7819389706a4c942454e71e/src/database/index.tsx#L233-L239

Note they are already set differently (opt-in vs. opt-out)

sprad: Shit, sorry.  I am conflating two issues.  One is the actual toggle of whether or not we want data sharing or reporting.  The other is whether or not we want show promotion for these features.

jessex: Yeah sorry for that confusion, but we're on the same page now. If you think it would be simpler, we can just change the names of the internal flag fields within `promoStatuses`.

daschi: Should we wait until a migration process is ready for the data? Should I add a note above property check inside `getBaselineScenario` in the meantime?

jessex: No, let's not wait until we have a migration process constructed. Let's just get this into a mergeable, working state for all users, and then we can manually remove the `showPromo` boolean fields we had added previously.

sprad: @daschi - For now, no need to wait for a migration process.  I think adding a note above the property check in `getBaselineScenario` that references the issue @jessex is creating would suffice ✔️

jessex: That issue lives here: #186

daschi:
@jessex @sprad I added a note, let me know if this is good to go!

src/database/index.tsx
```diff
@@ -174,6 +174,49 @@ export const getBaselineScenarioRef = async () => {
   return results.docs[0].ref;
 };
 
+export const saveScenario = async (scenario: {}): Promise<void> => {
+  try {
+    // We're cheating here because for the launch we know there is only a
+    // baseline scenario. In subsequent launches, we'll need to pass in
+    // the ID of the specific scenario that we want to save. See:
+    // https://github.com/Recidiviz/covid19-dashboard/issues/186
```
sprad, line 182: https://github.com/Recidiviz/covid19-dashboard/issues/129 is still the one that's applicable for this particular comment

src/database/index.tsx
```diff
@@ -174,6 +174,49 @@ export const getBaselineScenarioRef = async () => {
...
+export const getBaselineScenario = async (
+  baselineScenarioRef: firebase.firestore.DocumentReference | null,
+) => {
+  if (!baselineScenarioRef) return null;
+
+  const result = await baselineScenarioRef.get();
+  let scenario: Scenario = result.data() as Scenario;
+
+  // NOTE: We should be able to remove this check and save once this issue
+  // is resolved: https://github.com/Recidiviz/covid19-dashboard/issues/186
```
sprad, line 205: 👍

src/database/index.tsx
daschi: oops! will update

src/database/index.tsx
```diff
@@ -174,6 +174,49 @@ export const getBaselineScenarioRef = async () => {
...
+  if (scenario && !scenario.hasOwnProperty("promoStatuses")) {
+    scenario = {
+      ...scenario,
```
sprad, line 208: Since this is a "shallow" update (i.e. the `promoStatuses` lives at the "root" level of the Scenario document), you shouldn't need to copy all of the scenario's existing attributes over with the `...scenario`.  I believe the code here can just be:

```javascript
if (scenario && !scenario.hasOwnProperty("promoStatuses")) {
  await saveScenario({
    promoStatuses: {
      dataSharing: true,
      dailyReports: true,
      addFacilities: true,
    }
  });
}
```

Let me know if that doesn't make sense

src/page-multi-facility/MultiFacilityPage.tsx
```diff
@@ -10,15 +10,17 @@ import MultiFacilityImpactDashboard from "./MultiFacilityImpactDashboard";
...
+export type BaselineScenarioRef = {
   // TODO: Confine uses of Firestore APIs (document references, etc.) to the `database` module and make this type
   //       more specific.
   data?: any;
   loading: boolean;
 };
 
 const MultiFacilityPage: React.FC = () => {
-  const [baselineScenario, setBaselineScenario] = useState<ScenarioType>({
+  const [baselineScenarioRef, setBaselineScenarioRef] = useState<
```
sprad, line 21: Thanks for the naming updates 🙏

src/database/index.tsx
daschi: Oh okay -- I was also using that `scenario` variable to return from this function to then set the internal state for the component. Do you recommend still building that object here to return? I had originally tried to return the scenario from `await saveScenario(scenario)`, but ran into some issues with TypeScript since `saveScenario`'s return is a `Promise<void>` and `void` was incompatible with `Scenario` type.

src/page-multi-facility/types.tsx
```diff
@@ -22,8 +22,9 @@ export type Scenario = {
   baseline: boolean;
   dataSharing: boolean;
   dailyReports: boolean;
+  promoStatuses?: PromoStatuses | undefined;
   description: string;
-  roles: Map<string, string>;
+  roles: object;
```
sprad, line 27: Was there a reason this changed from `Map<string, string>` to `object`?

daschi: Oops, accidentally left that in from when I was trying to add a test and was seeing a type error when I tried to create the field as a Map type. I can change it back

src/database/index.tsx
sprad: Oh, I gotcha.  This additionally merges in the `promoStatus` to the object that gets returned.  Makes sense.  No need to change what you've got.

daschi:
I think we might want to update those linting errors in `ScenarioSidebar` in a separate issue/PR. They're for the `useEffect`s that are missing the `scenario` dependency, and when I try to add that as a dependency it goes into a loop. I think we may want to move those `updateScenario` functions into the form callbacks -- either `onChange` or and `onBlur` for the text area input.

sprad:
Ok, no worries!  Feel free to merge, but please create an issue for the known listing weirdness.

Issue #201, opened 2020-04-23
R(t) UX - Add R(t) treatment to Facility Details view


cawarren:
**Is your feature request related to a problem? Please describe.**
We're now ready to start adding dynamic rate of spread! This means showing users not just the projections for their facilities, but (based on the past several days' observations) the current rate of spread in those facilities.

Rate of spread is a number from 0-7ish, where <1 indicates that the virus is spreading too slowly to expand its reach and will extinguish, while >1 indicates it will continue to spread (the larger the number, the faster it spreads). This visual treatment shows the current number for each facility.

**This issue** focuses on adding rate of spread information to the Facility Details view.

**Describe the solution you'd like**
- Add the new 'Rate of spread' chart in the upper-left of the Facility Details view
- Adjust the page layout to match the current mocks (note that the page is now situated a bit more like a report - the facility name spans the top, both inputs and outputs are aligned to start at the same part of the page, etc.

**Describe alternatives you've considered**

**Additional context**
- This issue will depend on model support for R(t) (see issue #62) to launch.
- [Current mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=1636%3A130) - Note that this issue only covers the updates to the 'Facility Details' mock
- [Copy doc](https://docs.google.com/spreadsheets/d/1bnBsQXbtrRogruVkDPiAketsHUrv3H26ibw0ydh94D8/edit#gid=1594361286)

PR #207, opened 2020-04-23
Timeseries chart for R(t)

macfarlandian:
## Description of the change

Creates a new chart for a new metric we are developing, R(t) (rate of spread over time). As this is a statistical model we also show the confidence intervals. Features in common with the `CurveChart` (tooltip, axis styles) were factored out to the design system for reuse. 

Screen Shot 2020-04-23 at 3 21 08 PM

(There are some slight differences from the mocks due to some design issues that are still being worked out but they don't have any major implications for this PR, so no reason to hold it up IMO.)

I also included a container component that does some data transformation on some fake data, which I was using for development. I don't think the transformation logic is likely to change much once it's wired up, but if we think it's a bad idea to commit the dummy data to `master` I can take that out. (I'm of two minds; I think it makes this PR more comprehensible but if we were to somehow inadvertently release a UI to prod with dummy data that would obviously be bad! We have checked in dummy data before but now that we are getting into more of a release cadence maybe that's a practice we want to abandon.)

## Type of change

## Related issues

Part of #201 

### Development

- [x] Manual testing has been performed locally **(Verified no regressions in existing chart from refactoring)**

### Code review

### DIFF
--- a/src/rt-timeseries/RtTimeseriesContainer.tsx
+++ b/src/rt-timeseries/RtTimeseriesContainer.tsx
@@ -3,6 +3,11 @@ import mapValues from "lodash/mapValues";
 
 import RtTimeseries, { Record } from "./RtTimeseries";
 
+type RawRecord = {
+  date: string;
+  value: number;
+};
+
 const convertToDate = isoParse;
 
 function dateExists(value: { date: Date | null }): value is Record {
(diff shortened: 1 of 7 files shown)


src/rt-timeseries/RtTimeseriesContainer.tsx
```diff
@@ -0,0 +1,86 @@
...
+const convertToDate = isoParse;
+
+function dateExists(value: { date: Date | null }): value is Record {
+  // date formatting function may return null with invalid input
+  // so we do need a type guard to filter the results
+  return value.date != null;
+}
+
+const RtTimeseriesContainer: React.FC = () => {
+  // TODO: this is dummy data
```
jessex, line 15: Thanks for the thoughtful consideration of this in the PR description. I am fine placing this into a `__tests__` directory as a place where we could manually run the test, but I am becoming less comfortable with checking this kind of thing into master now that, as you said, release cadence is increasing, and user counts are climbing.

macfarlandian:
@jessex I resurrected the TestPage locally in order to develop this but I don't know whether we want to check in that change either

jessex:
Would rather not check that into master. However, I think it may be okay to have a process whereby for large, new changes that don't have a dedicated surface yet, we check in some brutalist test page for the sake of review, with LOUD, VISIBLE TODOs to not merge until it is removed. But I haven't thought this through very much. I don't think you should block this particular PR on it, in any event.

src/rt-timeseries/RtTimeseriesContainer.tsx
macfarlandian: what if I just commented it out for now, with a louder TODO? putting random stuff in a `__tests__` directory will start causing problems with Jest once we get that up and running

jessex: Sure seems reasonable for now.

Issue #181, opened 2020-04-22
Switch to using an established build framework (Gatsby or Next.js)


szhu:
Our current frontend build system, largely set up by PRs #1, #4, and #16, uses Parcel and a custom script for static rendering.

This system worked well initially, since it was really small and didn't include features we needed to understand but didn't need to use. Over time, the following things have happened:
- We found a few bugs in Parcel, which we've had to work around.
- Many more developers are working on the project simultaneously.
- We needed to make custom tooling to support new features we've added like auth, web manifests, and setting up a backend. Each time we needed to specifically make sure with the frontend build system.

We have these these things that we haven't done yet:
- #80 -- supporting IE11
- #179 -- setting up testing

All of these above would probably be easier to handle if we use a framework that handles a larger set of the things we need out of the box. **Notably, I actually think that it could be easier to switch our build system to a framework  that supports IE11 than to try to hack in IE11 support into our current build system.**

Here are the two major frameworks that do approximately what we want (statically generate an interactive site):
- [Gatsby](https://www.gatsbyjs.org/)
- [Next.js](https://nextjs.org/)

I spent the past few hours looking at their documentation and going through the setup for each, and they both seem easy to use. Some notable reasons to choose each:
- Gatsby seems to support IE11 even when the Node modules we use don't. [[Source](https://www.gatsbyjs.org/docs/browser-support/)] Next.js, on the other hand seems to have issues that are fairly recent regarding this still. The way to solve this seems to require using a third-party plugin.
- Next.js has a framework for server-side functions (haven't looked into whether it's compatible with the backend stuff we need to do), while Gatsby is static-only.
- Gatsby seems more "batteries included" and polished with less technical know-how to get set up.
- Next.js is maintained by ZEIT ([they rebranded yesterday to be called Vercel](https://vercel.com/blog/zeit-is-now-vercel)), which also maintains Now.

It looks like Gatsby is the more appropriate choice at the moment, but I'm still looking into it.

---

Fixes #80 (we will only choose a framework if it supports IE11)

jessex:
[...]

PR #179, opened 2020-04-21
Add jest/ts-jest testing framework

daschi:
## Description of the change

> I wanted to add a test for the promo behavior logic and I chatted with @macfarlandian about adding jest. Took a shot at it today and ran into some issues when I tried to use `react-testing-library` to render and test components. I'll try to pick this back up tomorrow to sort it out, but would love to pair on it if anyone is interested!

Some of the issues I saw that I think are related to react-testing-library: 
- All components and child components must have React explicitly imported
- For some reason I needed to import react as `import * as React from "react"` in the test and component files

I think it might be worthwhile to check out enzyme too and see if it has less issues. 

There's also a few packages I installed from trying to debug this that I'm not 100% sure we need: 
- @babel/preset-react
- jest-styled-components
- @testing-library/jest-dom

I'm also not super familiar with `parcel`, so for now I started with a mock file for svg/css assets in `src/__mocks__/`. 

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

Closes [#XXXX]

### Development

These boxes should be checked by the submitter prior to merging:

- [ ] Manual testing has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #209, opened 2020-04-24
Switch to importing React when using JSX

szhu:
## Description of the change

`<MyComponent />` in JSX translates to `React.createElement(MyComponent)` in JS. Thus, every time we use JSX angle brackets, `React` needs to be in scope. The community-accepted way to do this is to import React whenever we use JSX angle brackets.

When I set up the build system, I thought that we could make things simpler by just setting React globally. Then we could avoid importing React in most files!

This worked well when creating new components, but it had the following drawback: Every time we needed to add a new kind of entry point, we needed to figure out how to set React globally. This happened when we added static building, and again in #179 in which we're considering adding Jest tests. And now that I'm trying to add Gatsby (#181), I'm running into this issue again.

After discussing with @daschi a few days ago, we decided to go back to using the community-accepted practice of importing React in all files using JSX angle brackets. This PR makes this change.

Method:
1. Un-disable the Eslint rule that checks for importing React.
2. Run `yarn-eslint-check`.
3. Fix all occurrences.
4. Make sure `yarn-eslint-check` doesn't fail anymore.

## Type of change

### Development

### Code review


.eslintrc.json
```diff
@@ -56,9 +56,6 @@
     // types.
     "react/prop-types": "off",
 
-    // We've removed this requirement by setting window.React.
-    "react/react-in-jsx-scope": "off",
```
szhu, line 60: Significant change 1

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "es5",
     "module": "commonjs",
-    "jsx": "preserve",
+    "jsx": "react",
```
szhu, line 5: Significant change 2

src/entry-point-static/build.tsx
```diff
@@ -1,5 +1,5 @@
-import React from "react";
```
jessex, line 1: In this case, why did we delete this line as well, instead of just the global call directly below?

tsconfig.json
jessex: For anyone else who wanted to see the docs on this: https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage

src/entry-point-static/build.tsx
szhu: Oh that's just because nothing in this file requires this import anymore.

szhu:
What the force-pushes and then force-pushed back: I noticed that the description inside my second commit created a record at the bottom of this page:
`https://github.com/parcel-bundler/parcel/issues/1199`

So I tried removing the description but it just added a second record instead :/ so I just reset to the original

Issue #62, opened 2020-04-09
Add dynamic R(t) calculaton to model


macfarlandian:
As users update case counts over time, we can derive an observed rate of spread and use that to make adjustments to the initial constants per facility.

cawarren:
*Update:*
- @justkunz has implemented this in Python, and is working on some small revisions on the modeling side as well as some updated epidemiology variables.
- Juan and I are working on a UI treatment, but it shouldn't block implementation of dynamic R(t) calculation as part of the model.

cawarren:
As a note on why this will be prioritized fairly high this week: Many more facilities are now seeing cases, and the rate of spread varies significantly between them. 

A one-size-fits-all model worked well for directional use when forecasting different scenarios, but now that states are in the thick of things they need to be able to see a customized version of where they're at in facilities and whether they're on the right track.

If it's helpful to understand contextually what R(t) ends up looking like, check out rt.live as an example (that website is tracking it for the general population in different states, but using the same methodology we're implementing).

PR #219, opened 2020-04-24
Set up Python cloud functions

macfarlandian:
## Description of the change

Initial setup to let us add Python cloud functions to the dashboard application. These are essentially the same as the Firebase functions we already have, but because Firebase doesn't really deal with Python it requires some extra setup and different tooling. To avoid mixing JS and Python files in a confusing manner I have located them in a parallel folder under `/backend`.

The function itself is mostly a placeholder, though I tried to at least define a realistic API for it in consutlation with @justkunz , so do flag anything there you don't like. It's not yet wired up to Auth0 either, although we will probably want to do that before releasing the feature to prod.

I have verified that I can deploy this function to GCP and it does show up in our Firebase Functions console, so we at least don't need to do anything special to monitor it once it's up.

## Type of change

## Related issues

Part of #62 

### Development

### Code review


macfarlandian:
I added CORS handling because I realized we would need it once I tested fetching from the webapp. 

Also added more legit schema validation because I found an easy reference source that did both! I think that's good to have anyway, leaving it out of the initial commit was a shortcut.

backend/README.md
```diff
@@ -53,3 +52,36 @@ From the `backend` directory:
...
+
+In `python-functions` set up your virtual environment:
+
+```sh
+virtualenv venv
+source venv/bin/activate
+pip install -r requirements.txt
+```
+
+Unlike Firebase, we don't have tools to emulate these functions locally.
```
jessex, line 74: But to be clear, when you run the app locally it will still call out to the deployed Python function, so if you're working on anything other than the function itself, there's no concern, right?

backend/README.md
```diff
@@ -53,3 +52,36 @@ From the `backend` directory:
...
+```sh
+python -m unittest
+```
+
+## Deployment
+
+Deploy the functions using the `gcloud` command line tool for Google Cloud Platform.
+
+```sh
+gcloud functions deploy <FUNCTION_NAME> --runtime python37 --trigger-http --allow-unauthenticated
```
jessex, line 86: I haven't thought deeply about this in relation to the JS function since it has not changed since initial deploy, but what is your thinking regarding deploys? We could have one staging function and one production function, then set the URL as an environment variable and have staging/master and production configure those vars individually?

backend/python-functions/helpers.py
```diff
@@ -0,0 +1,68 @@
...
+            if request.method == 'OPTIONS':
+                return cors_options()
+
+            # If it's not a CORS OPTIONS request, still include the base header.
+            headers = {'Access-Control-Allow-Origin': '*'}
+
+            try:
+                if in_schema:
+                    request_json = request.get_json()
+                    jsonschema.validate(request_json, in_schema)
```
jessex, line 40: Do these validate calls automatically return 400/BAD_REQUEST on failure?

backend/python-functions/helpers.py
```diff
@@ -0,0 +1,68 @@
...
+def cors_options():
+    # Allows GET and POST requests from any origin with the Content-Type
+    # header and caches preflight response for an 3600s
+    headers = {
+        'Access-Control-Allow-Origin': '*',
+        'Access-Control-Allow-Headers': 'Content-Type',
+        'Access-Control-Max-Age': '3600'
+    }
+
+    return ('', 204, headers)
```
jessex, line 68: Nit: newline

backend/README.md
macfarlandian: that's right. i am working on a subsequent PR to set that up

backend/python-functions/helpers.py
macfarlandian: no, they just raise an exception that gets caught on L50 below; then we return a 500 response on L51. It's very generic, but we could tease that out to provide more meaningful error codes if you think that's worthwhile now

jessex: That seems sufficient for now.

backend/README.md
macfarlandian: hmmmm ..... i definitely agree that this deploy process leaves something to be desired. I'm not sure how we would vary the configuration by branch (let me know if there is something specific you had in mind?) but it shouldn't be too tough to write our own little deploy scripts that would take an argument to specify the environment. so like `python deploy.py production` or whatever ... I will take a crack at that

backend/README.md
```diff
@@ -53,3 +52,59 @@ From the `backend` directory:
...
+## Development
+
+While we cannot emulate the Cloud Functions environment locally, we can deploy separate
+development and production versions of the function to facilitate end-to-end development
+and more rigorous pre-release testing.
+
+All developers' local builds of the dashboard will consume the `development` version of
+the function, so please be aware of how any changes you make will affect others and take
+reasonable measures to avoid breaking it and let others know if it will be unavailable
+or unstable for a time while you are actively developing it.
```
macfarlandian, line 90: @jessex note that a subsequent PR to wire up the JS app will have to implement this using an environment variable or some such

backend/README.md
jessex: I am thinking we would use environment variables to vary by conceptual tier, in one of two ways:
1. Production vs Non-Production
2. Production vs Staging vs Local

I would go with option 1 for now because I don't see enough value in option 2 to outweigh the additional complexity. So then we'd use an env var that points to the right Cloud Function path and set that env var directly in Now.sh production to point to the prod path, and set it elsewhere to point to the stage path. The only thing to figure out there would be how to ensure that all Now.sh deployments _except_ production share the same env var. But that should be an imminently solvable problem!

macfarlandian: ah yes got it. that is actually pretty much what I was thinking (and have been working on), I think I wasn't sure whether you were talking about the envvar would also affect the function deploy process itself!

macfarlandian:
@jessex had to make one additional small tweak to get the tests to pass and then verify that it actually ran properly in the cloud ... so I am already seeing how necessary it is to have the dev deployment option!

.gitignore
```diff
@@ -29,3 +30,8 @@
 # other people from accidentally committing them.
 /package-lock.json
 /pnpm-lock.yaml
+
+# Python
+venv
+__pycache__
+.mypy_cache
```
szhu, line 37: Hi @macfarlandian, can you move these into `backend/.gitignore`?

macfarlandian: oops too late but yes I will

Issue #144, opened 2020-04-19
Figure out a way to remove JSON shenanigans in model persistence


jessex:
**What needs to be done? Why does it need to be done?**
As first added and noted [here](https://github.com/Recidiviz/covid19-dashboard/pull/143/files#r410903129), the logic to invoke the model save function from the frontend involves serializing and immediately deserializing the model inputs as json. There should very likely be a better way.

szhu:
This needs to be fixed to make IE11 work.

> Browsers with native Map support serialize Maps as `{}`, browsers without native Map support try to serialize the Map polyfill object and get a circular reference error.

PR #225, opened 2020-04-25
Switch build framework to Gatsby

### LINKED ISSUE #80: Provide IE11 support

### LINKED ISSUE #181: Switch to using an established build framework (Gatsby or Next.js)

szhu:
Note to reviewers: This PR is HUGE. Please review commit-by-commit!

## Description of the change

Switches the build framework to Gatsby. Everything that the site used to support should still work, including:

- React
- TypeScript
- Routing
- Styling: custom fonts, Tailwind, global styles, styled-components, ~emotion~*
  - \* react-spinners works fine without explicit emotion support, so no change was needed
- Page titles
- Favicon and web manifest
- Login gating
- [ ] Google analytics (we should test this end-to-end to verify that it works)

New things:

- IE11 is now supported (static builds only)
- The image sizes for the web manifests/favicons are generated automatically now
- Website contacts/relies on fewer external domains. 
  - Before:
    ```
    code.jquery.com
    d3js.org
    docs.google.com
    fonts.googleapis.com
    fonts.gstatic.com
    model-login.recidiviz.org
    unpkg.com
    www.google-analytics.com
    www.googletagmanager.com
    ```
  - After:
    ```
    docs.google.com
    model-login.recidiviz.org
    www.google-analytics.com
    ```
    

Tested:
- `yarn dev` works
- `yarn build && yarn gatsby serve` works
- Tested that the built site works on Chrome and IE11 (using browserstack.com)
- Tested that the built site looks the same or better when JS is turned off
- `yarn test` still works

## Type of change

## Related issues

Fixes #181 (Switch to Gatsby)
Fixes #80 (Support IE11)
Touches #144 (`JSON.parse(JSON.stringify(...))`)

### Development

### Code review


szhu:
@jessex the build doesn't work because the output folder is now `public/` instead of `dist/`. When we're ready to deploy this, we can switch that over here:
https://vercel.com/recidiviz/covid19-dashboard/settings

src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -44,14 +44,37 @@ const FacilityInputForm: React.FC = () => {
...
+      // We don't need to save localeDataSource.
+      localeDataSource: undefined,
+    };
+
     saveFacility({
       id: facility?.id,
       name: facilityName || null,
       description: description || null,
       systemType: systemType || null,
-      modelInputs: JSON.parse(JSON.stringify(model))[0],
```
sprad, line 52: @szhu - Great work on this PR.  For what it's worth, it may be possible to simplify this by avoiding the parse/stringify code and the JsonSerialization by just doing this

```javascript
modelInputs: model[0],
```

I've tested that and it works locally on my machine, but I don't have a way to test on IE 11

sprad: TL; DR: Ignore my comment above, it actually doesn't work.  Details below if you're interested:

Actually, nevermind.  This does allow for the model to be saved, but it serializes and saves things in a way that blows up when we try to read the data back in for form population.  For example, we currently store the dates associated with planned releases as strings which turns out to be ideal because that plays nicely with our `DateInput` components.  With the suggestion above (i.e. `modelInputs: model[0],`), the planned release dates get saved as actual timestamps in Firestore which blow things up when we try to load the form on the facility information page.

src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -44,14 +44,37 @@ const FacilityInputForm: React.FC = () => {
...
+      modelInputs,
```
jessex, line 75: @szhu what you have here works in both IE11 and other browsers? What testing have we done?

szhu: This seems to work in both IE11 and Chrome. (I can save a facility and recall it. Previously, I could not save the facility in IE11.) It appeared safe to omit `localeData` because even in Chrome, `JSON.stringify(localeData)` evaluates you `{}`, so I don't think it was being saved before.

I'm not as familiar as some you with what the correct functionality is or what exactly `localeData` is, so I plan on doing a call to test this out (like the one I had with @macfarlandian on Friday) before merging it.

src/page-multi-facility/FacilityInputForm.tsx
szhu: Note about the TS definition: The JsonSerializable code doesn't run at all. It's a static check guaranteeing that the variable at this point is serializable. This helps prevent this entire class of bugs from being discovered only after we try testing it!

There are some things it doesn't check for, for example whether our code deals with Dates being converted into strings, and undefined being converted into null. We can be more explicit about those by removing the Date and undefined types from the definition.

It realize that might be hard to describe the intent of that code over text -- let's do a call tomorrow and talk about it!

README.md
```diff
@@ -15,28 +15,26 @@ yarn install
...
 
-Built files can be found in `/dist`.
+Built files can be found in `public/`.
 
-An easy way to try running the built files:
+An easy way to test out the built files:
 
 ```sh
-bash -c "cd dist && npx serve"
+yarn gatsby serve
```
jessex, line 33: Let's add a command to `package.json` that wraps around this call, e.g. `yarn run-build` or something with an _even better name_

bin
```diff
@@ -1 +0,0 @@
-node_modules/.bin
```
jessex, line 1: What was this used for previously?

gatsby-config.js
```diff
@@ -0,0 +1,42 @@
...
+  plugins: [
+    "gatsby-plugin-postcss",
+    "gatsby-plugin-react-helmet",
+    "gatsby-plugin-styled-components",
+    "gatsby-plugin-typescript",
+
+    {
+      resolve: "gatsby-plugin-google-analytics",
+      options: {
+        trackingId: "UA-117977508-2",
```
jessex, line 13: Note to self: to test this in staging, I'll need to temporarily disable the Google Analytics filter I set up which "blocks" localhost and now.sh sub-prod traffic.

src/hooks/useQueryParams.tsx
```diff
@@ -44,9 +45,9 @@ const useQueryParams = (
   defaultValues: QueryParams,
   validKeys?: string[],
 ): UseQueryParamsOutput => {
-  const history = useHistory();
+  const location = useLocation();
```
jessex, line 48: Curious: What necessitated this change?

src/page-home/HomePage.tsx
```diff
@@ -1,5 +1,5 @@
+import { Link } from "gatsby";
```
jessex, line 1: Similar to my previous comment: Why did we switch all of these navigation components to something provided by Gatsby? At first/immediate glance, switching core functionality from React out to a more specific/implementation-driven set of components makes me _slightly_ nervous.

src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -1,12 +1,13 @@
...
 import React, { useContext, useState } from "react";
-import { useHistory } from "react-router-dom";
 import styled from "styled-components";
 
 import { saveFacility } from "../database/index";
 import InputButton from "../design-system/InputButton";
 import InputDescription from "../design-system/InputDescription";
 import InputNameWithIcon from "../design-system/InputNameWithIcon";
 import ChartArea from "../impact-dashboard/ChartArea";
+import { PlannedRelease } from "../impact-dashboard/EpidemicModelContext";
```
jessex, line 10: I think this was a mistaken addition.

src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -44,14 +44,37 @@ const FacilityInputForm: React.FC = () => {
   const model = useModel();
 
   const save = () => {
+    // When saving an object to Firestore, Firestore will reject it if it is not
+    // serializable. This type enforces that what we are saving is serializable.
+    // In the future, we add new properties to the model that are not
```
jessex, line 49: Grammar nit: "In the future, if we add..."

gatsby-config.js
```diff
@@ -0,0 +1,42 @@
...
+        head: true,
+      },
+    },
+
+    {
+      resolve: "gatsby-plugin-manifest",
```
jessex, line 19: Does this manifest not need/support specifying multiple icon sizes, like our previous `manifest.webmanifest`?

src/pages/unsupported-browser.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+import UnsupportedBrowserPage from "../page-unsupported-browser/UnsupportedBrowserPage";
```
jessex, line 3: Would you mind also updating the copy for this page now that we will be able to support IE11? Something like, "We’re sorry, legacy browsers such as versions of Internet Explorer prior to IE11 are not..."

src/site-provider/SiteProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+      onRedirectCallback={onRedirectCallback as any}
+    >
+      <LocaleDataProvider>
+        <FacilityContext.Provider value={{ facility, setFacility }}>
+          {props.children}
+        </FacilityContext.Provider>
+      </LocaleDataProvider>
+    </Auth0Provider>
+  );
+};
```
jessex, line 41: We no longer need the "render or hydrate" logic that used to exist in `src/entry-point-client/index.tsx`?

src/page-multi-facility/FacilityInputForm.tsx
macfarlandian: I think there's really no reason for us to be serializing `localeData` anyway — we shouldn't be saving it, that is the data structure we are loading and building from a separate datasource in Google Sheets. There is an enumerated list of `persistedKeys` in the `EpidemicModelContext` module and anything that's not in that list can be safely discarded (and indeed SHOULD be discarded) for persistence operations

src/hooks/useQueryParams.tsx
szhu: Gatsby uses Reach Router (`@reach/router`) instead of React Router (`react-router`), which has a very similar but different API. Here is the complete set up substitutions I made:
- `useLocation()` => `useLocation()` (no change, they both have `useLocation()`)
- `useHistory().location` => `useLocation()`
- `useHistory().push(...)` => `navigate()`
- `useHistory().replace(...)` => `navigate(..., {replace: true})`

We can't mix and match routing libraries because each library looks for its own provider.

Note: React Router's API will eventually look more like the Reach Router one.
https://reacttraining.com/blog/reach-react-router-future/

src/page-home/HomePage.tsx
szhu: React Router isn't an official React library, it's a third-party library just like Reach Router is.

It's probably a bad to use two routing libraries simultaneously, as I imagine they might manage their own state.

src/site-provider/SiteProvider.tsx
szhu: Correct, it's handled for us.

gatsby-config.js
szhu: It does! See the PR description and https://covid19-dashboard-lj3mj6imu.now.sh/manifest.webmanifest

bin
szhu: I added this as a shortcut for us to directly run installed Node commands, for example, `bin/parcel`.  Note that these are for binscripts that are come with the packages we use, not the scripts we have defined in `package.json`.
- No one ever needed to use it since we set up scripts for them
- I don't think anyone but me actually ran it
- It caused issues like #19
- I realized that Yarn supports a just-as-easy way to do this: `yarn parcel`, etc.

README.md
szhu: I want the commands to be separate because you might want to kill the server but restart it later without recompiling everything. Would it be okay to have `yarn gatsby serve` as `yarn serve` or `yarn serve-built`?

I guess we can also have `yarn build-and-serve` as a wrapper around both commands. I just want to make the individual commands very discoverably.

jessex: Oh that's what I had in mind, sorry for the confusion from my bad name. I don't want the commands to be clobbered together by any means.

szhu:
> As regards the `dist/` --> `public/` switch, I'm happy to update config in now.sh right before merge of this PR. Alternatively, if possible, we could consider configuring the public directory for Gatsby to be `dist/` to avoid the incompatibility, but I don't feel strongly about it.

@jessex 
- I don't think it's supported: `https://github.com/gatsbyjs/gatsby/issues/18975#issuecomment-545905403`
- I also think we should use `public` moving forward instead of `dist` because It's the Now default. (Given this, note that you should turn off the override instead of writing in `public` when making the change change.)

src/hooks/useQueryParams.tsx
jessex: Ah I had no idea that was a 3rd party library. Given all of this material, I feel better about this!

src/page-multi-facility/FacilityInputForm.tsx
szhu: Okay cool thanks @macfarlandian! That's the kind of reassurance I was looking for 🙌

src/pages/unsupported-browser.tsx
szhu: Oh, this page doesn't work right now -- I took out the logic to detect old browsers. (I could have left it in, but there's no guarantee that it would have worked, so I didn't want to give a false sense of security.) I'll create an issue so that we can decide whether to add in that functionality (and what that will look like) or just remove it.

src/page-multi-facility/FacilityInputForm.tsx
sprad: We use `persistedKeys` to filter out values we're not interested in saving at the persistence layer:  https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/index.tsx#L313-L315

szhu:
Here are some things that didn't work when @jessex and I tested manually.

**Critical to address:**

- The "delete scenario" functionality crashes the app
- (We haven't tested the new user flow yet and we should)

**Possibly easy to address:**

- Entering "7" in the "Bunk capacity" changed it to `7.0000000001`, hard to input another number to make it "75"

**Important to address but not critical, and also possibly hard to debug/fix:**

- Everything is really slow (but just be IE11 and/or testing in a VM)
- Sometimes we get these errors:
  - When loading the main list: `Network Error 0x2eff`. Doesn't appear to break functionality
  - When navigating back to the main page: image
  - When navigating back to the main page: image

src/page-multi-facility/FacilityInputForm.tsx
szhu: @sprad and I resolved this on a call.

Here's what we did:
- Used just `model[0]`
- Placed `JSON.parse(JSON.stringify())` inside `saveFacility`, after the call to `_.pick()`

See here for further discussion: https://github.com/Recidiviz/covid19-dashboard/pull/225/files#r415377061

src/page-multi-facility/FacilityInputForm.tsx
szhu: @sprad and I resolved this on a call following the above advice. The `persistedKeys` correctly already filters out `localeDataSource`.

szhu:
@jessex I wasn't able to switch the output folder for that last commit. The setting to toggle this seems to be currently broken in Now:
https://vercel.com/recidiviz/covid19-dashboard/settings
image

szhu:
Okay, did a little more sanity testing and it seems like everything is working. I'll merge this now!

Follow-up tasks:
- I should make sure that this doesn't break #234, which got merged in the late afternoon today. I don't know how that new page works, but if I wait until tomorrow to merge this PR, then we'll just end up with another set of newly merged changes that we have to integrate 🙃  (I messaged @jovergaag to follow up on this!)
- Remaining issues for IE11 here #232

szhu:
Updating Now settings:

| Before    | After    |
| ----------| -------- |
| ![before] | ![after] |

[before]: https://user-images.githubusercontent.com/1570168/80435936-c0e4d580-88b2-11ea-8993-e365057bc722.png
[after]: https://user-images.githubusercontent.com/1570168/80435952-c9d5a700-88b2-11ea-8199-b206cf0ec8cc.png

PR #245, opened 2020-04-28
Handle Firestore timestamps more explicitly

### LINKED ISSUE #144: Figure out a way to remove JSON shenanigans in model persistence

szhu:
## Description of the change

This PR description will be replaced. See PR comments for details.

## Type of change

## Related issues

Fixes #144 

### Development

### Code review

### DIFF
--- a/src/design-system/InputDate.tsx
+++ b/src/design-system/InputDate.tsx
@@ -80,7 +80,11 @@ const InputContainer = styled.div`
 
 const InputDate: React.FC<InputBaseProps<Date>> = (props) => {
   const { labelAbove, labelHelp, onValueChange } = props;
-  const inputValue = useInputValue(props);
+  let inputValue = useInputValue(props);
+
+  if (inputValue && "toDate" in inputValue) {
+    inputValue = (inputValue as any).toDate();
+  }
 
   return (
     <InputContainer>
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -317,13 +317,6 @@ export const saveFacility = async (
       // the original.
       facility.modelInputs = pick(facility.modelInputs, persistedKeys);
 
-      // Convert the dates in plannedReleases to strings. Otherwise, Firestore
-      // will serialize these dates to timestamps in a way that is not
-      // compatible with our date picker library. Instead, save these dates
-      // values as strings in Firestore.
-      // https://github.com/Recidiviz/covid19-dashboard/issues/144
-      facility.modelInputs = JSON.parse(JSON.stringify(facility.modelInputs));
-
       facility.modelInputs.updatedAt = currrentTimestamp();
 
       // TODO: For now, this assumes we're always entering data as of "today"


src/design-system/InputDate.tsx
```diff
@@ -80,7 +80,11 @@ const InputContainer = styled.div`
 
 const InputDate: React.FC<InputBaseProps<Date>> = (props) => {
   const { labelAbove, labelHelp, onValueChange } = props;
-  const inputValue = useInputValue(props);
+  let inputValue = useInputValue(props);
+
+  if (inputValue && "toDate" in inputValue) {
+    inputValue = (inputValue as any).toDate();
+  }
```
szhu, line 87: Currently, this PR is a proposal for how to address #144.

The problem we need to solve is that if we naively save `modelInputs`, it contains some Dates. Firebase saves Dates as Firebase Timestamps. When we get the data back, DatePicker doesn't know how to use the Timestamps.

Existing solution: Use `JSON.parse(JSON.stringify(...))` to turn dates into strings.

Proposed here: Leave Dates as Dates, and allow Firebase to save them and retrieve them as Firebase Timestamps. Only when when the Timestamp reaches the date picker does it convert it back to a date using `toDate` [[source](https://stackoverflow.com/a/52247484/782045)]

I don't think this proposal is the correct solution either, but I think it's a simple proof of concept (it works!) that can lead to an efficient discussion about what the right solution should be.

jessex: Although the json solution is hacky, I think the location where it lives is correct. This proof-of-concept, if carried out, would mean that all date fields would be Firebase timestamps by default and anywhere that we needed a more portable/canonical date type, we'd have to `toDate` it in that location. There are features coming up in the next week or two that will need to make more use of dates than we have so far, so I _think_ I'd prefer it if the client treated all date fields with canonical date types throughout the app until it reached `src/database`, at which point it would translate into Firebase timestamps on write and back to dates on read. But we'd want to be careful about maintaining date precision and time zones. There could be better approaches!

szhu: Yeah makes sense. I think this makes sense for @sprad and I to discuss on a call when we're both free. Things to discuss:
- What is the canonical date format we're going to use in the frontend?
- How to make sure things are correctly converted just before/after Firestore. Two general approaches: do a deep conversion of nested entries, looking for datelike objects or do a precise conversion of just the nested entries that need to be converted
- Date precision, timezones, and making sure the comparison operators work correctly with the ones we do choose (if we might need to use them in the future)

sprad: I've been digging into this a bit because I believe we're going to run into the same issue when we begin to store user input for the "observed at" date input as part of this work:  https://github.com/Recidiviz/covid19-dashboard/pull/251.

The good news is that I don't think we have to do anything special to handle saving JavaScript Dates into Firestore.  Firestore will automatically save them as Timstamps, which I believe is fine.

We do have an issue however when we read these timestamps back out of Firestore though.  I believe the ideal solution is the one proposed by @jessex (i.e. convert these timestamps in JavaScript Date objects in the persistence layer).  @szhu describes the approaches for handling this conversion well in his second bullet point above, but I'd like to add some more specifics here.  To make this happen we would need to:

1.  Remove the `JSON.stringify` hack so that the Planned Release and Observed At dates can be stored as timestamp values (and not strings) in Firestore.  ~~This also means we'll need to manually migrate the data for the existing planned release date strings to firebase timestamps.  This is made a bit challenging due to the fact that there are copies of these values stored within the versioned `modelInputs`.  Fortunately, not many users have made use of the planned release feature yet so there won't be too many instances to update.  We'll just need to ensure we're thorough as we go through each facility and its current and versioned model inputs.~~. We can avoid migrating existing data if we make the date conversion code in Number 2 smart enough to deal with strings.

2.  When we read from Firestone we're going to have to manually convert all relevant instances of timestamps to JavaScript dates.  Essentially, we'll need to write code to act as a lightweight "ORM" to handle the mapping of Firestore Timestamps to JavaScript Dates for the following fields:

**Scenario**

- `createdAt`
- `updatedAt`

**Facility**

- `createdAt`
- `updatedAt`
- `modelInputs.plannedReleases.date` (for each planned release entry)
- `modelInputs.observedAt`
- `modelInputs.updatedAt`

We'll also need to repeat this process for each of the `modelInput` fields listed above for each document stored in the `modelVersions` sub-collection.

I was somewhat hopeful we could just get this for free using TypeScript.  For example, assuming the following type (notice `createdAt` and `updatedAt` are declared as `Date`):

```javascript
export type Facility = {
  id: string;
  scenarioId: string;
  name: string;
  description?: string;
  systemType?: string;
  modelInputs: EpidemicModelPersistent;
  createdAt: Date;
  updatedAt: Date;
};
```

I was hoping the code `facility = doc.data() as Facility` would just magically convert from Timestamp to Date, but that does not seem to be the case.  As such, I believe we're going to have to create our own custom mapping code as described above.

I'm working on a draft PR now to better explain my rambling above so if what I wrote doesn't make sense stay tuned...

sprad: I'm not super familiar with the JavaScript ecosystem so if it feels like I'm complicated things or if it feels like this is a solved problem (using something click [class-transformer](https://github.com/typestack/class-transformer)), let me know.

sprad: @szhu and I spoke on the phone tonight and decided the approach listed above (i.e. having the persistence layer return native JavaScript Date objects to client code) is the best/most robust path moving forward.  As such, we will need to have some way to map Firestore Timestamps to JavaScript Date objects.  Long term, we'll need to do something like I proposed in my comment above, but for now, we'll be able to get away with setting some Firestore configuration as follows*:

```javascript
  const firestore = firebase.firestore();
  firestore.settings({ timestampsInSnapshots: false });
```

With this configuration, Firestore will automatically return JavaScript Date objects for Timestamp fields which is perfect.  Unfortunately, this setting is [deprecated](https://firebase.google.com/docs/reference/js/firebase.firestore.Settings#optional-timestampsinsnapshots) and will be removed in an upcoming release of Firestore.  You also will notice this error in your browsers console with this setting in place:

`@firebase/firestore: Firestore (7.13.2): Support for 'timestampsInSnapshots: false' will be removed soon. You must update your code to handle Timestamp objects.`

If we want to prioritize delivery speed, this is our best option for now, but clearly we'll need a more robust Timestamp to Date mapping solution in the not so distant future.

\* Props to @szhu for finding this configuration setting

szhu: It's important to also note why the option is deprecated:

> Previously, Firestore returned timestamp fields as Date but Date only supports millisecond precision, which leads to truncation and causes unexpected behavior when using a timestamp from a snapshot as a part of a subsequent query.

We should look at whether we need nanosecond precision (we probably don't). If we need it, then we should use Timestamps. If we don't, then we need to make sure that non-millisecond-aligned Timestamps never enter the system.

jessex: This will be closed with the merging of #263, yes?

sprad: Yuuuuuup

sprad:
Closing in favor of #263 

Thanks for the guidance and feedback @szhu

PR #252, opened 2020-04-29
Do real Rt calculation in the python cloud function

### LINKED ISSUE #62: Add dynamic R(t) calculaton to model

macfarlandian:
## Description of the change

Replaces the placeholder `calculate_rt` function with real computation. Credit where due: everything in `realtime_rt` is the handiwork of @justkunz; all I really did was construct the response object and write tests. 

I also changed the API to pass the dates as strings instead of timestamps. This conforms better to what the compute function expects, but it also has the benefit of being more explicit and able to be validated as a proper date by the schema, so it's better all around IMO.

I tested this end-to-end on top of #250 and verified it works! (The fetches generally took ~2-3 seconds with just two days of data, FWIW.)

## Type of change

## Related issues

Part of #62 

### Development

### Code review

### DIFF
--- a/backend/python-functions/realtime_rt.py
+++ b/backend/python-functions/realtime_rt.py
@@ -51,7 +51,6 @@ def get_posteriors(sr, sigma=0.15):
     process_matrix /= process_matrix.sum(axis=0)
 
     # (4) Calculate the initial prior
-    #prior0 = sps.gamma(a=4).pdf(r_t_range)
     prior0 = np.ones_like(r_t_range)/len(r_t_range)
     prior0 /= prior0.sum()
 
@@ -76,7 +75,7 @@ def get_posteriors(sr, sigma=0.15):
         #(5b) Calculate the numerator of Bayes' Rule: P(k|R_t)P(R_t)
         numerator = likelihoods[current_day] * current_prior
 
-        #(5c) Calcluate the denominator of Bayes' Rule P(k)
+        #(5c) Calculate the denominator of Bayes' Rule P(k)
         denominator = np.sum(numerator)
 
         # Execute full Bayes' Rule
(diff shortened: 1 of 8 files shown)


backend/python-functions/realtime_rt.py
```diff
@@ -0,0 +1,143 @@
...
+    # (3) Create the Gaussian Matrix
+    process_matrix = sps.norm(loc=r_t_range,
+                              scale=sigma
+                             ).pdf(r_t_range[:, None])
+
+    # (3a) Normalize all rows to sum to 1
+    process_matrix /= process_matrix.sum(axis=0)
+
+    # (4) Calculate the initial prior
+    #prior0 = sps.gamma(a=4).pdf(r_t_range)
```
jessex, line 54: Should this be removed prior to merging?

backend/python-functions/realtime_rt.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Note that this takes a while to execute - it's not the most efficient algorithm
+    try:
+        hdis = highest_density_interval(posteriors, p=.9)
+    except:
+        raise ValueError('Unable to compute R(t) with the provided data')
+
+    most_likely = posteriors.idxmax().rename('ML')
+    result = pd.concat([most_likely, hdis], axis=1)
+    return result
```
jessex, line 143: Nit: newline

backend/python-functions/test.py
```diff
@@ -41,3 +38,23 @@ def test_invalid_input(self):
         resp = self.get_response_json()
 
         self.assertIn('error', resp)
+
+    def test_empty_input(self):
+        data = {
+        'dates': [],
```
jessex, line 44: Nit: indentation here and the line below, and same in the next test case.

backend/python-functions/realtime_rt.py
justkunz: That's my bad! This can be removed

PR #260, opened 2020-04-29
Curve functions accept numeric rate of spread

macfarlandian:
## Description of the change

Both the R(t) feature and Excalibur are going to need to generate curves based on arbitrary rate of spread factors that aren't captured by our existing low/med/high enums. This PR changes the interface for our curve functions to just require numeric inputs. 

There was existing logic in the curve function to translate the enum key into numeric values; that has been factored out into a separate function. Components that were dumping the `EpidemicModelContext` directly into curve functions now need to use this new function to transform the context data structure first. Future components for the features mentioned above may provide these values by other means, which may or may not depend on the enum value selected by the user. There may come a point where we deprecate and remove that enum entirely, but we're not there yet, so for now this new function bridges the gap.

Also added some tests for the curve function to help me make this change with confidence. They mainly just replicate the manual sanity checks we have been doing on model inspection page (inspecting table structure, computing checksums).

## Type of change

## Related issues

Part of #62 

### Development

### Code review

### DIFF
--- a/src/infection-model/index.tsx
+++ b/src/infection-model/index.tsx
@@ -1,5 +1,6 @@
 import { range, sum } from "d3-array";
 import ndarray from "ndarray";
+import { Omit } from "utility-types";
 
 import { EpidemicModelInputs } from "../impact-dashboard/EpidemicModelContext";
 import {
@@ -14,10 +15,13 @@ export type CurveData = {
   staff: ndarray;
 };
 
-export interface CurveFunctionInputs extends EpidemicModelInputs {
-  rateOfSpreadCells?: number;
-  rateOfSpreadDorms?: number;
-}
+export type CurveFunctionInputs = Omit<
+  EpidemicModelInputs,
+  "rateOfSpreadFactor"
+> & {
+  rateOfSpreadCells: number;
+  rateOfSpreadDorms: number;
+};
 
 function prepareAgeGroupPopulations({
   age0Population,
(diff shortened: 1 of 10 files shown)


src/infection-model/seir.tsx
```diff
@@ -272,15 +259,13 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
...
     rateOfSpreadCells -
     (1 - facilityOccupancyPct) *
       (rateOfSpreadCells - rateOfSpreadCellsAdjustment);
-  let rateOfSpreadDorms = R0Dorms[rateOfSpreadFactor];
   const rateOfSpreadDormsAdjustment = 1.7; // magic constant
-  rateOfSpreadDorms =
+  const rateOfSpreadDormsAdjusted =
     rateOfSpreadDorms -
     (1 - facilityOccupancyPct) *
       (rateOfSpreadDorms - rateOfSpreadDormsAdjustment);
```
macfarlandian, line 271: @justkunz it's not strictly in scope for this change but I was wondering whether you think all these adjustments will still be valid when the input values come from an R(t) calculation?

src/infection-model/index.tsx
```diff
@@ -14,6 +15,14 @@ export type CurveData = {
   staff: ndarray;
 };
 
+export type CurveFunctionInputs = Omit<
```
jessex, line 18: So this whole operation means that `CurveFunctionInputs` has all fields in `EpidemicModelInputs` except for `rateOfSpreadFactor` but also has `rateOfSpreadCells` and `rateOfSpreadDorms` added. Right?

macfarlandian:
@jessex that's the gist of it, yes.

src/infection-model/index.tsx
macfarlandian: that is correct

Issue #262, opened 2020-04-30
Populate Domain Objects with Consistent Types


sprad:
**Is your feature request related to a problem? Please describe.**

We currently manage dates and times inconsistently in Firestore and in our code.  Specifically, we are storing automatically managed dates (i.e. `createdAt`, `updatedAt`, etc.) as Firestore Timestamps, but we are storing user input values (i.e. Planned Release dates) as strings.

This creates issues for client code interacting with data returned from Firestore because it must know how to deal with these different data types.  This leads to inconsistent, confusing, and buggy code.

**Describe the solution you'd like**

A better solution would be to populate our domain objects (i.e. `Facility`, `Scenario`, `PlannedRelease`, etc.) with specific types so that client code can easily know about and interact with data in a consistent manner.  At the moment, this mostly applies to Date like objects.  See the conversation [here](https://github.com/Recidiviz/covid19-dashboard/pull/245/files#discussion_r417629653) for a list of the current fields we need to explicitly manage.

**Note:**  This solution should be implemented such that we do not have to do any manual data migrations.  This means that the code to populate Date objects will need to be able to handle either Strings or Firestore Timestamps.

**Describe alternatives you've considered**

Firestore does have a mechanism for returning its Timestamps as dates, but this is a deprecated approach that will break our code when Google releases the next version of Firestore "soon."  For more details see [this conversation](https://github.com/Recidiviz/covid19-dashboard/pull/245/files#r417727550).

szhu:
@sprad just checking, this is closed by the PR, right?

sprad:
@szhu - Yup!  It's done

PR #263, opened 2020-04-30
Populate Domain Objects with Consistent Types

### LINKED ISSUE #144: Figure out a way to remove JSON shenanigans in model persistence

sprad:
## Description of the change

* Populating our Domain objects with consistent JavaScript Dates so that client code can be written consistently without worrying about dealing with multiple date-like types.

## Type of change

## Related issues

Closes #144 #262

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -5,15 +5,16 @@ import "firebase/auth";
 import "firebase/firestore";
 
 import createAuth0Client from "@auth0/auth0-spa-js";
-import { parseISO } from "date-fns";
 import { pick, sortBy } from "lodash";
 
 import config from "../auth/auth_config.json";
-import {
-  PlannedRelease,
-  persistedKeys,
-} from "../impact-dashboard/EpidemicModelContext";
+import { persistedKeys } from "../impact-dashboard/EpidemicModelContext";
 import { Facility, ModelInputs, Scenario } from "../page-multi-facility/types";
+import {
+  buildFacility,
+  buildModelInputs,
+  buildScenario,
+} from "./type-transforms";
 
 // As long as there is just one Auth0 config, this endpoint will work with any environment (local, prod, etc.).
 const tokenExchangeEndpoint =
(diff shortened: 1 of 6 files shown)


src/database/index.tsx
```diff
@@ -123,6 +119,68 @@ const buildUpdatePayload = (entity: any) => {
...
 
+const timestampToDate = (timestamp: firebase.firestore.Timestamp): Date => {
+  return timestamp.toDate();
+};
+
+const buildPlannedRelease = (plannedReleaseData: any): PlannedRelease => {
+  let plannedRelease: PlannedRelease = plannedReleaseData;
+
+  const plannedReleaseDataDate = plannedReleaseData.date;
+  if (plannedReleaseDataDate instanceof firebase.firestore.Timestamp) {
```
sprad, line 130: This type check allows us to operate with the existing planned release string dates without the need for a data migration

src/infection-model/rt.tsx
```diff
@@ -79,8 +79,10 @@ const getRtInputsForFacility = async (
   const dates: string[] = [];
 
   modelVersions.forEach((model) => {
+    const seconds = model.observedAt.getTime() / 1000;
+
     dates.push(
-      formatISO(fromUnixTime(model.observedAt.seconds), {
+      formatISO(fromUnixTime(seconds), {
```
sprad, line 85: @macfarlandian - The persistence layer is now consistently returning native JavaScript Dates instead of Firestore Timestamps.  This update should function equivalently to what was previously here but to the extent it's useful to work with actual dates, feel free to modify the rt and other modeling code.

Is there a way I can test the RT code locally?  I didn't see this module exported anywhere

src/database/index.tsx
```diff
@@ -383,13 +431,6 @@ export const saveFacility = async (
       // the original.
       facility.modelInputs = pick(facility.modelInputs, persistedKeys);
 
-      // Convert the dates in plannedReleases to strings. Otherwise, Firestore
-      // will serialize these dates to timestamps in a way that is not
-      // compatible with our date picker library. Instead, save these dates
-      // values as strings in Firestore.
-      // https://github.com/Recidiviz/covid19-dashboard/issues/144
-      facility.modelInputs = JSON.parse(JSON.stringify(facility.modelInputs));
```
sprad, line 391: Later! ✌️

src/database/index.tsx
```diff
@@ -123,6 +119,68 @@ const buildUpdatePayload = (entity: any) => {
   return payload;
 };
 
+const timestampToDate = (timestamp: firebase.firestore.Timestamp): Date => {
+  return timestamp.toDate();
+};
+
+const buildPlannedRelease = (plannedReleaseData: any): PlannedRelease => {
```
jessex, line 126: In the interest of keeping this file manageable, I might move these new functions out to a helper class in the package, e.g. `src/database/type-transforms.tsx` or `src/database/builders.tsx` or `src/database/date-handlers.tsx` or something.

src/database/index.tsx
```diff
@@ -123,6 +119,68 @@ const buildUpdatePayload = (entity: any) => {
...
+  let plannedRelease: PlannedRelease = plannedReleaseData;
+
+  const plannedReleaseDataDate = plannedReleaseData.date;
+  if (plannedReleaseDataDate instanceof firebase.firestore.Timestamp) {
+    plannedRelease.date = timestampToDate(plannedReleaseDataDate);
+  } else {
```
jessex, line 132: So this else assumes that the only other type option is a string, right? Is that a safe assumption on the way out of Firestore?

src/database/index.tsx
```diff
@@ -123,6 +119,68 @@ const buildUpdatePayload = (entity: any) => {
...
+  document: firebase.firestore.DocumentData,
+): Facility => {
+  const documentData = document.data();
+
+  let facility: Facility = documentData;
+  facility.id = document.id;
+  facility.scenarioId = scenarioId;
+  facility.createdAt = timestampToDate(documentData.createdAt);
+  facility.updatedAt = timestampToDate(documentData.updatedAt);
+  facility.modelInputs = buildModelInputs(documentData.modelInputs);
```
jessex, line 168: Do we need to handle `modelVersions`?

src/infection-model/rt.tsx
macfarlandian: in latest master it should be testable now ... you just have to enable the feature flag via environment variable (instructions in README) and restart your dev build

src/database/index.tsx
sprad: It should be a safe assumption because that's the only thing we previously stored for the Planned Release dates.  That said, I can turn this into an `else if` to check for String explicitly and throw an exception in the final `else` case.

src/database/index.tsx
sprad: We do need to handle `modelVersions`, but we don't need to handle them here.  Data returned from Firestore is "shallow" meaning that sub-collections (i.e. `modelVersions`) are not returned.  Note that we recently added a method for grabbing the model versions called `getFacilityModelVersions` and I did update that code to make use of this new `buildModelInputs` transformer.  The context is lost in this PR because it's just a one-line update, but you can see the whole method in all its glory here (see line 369):  https://github.com/Recidiviz/covid19-dashboard/blob/dec45a60e42f95c206efdb91f62564f6f40f22fd/src/database/index.tsx#L349-L378

jessex: Right on! I did notice that you had updated `getFacilityModelVersions` but what I did _not_ know was that these returns were all shallow (or rather, I forgot, after the last time you told me). This makes sense. Thanks!

src/database/index.tsx
jessex: I think I would like that defensiveness--in either approach we get an error but in this approach we can better differentiate between the two error cases of "not an expected type" and "not a parseable string."

sprad:
@jessex PR feedback addressed in the latest commit

Issue #196, opened 2020-04-23
Fix inconsistent text box fields of scenario details page


jessex:
**Describe the bug**
Through rapid movement, the three different text boxes on the Scenario Details page have all become different in style/manner:
* The scenario rename field shows a pencil on hover, then you click and it becomes an obvious text box. When you hit enter or click away, it goes back to regular text
* The facility rename field shows a pencil on hover, but when you click the only cue that it is editable is that a cursor starts blinking. There's no style change otherwise
* The scenario description field is just a permanent text box

We should make these a consistent experience for users.

**Expected behavior**
Everything should behave in the manner of the scenario rename field: by default it looks like regular text, a pencil icon is shown on hover, it turns into an obvious text box when you click, and it goes back to normal when you're done.

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

**Additional context**

jessex:
The scenario rename and scenario description fields on the scenario details page been made consistent by #206.

amyrlam:
I can claim this ticket, unless another one seems better. Will take a look at the linked PR's

Issue #176, opened 2020-04-21
Add scenario libraries view


cawarren:
**Is your feature request related to a problem? Please describe.**
We currently only support the baseline scenario in the web model, but ask that the user keep this model reflecting the real-world. This makes it difficult for the user to adequately forecast different scenarios and what-ifs (e.g., 'What if we were to release some of the low-risk members of the incarcerated population? What if we were to make space to move more people out of bunked housing and into cells?')

In order to accomplish this, we need a place in the UI for users to view, open, add, or delete their scenarios.

**Describe the solution you'd like**
Creation of the 'Library' modal dialog for viewing the user's scenarios. See mocks below.

**Additional context**
- This is a child issue of issue #168 
- [Current mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=910%3A130)

amyrlam:
Is this modal, which would pop up after clicking the InputNameWithIcon folder icon

![image](https://user-images.githubusercontent.com/1372946/80773014-76ab6080-8b0d-11ea-836d-0fef29e552c9.png)

jessex:
Yep! This modal is the library view. We should ask @cawarren whether each chart is meant to be _one_ of the facility model charts for that scenario, say the first one in alphabetical order, or whether it's supposed to be a composite across the entire scenario, like what's happening on the impact page (Excalibur).

cawarren:
The designer's intent was for this to be the summed output for the scenario. If that's challenging from an implementation standpoint though, I'd feel free to grab the first facility chart - really this is eye candy for this view, it's not a functional component of the model output.

PR #264, opened 2020-04-30
[WIP] Address inconsistent text box fields

amyrlam:
## Description of the change

_To be edited, will add "WIP comments" as a comment._

## Type of change

## Related issues

https://github.com/Recidiviz/covid19-dashboard/issues/196

### Development

### Code review

### DIFF
--- a/src/page-multi-facility/FacilityRow.tsx
+++ b/src/page-multi-facility/FacilityRow.tsx
@@ -7,6 +7,7 @@ import Colors, { MarkColors as markColors } from "../design-system/Colors";
 import { DateMMMMdyyyy } from "../design-system/DateFormats";
 import iconEditSrc from "../design-system/icons/ic_edit.svg";
 import InputTextArea from "../design-system/InputTextArea";
+import TextLabel from "../design-system/TextLabel";
 import CurveChartContainer from "../impact-dashboard/CurveChartContainer";
 import {
   totalConfirmedCases,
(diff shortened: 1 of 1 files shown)


amyrlam:
inputtext

@jessex @szhu question for standup as didn't quite finish yet, thought draft pr would be the best way to ask

i'm proposing that we get rid of the edit function in the right hand side facility row, seems somewhat confusing UX (but maybe there is some need there, I don't know about). then all facility edits would happen on the /facility route with a "save"

- would need to adjust styling in the facility row textarea so get cursor: pointer
- could get rid of textarea and just be text, but I think it was a textarea to address the overflow. we could address that in other ways but unclear if NVA given time constraints

after talking to sean yesterday, kind of realized that i wasn't sure why the subclick was there in the FacilityRow, when most facility stuff only persists edit on save 

//

can still do design tweaks to the input interaction, if is a good use of time
inputtext2
- removing the folder icon from the component
- size jump on edit as input is going over the folder icon
- not having the pencil there on all hover

amyrlam:
cc @cawarren hi andrew, this is the question i mentioned at standup, lmk what you think

jessex:
Thanks for the breakdown and screengrabs, @amyrlam!
* I would personally not be opposed to removing the facility rename functionality from the facility row if it leads to overall improved UX. I think it's a nice-to-have feature to avoid a few extra clicks in what might otherwise just be a typo fix, but I don't feel strongly about it and I somehow doubt it's a frequently used feature outside of testing. @cawarren would probably have more conviction one way or another on this one.
* If we did get rid of it, we would want to make sure that we did handle long facility names gracefully--that's worth the time spent. Some state prison names, for example, are really, really long and our userbase may be apt to use the full official names more often than not.
* I think fixing the size jump is probably most important of those three UX quirks you identified. I think having the folder icon still present regardless of whether we are currently editing the scenario name or not would be nice. But I don't have any issue with the current pencil behavior, personally. @cawarren again would have stronger conviction on these points.

amyrlam:
thanks for chiming in. yeah i can also keep the edit in the FacilityRow, and use `InputDescription` to keep it a textarea, which is addressing overflow rn

amyrlam:
talked to andrew
- opening a new branch now
- it may make sense to remove the folder from the component, because it will be used here https://github.com/Recidiviz/covid19-dashboard/issues/176 to open a modal

Issue #240, opened 2020-04-27
[Excalibur] Create 2 curved charts: one for original values and one for current values


rocktacular:
follow design to show 2 charts
1. for each chart, show infectious, hospitalized, fatalities
2. for each chart, show X & Y axis labels

We should be able to use existing component for this.

jovergaag:
PRs #268 and #267 Close this issue.

PR #267, opened 2020-04-30
[Excalibur] Add the current projection chart 

daschi:
## Description of the change

> Adds the current project curve chart with data from all facilities on a scenario. This does not currently whitelist any facilities, I think we could do that as a separate issue (cc @sychang ) so not to hold up the rest of the components.

## Type of change

## Related issues

Part of #240 

### Development

### Code review


src/page-response-impact/responseChartData.tsx
```diff
@@ -0,0 +1,57 @@
...
+  if (!facilitiesInputs.length)
+    return {
+      // NOTE: We should guard against this in the dashboard and remove
+      // this default return from this function
+      exposed: [],
+      fatalities: [],
+      hospitalized: [],
+      infectious: [],
+    };
+  const facilitiesProjectionData = calculateCurveData(facilitiesInputs);
```
jessex, line 45: Future thought: If we run into performance issues, we should break this out such that we call `calculateCurveData` for each individual input, then await all of them finishing before proceeding.

src/page-response-impact/responseChartData.tsx
```diff
@@ -0,0 +1,57 @@
...
+    (output) => output.incarcerated.data,
+  );
+  const staffData = facilitiesProjectionData.map((output) => output.staff.data);
+  const combinedData = zip(...incarceratedData, ...staffData);
+  const summedData = combinedData.map((row) => {
+    return row.reduce((sum, value) => {
+      return (sum += value);
+    }, 0);
+  });
+  return ndarray(summedData, [90, 9]);
```
jessex, line 32: Nit: can we extract 90 and 9 out to well-named constants, so that future readers who are less well-versed with the domain model understand what they mean?

daschi: Yes sounds good. I know 90 is the number of days, and 9 seems to be the number of SEIR categories, do these variables make sense? 

`const NUM_DAYS = 90; const NUM_SEIR_CATEGORIES = 9`

rocktacular: this info helps me 😃

Issue #269, opened 2020-04-30
Reuse updated chart components on the model inspection page


macfarlandian:
**Is your feature request related to a problem? Please describe.**
`ModelInspectionPage` is drifting dangerously far from the functionality on the main page: the chart legend works differently, the data is fetched differently. It's at risk of becoming burdensome to maintain if more drift occurs.

**Describe the solution you'd like**
Update this page to reuse the chart components on the MultiFacility page. Mainly this will involve substituting generic handlers in the places where particular curves are enumerated (the legend, the chart data transformation hook). That way, on the multifacility page we can provide an enumerated list of curves but on the model inspection page we can simply render all of them. This page will still have to implement separate data fetching logic (using the `useProjectionDataFromUserInput` hook) because it doesn't currently support R(t), nor is it expected to imminently.

**Describe alternatives you've considered**
I've already partially implemented this in a local branch but didn't have time to finish. Other than some TypeScript weirdness it didn't seem too bad.

**Additional context**

PR #270, opened 2020-04-30
Decouple viz containers from EpidemicModelContext

macfarlandian:
## Description of the change

This is more table-setting for incorporating Rt into the projection model; since it was already getting large I decided to pause here for a PR rather than rolling all the Rt stuff into it also. 

Similar to #260 this is further decoupling the visualization components (chart and table) from the EpidemicModel context; instead of fetching their data directly from the model they now accept it as a prop. There is a little bit of state passthrough happening here but it didn't seem burdensome enough to warrant a new Context or extension of some existing one. Instead I added a few stateful hooks to handle the necessary data transformation.

Finally I deleted some unused components from the old "impact dashboard" feature that were broken by these changes. They were just cruft anyway but this was the forcing function to actually get rid of them.

(To give you a preview of how this relates to the R(t) feature, in place of `useProjectionDataFromUserInput` we will substitute a different hook that passes the Rt values into the curve function on the multi-facility pages. The `useProjectionDataFromUserInput` hook would still be useful for the `/inspect-model` page, which is why it's not just throwaway work here, but to expedite the Rt feature I propose we defer the migration of that page, per #269. I started to do it on this branch but it didn't seem worth the time now.)

## Type of change

## Related issues

Part of #62 

### Development

### Code review


jessex:
@jovergaag and @daschi: I think this will produce merge conflicts for your recent PRs for the Excalibur projection charts, but we need this one to go in first for the R(t) feature. Just a friendly heads up! Let us know if you run into any issues after this is merged.

Issue #271, opened 2020-04-30
TypeError in python fn for Rt calculation when adding a new facility version


daschi:
**Describe the bug**
I'm seeing the following error from the Rt python function: 
`TypeError: '<' not supported between instances of 'Timestamp' and 'float'`

**To Reproduce**

1. Go to the multi facility page
2. Click on one of the facilities
3. Edit the facility staff or incarcerated cases and save 

Note: since I'm trying to add a new feature, I am console logging the return from the function `getRtDataForFacility(facility)`. 

**Expected behavior**
Expected the Rt calculation to return values for the multiple versions of the facility, but it returned `null` instead. 

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

**Desktop (please complete the following information):**

- OS [e.g. Windows 10]: MacOS Catalina 10.15
- Browser [e.g. chrome, edge]: Chrome
- Version [e.g. 22]: 81

**Additional context**
full stacktrace:
```
File "/user_code/helpers.py", line 41, in wrapped
    function_output = f(request_json)
  File "/user_code/main.py", line 20, in calculate_rt
    result_df = compute_r_t(request_json)
  File "/user_code/realtime_rt.py", line 133, in compute_r_t
    posteriors, _ = get_posteriors(smoothed, sigma=.25)
  File "/user_code/realtime_rt.py", line 77, in get_posteriors
    numerator = likelihoods[current_day] * current_prior
  File "/env/local/lib/python3.7/site-packages/pandas/core/ops/__init__.py", line 776, in f
    self, other, pass_op, fill_value=fill_value, axis=axis, level=level
  File "/env/local/lib/python3.7/site-packages/pandas/core/ops/__init__.py", line 616, in _combine_series_frame
    left, right = self.align(other, join="outer", axis=axis, level=level, copy=False)
  File "/env/local/lib/python3.7/site-packages/pandas/core/frame.py", line 3834, in align
    broadcast_axis=broadcast_axis,
  File "/env/local/lib/python3.7/site-packages/pandas/core/generic.py", line 8486, in align
    fill_axis=fill_axis,
  File "/env/local/lib/python3.7/site-packages/pandas/core/generic.py", line 8603, in _align_series
    other.index, how=join, level=level, return_indexers=True
[...]

PR #273, opened 2020-04-30
[Excalibur] Add the r(t) stats summary 

### LINKED ISSUE #239: [Excalibur] Create R chart to show change in rate of transmission over time
As per design:
1. show initial/original R0
2. show current R
3. show changes over time 
4. show `y=1` line
5. tooltip to display R(t) value on hover

Questions: 
1. where are the historical data for this calculation? will it only be a straight line between 2 points?
2. is R(t) exposed on the model engine we have?

**Updated requirements:** 
Because R(t) can only be modelled on a per-facility basis and does not make sense to calculate for the system as a whole, we will show some summary statistics about the R(t) across the facilities:

1. X of Y facilities currently have an R(t) < 1
2. Average r(t) reduction across facilities: E.g., if f1 went from 3.7→ 1.7 and f2 went from 3.0→ 2.0, the average change across those facilities would be facility r(t) reduced by an average of 1.5
3. Max r(t) change: Difference between r(t)s between the facility with the original max r(t) and the facility with the current max r(t) (e.g., if original r(t)s were f1=5 and f2=3, and currently we have f1=2 and f2=2.5, then the difference would be 5-2.5 = max r(t) reduced by 2.5)

daschi:
## Description of the change

> This adds three r(t) summary points to the page:
1. number of facilities below r(t) 1
2. average change of r(t) across facilities 
3. max change of difference across facilities

I'm opening this is as a draft PR because I'm blocked from testing this locally with the model until issue https://github.com/Recidiviz/covid19-dashboard/issues/271 is resolved. 

## Type of change

## Related issues

Closes #239

### Development

### Code review


src/infection-model/rt.tsx
```diff
@@ -117,3 +118,11 @@ export const getRtDataForFacility = async (
...
   }
 };
+
+export const getRtDataForFacilities = async (facilities: Facilities) => {
+  return await Promise.all(
+    facilities.map(async (facility) => {
+      return await getRtDataForFacility(facility);
+    }),
+  );
+};
```
macfarlandian, line 128: there's a similar function now on the MultiFacility page ... probably we should consolidate them into one (i think having it live here is probably fine)

src/page-response-impact/rtStatistics.tsx
```diff
@@ -0,0 +1,42 @@
+import { RtRecord } from "../infection-model/rt";
+
+export function numFacilitiesWithRtLessThan1(
+  facilitiesRtRecords: RtRecord[][],
+) {
+  const latestRtValues = facilitiesRtRecords.map((rtRecords) => {
+    // Note: This assumes the first in the array is the latest because of the
+    // sorting on line 104 in infection-model/rt.tsx
+    return rtRecords[0].value;
```
macfarlandian, line 9: i'm pretty sure this is backwards? the last item should be the latest, i.e., they should be in chronological order

I myself have been wondering how safe it is to always assume they'll be in the right order (they currently _should be_ but this could change without warning and break various things) ... at the very least maybe adding a test on `cleanRtData` would make this a safer assumption

src/page-response-impact/rtStatistics.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function averageRtReductionAcrossFacilities(
+  facilitiesRtRecords: RtRecord[][],
+) {
+  const numFacilities = facilitiesRtRecords.length;
+  const rtDifferences: number[] = facilitiesRtRecords.map(
+    (rtRecords: RtRecord[]) => {
+      const latestRtValue = rtRecords[0].value;
+      const oldestRtValue = rtRecords[rtRecords.length - 1].value;
```
macfarlandian, line 21: seeing how much we need to access these (not just here but in other Rt-related features too) is making me think we ought to have helper functions like `getOldestRt`, `getNewestRt`, or whatever, to encapsulate that logic and make it easier to change

src/page-response-impact/rtStatistics.tsx
```diff
@@ -0,0 +1,42 @@
...
+      return oldestRtValue - latestRtValue;
+    },
+  );
+  const sumDifferences = rtDifferences.reduce(
+    (sum, rtDifference) => sum + rtDifference,
+    0,
+  );
+  return sumDifferences / numFacilities;
```
macfarlandian, line 29: this is fine but calling `lodash.mean` or `d3.mean` on `rtDifferences` would be more concise. at the very least we should use `rtDifferences.length` rather than `numFacilities` in case there is a future change such that they get out of sync (like, we drop invalid values from the list of differences or something like that ... it's not completely far-fetched)

src/page-response-impact/RtSummaryStats.tsx
```diff
@@ -0,0 +1,40 @@
...
+  );
+  const maxRtChange = rtStats.maxRtChange(facilitiesRtRecords);
+  return (
+    <div>
+      <p className="mt-5 mb-5">{`${numFacilitiesWithRtLessThan1} of ${numFacilities} facilities currently ${
+        numFacilities > 1 ? "have" : "has"
+      } an R(t) < 1`}</p>
+      <p className="mt-5 mb-5">
+        Average R(t) reduction across facilities:{" "}
+        {averageRtReductionAcrossFacilities}
```
macfarlandian, line 33: would recommend coercing this into a nicer display with `numeral(x).format("0.0")`, these values tend to have floating point arithmetic artifacts

src/infection-model/rt.tsx
daschi: Sounds good...I'll try to update this so that we can retain the rtData caching/dispatching that's ont he multi facility dashboard

src/page-response-impact/rtStatistics.tsx
daschi: That sounds good, also I can write those to check the data instead of relying on the order of the array.

src/page-response-impact/ResponseImpactDashboard.tsx
```diff
@@ -39,6 +43,8 @@ import {
 const ResponseImpactDashboard: React.FC = () => {
   const { data: localeDataSource } = useLocaleDataState();
   const [scenarioState] = useScenario();
+  const { rtData } = useContext(FacilityContext);
+  const useRt = useFlag(["useRt"]);
```
daschi, line 47: @macfarlandian @jessex I'm not sure whether we would want this feature flag on the impact page, but I started with it here. Would we want to remove this section of the impact page if they don't select the use Rt functionality?

daschi:
I think this could be ready for another look, I haven't been able to really test this locally because my data isn't set up enough to compute `rt` for my fake facilities. I was wondering if there's already a data fixture that we could use for development that would set up enough data for modeling Rt, or if that's something other people would find helpful. 

I think some open questions are: 
- What "total facilities" number do we use on the donut chart when there isn't Rt data for every facility model (for now @sychang suggested always using the total number of facilities on the scenario)

src/page-response-impact/ResponseImpactDashboard.tsx
jessex: I could be wrong but I think that Excalibur is probably an Rt-dependent feature in general and that we assume Rt is enabled in order to use it. That feature flag was a temporary measure to avoid breaking all other functionality while it was in flight and should go away in the near future, so I'd opt in favor of just removing this flag from here, unless @macfarlandian disagrees.

jessex:
@daschi we don't have any special utility that creates more data for you in dev. I would just pick a facility or two and use the add cases modal to add a variety of data points in the past leading up to the present day. That will give you some Rt functionality.

src/page-response-impact/ResponseImpactDashboard.tsx
macfarlandian: yes that's correct, it was just a precaution to keep master release-safe while work was in progress. not intended to be a long-term thing and should probably be ripped out now that it has been successfully released. i'll make an issue so we don't forget

macfarlandian:
you're not the first person to ask about this  @daschi so it might be worth me investing a little time in. but for now the rule of thumb is that you need at least two separate days of data and the total number of cases needs to increase by at least 10 people day-over-day

src/page-response-impact/RtSummaryStats.tsx
```diff
@@ -0,0 +1,157 @@
...
+    .filter(filterRtData)
+    .map((rtData: RtData) => rtData.Rt);
+
+  const numFacilitiesWithRtLt1 =
+    rtStats.numFacilitiesWithRtLessThan1(facilitiesRtRecords) || 0;
+
+  const averageRtReductionAcrossFacilities =
+    rtStats.averageRtReductionAcrossFacilities(facilitiesRtRecords) || 0;
+
+  const numFacilitiesWithRtGt1 = totalFacilities - numFacilitiesWithRtLt1;
```
macfarlandian, line 75: shouldn't this be subtracted from `facilitiesRtRecords.length` instead? here we are including facilities where we don't actually know what the Rt is.

src/page-response-impact/RtSummaryStats.tsx
```diff
@@ -0,0 +1,157 @@
...
+        subtitle: "with R(t) < 1.0",
+      },
+    ],
+  };
+
+  return (
+    <RtSummaryStatsContainer>
+      <ResponsiveOrdinalFrame
+        {...frameProps}
+        svgAnnotationRules={(d: any) => {
```
macfarlandian, line 102: why not just put this in `frameProps`?

src/page-response-impact/RtSummaryStats.tsx
```diff
@@ -0,0 +1,157 @@
...
+                </text>
+                <text
+                  style={{
+                    lineHeight: "16px",
+                    fontFamily: "Poppins",
+                    transform: "translate(0px, 20px)",
+                  }}
+                  fontSize="12"
+                  fontWeight="normal"
+                >
```
macfarlandian, line 131: the styles and display attributes here are almost all static; couldn't these just be styled components instead?

src/page-response-impact/rtStatistics.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { getNewestRt, getOldestRt, RtRecord } from "../infection-model/rt";
+
+export function numFacilitiesWithRtLessThan1(
+  facilitiesRtRecords: RtRecord[][],
+) {
+  const latestRtValues = facilitiesRtRecords.map((rtRecords) => {
+    return getNewestRt(rtRecords)?.value;
+  });
+  return latestRtValues.filter((v) => v && v < 1).length;
```
macfarlandian, line 11: `v` could be `0` here, right? in which case we would actually want to include it in the count

src/page-response-impact/rtStatistics.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export function averageRtReductionAcrossFacilities(
+  facilitiesRtRecords: RtRecord[][],
+) {
+  const rtDifferences: (number | undefined)[] = facilitiesRtRecords.map(
+    (rtRecords: RtRecord[]) => {
+      const latestRtValue = getNewestRt(rtRecords)?.value;
+      const oldestRtValue = getOldestRt(rtRecords)?.value;
+      return latestRtValue && oldestRtValue && oldestRtValue - latestRtValue;
```
macfarlandian, line 21: again I think we need to make sure we are not excluding zeroes here

src/hooks/useFacilitiesRtData.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Facilities } from "../page-multi-facility/types";
+
+const useFacilitiesRtData = (facilities: Facilities | null, useRt = false) => {
+  const { rtData, dispatchRtData } = useContext(FacilityContext);
+
+  async function getRtDataForFacilities(facilities: Facilities) {
+    return await Promise.all(
+      [...facilities].map(async (facility) => {
+        // don't fetch data if we already have it
+        if (rtData && rtData[facility.id]) return;
```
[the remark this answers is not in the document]
daschi: Yeah that makes sense, maybe we could extend this object to be closer to a `{loading, failed, data } setup so that we can check whether the request failed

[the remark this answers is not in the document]
macfarlandian: yeah works for me ... this undefined vs null thing has caused a problem literally everywhere anyone has tried to work with it (including myself immediately after I implemented it) so I think your suggestion is a good one as well

src/page-response-impact/RtSummaryStats.tsx
daschi: Oh okay yeah -- I think I misunderstood a conversation with @sychang about this. I'll check with her real quick to confirm it should be the total number of facilities we have Rt data for.

src/page-response-impact/RtSummaryStats.tsx
daschi: Yes totally! I'll clean these up

src/page-response-impact/RtSummaryStats.tsx
daschi: Okay so from @sychang:
"X will be the # of facilities 1) that we have data for 2) that have rt <1
Y is the total number of facilities that they have modelled overall, not just the number we have rt values for"

So I think that means we should keep it as is with `rtDataValues.length` (which I'm assuming would have a value for every facility modeled, even if some return null)

src/page-response-impact/rtStatistics.tsx
daschi: Oh yeah good catch, I always forget that 0 evaluates to false.

src/page-response-impact/RtSummaryStats.tsx
macfarlandian: yeah that makes sense ... in that case I think you should rename this variable to avoid future confusion about what this is supposed to represent

src/page-response-impact/ResponseImpactDashboard.tsx
daschi: Okay cool, I've removed it from the response impact page. I'm still referencing it in the function `useFacilitiesRtData(facilities, true)` on line 88 which would just need to be updated once we remove the flag.

src/page-response-impact/RtSummaryStats.tsx
```diff
@@ -72,14 +102,15 @@ const RtSummaryStats: React.FC<Props> = ({ rtData }) => {
   const averageRtReductionAcrossFacilities =
     rtStats.averageRtReductionAcrossFacilities(facilitiesRtRecords) || 0;
 
-  const numFacilitiesWithRtGt1 = totalFacilities - numFacilitiesWithRtLt1;
+  const numFacilitiesWithRtGt1 =
+    totalFacilitiesInScenario - numFacilitiesWithRtLt1;
```
macfarlandian: i am still going to object to this because this is not the number of facilities with Rt greater than 1 ... would call it like `numFacilitiesRemainder` or something else not too wordy that doesn't make any false assertions about its contents

daschi: Oh yeah, that makes sense--I misunderstood which variable to rename. I'll update this now.

Issue #258, opened 2020-04-29
log Rt cloud function inputs and outputs


macfarlandian:
will be useful for debugging issues in the wild

PR #276, opened 2020-05-01
Use R(t) value in projection model

### LINKED ISSUE #62: Add dynamic R(t) calculaton to model

macfarlandian:
## Description of the change

When enabled by feature flag, this uses the latest fetched R(t) value as the "rate of spread" input to the projection model. (When not enabled, there should be no change in functionality.)

A few things had to happen to enable this:

- I ran into trouble with the new curve function hooks I just wrote, because I forgot you can't use hooks conditionally (they have to run in the same order on every render), so now there is just one hook for getting the curve that takes optional Rt input, and decides what data transformations to apply based on that.
- Based on advice from @justkunz I also lifted more logic related to the qualitiative R0 user input out of the curve function and into the related data transformation, to make sure it was not erroneously applied to Rt input.
- All the projection visualizations need to be in a loading state while they are waiting for Rt data. The most straighforward way to accomplish this seemed to be to let the curve data be `undefined` while it's waiting for inputs, and let that `undefined` flow all the way through the data pipeline to be handled by the chart and table components (they already mostly handled it, I just had to clean up the legend toggle logic a bit). 
- Because of that change, I had to add a type guard on top of @daschi 's recently merged changes to filter these undefined values out of the Excalibur chart data pipeline.

## Type of change

## Related issues

Closes #62 

### Development

### Code review

### DIFF
--- a/backend/python-functions/realtime_rt.py
+++ b/backend/python-functions/realtime_rt.py
@@ -124,8 +124,9 @@ def compute_r_t(historical_case_counts):
 
     _, smoothed = prepare_cases(case_df, cutoff=25)
 
-    # Raise an error if there are no valid cases to use
-    if len(smoothed) == 0:
+    # Raise an error if there are not enough valid cases to use
+    # we need at least two days to represent change over time
+    if len(smoothed) < 2:
         raise ValueError('Input has too few cases to compute R(t)')
 
     # Note that we're fixing sigma to a value just for the example
--- a/backend/python-functions/realtime_rt.py
+++ b/backend/python-functions/realtime_rt.py
@@ -122,7 +122,7 @@ def compute_r_t(historical_case_counts):
     case_df.index = pd.to_datetime(case_df.index)
     case_df.index.name = 'date'
 
-    _, smoothed = prepare_cases(case_df, cutoff=25)
+    _, smoothed = prepare_cases(case_df, cutoff=10)
 
     # Raise an error if there are not enough valid cases to use
     # we need at least two days to represent change over time
--- a/src/impact-dashboard/ChartArea.tsx
+++ b/src/impact-dashboard/ChartArea.tsx
@@ -30,7 +30,7 @@ const LegendContainer = styled.div`
 `;
 
 const ChartArea: React.FC<{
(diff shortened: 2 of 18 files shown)


macfarlandian:
@jessex yeah i saw some pretty weird looking charts locally also, which I chalked up to not having realistic test data. The Rt model is much more sensitive to that than the previous curves were. 

That "Facility B" error is new to me, that's definitely one to look into if you can give me more info on how to repro (i.e. what your data looks like). We have a ticket #258 to improve python logging which will help us with cases like this but it's not implemented yet unfortunately.

There is a slightly fuzzy threshold of 25 _daily cases_  (not versions), below which the Rt model cannot produce a reasonable result and returns an error. That accounts for those errors that you're seeing. (@justkunz has been looking into trying to lower that threshold because it's probably quite limiting in practice.)

It's not yet clear to me how much of the weirdness we are seeing is due to bugginess in our application versus expected model behavior. I do agree though that we need a better error handling solution for these charts (it was a bit of a blind spot for me honestly, but I also didn't get any python errors ... you have a much more extensive set of test data than I do though).

jessex:
Summarizing issues we identified with help from @justkunz and a screenshare:
1. Filter out observed cases on the client side before sending data to the Python function, so that we only send one per day (the most recent one in a day)
2. Fix a bug in the Python function that causes a possible NaN result to be auto-converted to 0 and returned as an adequate result, leading to silly graphs -- instead an error should be returned by the function
3. Update the handling of model calculation/rendering so that if the R(t) result came back as an _error_, we fall back to displaying "classical" R(0) with just user inputs
  * We will want UI treatment for this so that the user is aware of what is being displayed and what is not being displayed, and why, but that will come in a follow-up issue after the next release
4. Optionally change the case difference threshold to a lower number to be more aggressive about when we rely on R(t) at the expense of accepting larger confidence intervals - needs to be discussed with @cawarren and confirmed here by @jessex

macfarlandian:
@jessex all four of those issues should be addressed with my latest push

PR #282, opened 2020-05-01
[Excalibur] response impact - reduction card data

### LINKED ISSUE #281: [Excalibur] Calculate data to pass into 4 cards for staff/inc hospitalized and fatalities
1. Based on original and current scenarios, determine the reduction in numbers.
2. Hook up to existing UI cards

rocktacular:
## Description of the change

1. Calculate hospitalizations and fatalities across all facilities for current scenario for both staff and incarcerated.
2. repeat for original scenario
3. pass data into cards

Most of the heavy lifting is in the `ResponseImpactCardStateUtils` file. `ResponseImpactDashboard` assembles and maintains the state for the data object. A couple other files I just needed to export some functions, and then the card container I needed to pass those props down and hook the data up.

## Type of change

- [X] New feature (non-breaking change that adds functionality)

## Related issues

Closes #281 

### Development

### Code review

### DIFF
--- a/src/page-response-impact/ResponseImpactDashboard.tsx
+++ b/src/page-response-impact/ResponseImpactDashboard.tsx
@@ -134,6 +134,9 @@ const ResponseImpactDashboard: React.FC = () => {
   const [releaseCardData, setReleaseCardData] = useState<
     releaseCardDataType | undefined
   >();
+  const [statePrisonPopulation, setStatePrisonPopulation] = useState<
+    number | undefined
+  >();
   const scenario = scenarioState.data;
   const [, setFacilities] = useState({
     data: [] as Facilities,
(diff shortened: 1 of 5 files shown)


rocktacular:
Here it is hooked up. Looks like there may be a rounding issue sometimes that causes to go to -1.

image

src/page-response-impact/utils/ResponseImpactCardStateUtils.tsx
```diff
@@ -0,0 +1,88 @@
...
+    hospitalized: number;
+    fatalities: number;
+  };
+  staff: {
+    hospitalized: number;
+    fatalities: number;
+  };
+};
+
+export function buildReductionData(
```
jovergaag, line 18: Beautiful

src/page-response-impact/ResponseImpactDashboard.tsx
```diff
@@ -152,6 +161,22 @@ const ResponseImpactDashboard: React.FC = () => {
     fetchFacilities();
   }, [scenarioState?.data?.id]);
 
+  // calculate data for cards
```
jovergaag, line 164: This component is starting to get pretty big. I don't think its necessary to do it in this PR but we should probably start thinking about breaking it up.

rocktacular: hah for sure. You should have seen it before I moved a couple functions/types to the Utils. 

Depending on the other components' needs, I might be able to move all of this down to the card container component.

rocktacular:
ok to resolve linting issues, I did these 2 things:
1. moved `scenarioState?.data?.id` to its own var `scenarioId`. linter did not like complex (multi-level) objects in the `useEffect` dependency arr.
2. move `fetchFacilities` function into the `useEffect` where it is called. linter asked to the fn into dep array, but then that caused an infinite loop, so linter suggested moving the function def inside the `useEffect`

PR #283, opened 2020-05-01
Adding Support for Duplicating Scenarios

### LINKED ISSUE #175: Add support for duplicating scenarios
We currently only support the baseline scenario in the web model, but ask that the user keep this model reflecting the real-world. This makes it difficult for the user to adequately forecast different scenarios and what-ifs (e.g., 'What if we were to release some of the low-risk members of the incarcerated population? What if we were to make space to move more people out of bunked housing and into cells?')

One of the more important features included in this is the ability to duplicate scenarios - that will make it easy to build a new scenario off of an existing baseline, rather than from scratch (which can be a lot of effort).

Add support for duplicating a scenario (including metadata and facilities) into a new scenario. This is meant to be a 'copy' operation - there is no expected long-term linkage or relationship between the new scenario and the old one, or the facilities in either one.

**Additional context**
- This is a child issue of issue #168 
- [Current mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=910%3A130)

sprad:
## Description of the change

Adding the persistence layer code to allow for duplicating scenarios

## Type of change

## Related issues

Closes #175 

### Development

Upon duplication a scenario I confirmed that:
1.  A new scenario was created with the requested name and description copy ✅ 
2.  All existing facilities and their modelInput versions were copied over ✅ 

Additional Cases:

1.  Tested that copying a Scenario without any facilities worked ✅ 
2.  Throwing an exception within the batch performs a rollback on the commit ✅ 

Performance:

I "performance tested" this by duplicating a Scenario that had 10 facilities and each of those facilities had 8 versions of modelInputs.  The duplication ran in about 2-3 seconds.  I think this should be fine for most cases, but we'll need to keep an eye on this if we expect users to add a significantly higher number of facilities and to be creating a lot of new versions (i.e. entering new case information frequently).

### Code review

### DIFF
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -19,7 +19,7 @@ service cloud.firestore {
       allow read, write: if userCanAccess(resource);
 
       match /{documents=**} {
-        allow read, write: if userCanAccess(get(/databases/$(database)/documents/scenarios/$(scenario)));
+        allow read, write: if userCanAccess(getAfter(/databases/$(database)/documents/scenarios/$(scenario)));
       }
     }
   }
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -5,10 +5,11 @@ import "firebase/auth";
 import "firebase/firestore";
 
 import createAuth0Client from "@auth0/auth0-spa-js";
-import { startOfDay, startOfToday } from "date-fns";
+import { format, startOfDay, startOfToday } from "date-fns";
 import { pick, sortBy } from "lodash";
 
 import config from "../auth/auth_config.json";
+import { MMMMdyyyy } from "../constants";
 import { persistedKeys } from "../impact-dashboard/EpidemicModelContext";
 import { Facility, ModelInputs, Scenario } from "../page-multi-facility/types";
 import {
@@ -257,11 +258,14 @@ export const saveScenario = async (scenario: any): Promise<Scenario | null> => {
 
 export const getFacilities = async (
   scenarioId: string,
-): Promise<Array<Facility> | null> => {
+): Promise<Facility[]> => {
   try {
     const scenario = await getScenario(scenarioId);
 
-    if (!scenario) return null;
+    if (!scenario) {
+      console.error(`No scenario found for scenario: ${scenarioId}`);
+      return [];
+    }
 
     const db = await getDb();
 
@@ -282,7 +286,7 @@ export const getFacilities = async (
 
     console.error(error);
 
-    return null;
+    return [];
   }
 };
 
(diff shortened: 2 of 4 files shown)


backend/firestore.rules
```diff
@@ -19,7 +19,7 @@ service cloud.firestore {
       allow read, write: if userCanAccess(resource);
 
       match /{documents=**} {
-        allow read, write: if userCanAccess(get(/databases/$(database)/documents/scenarios/$(scenario)));
+        allow read, write: if userCanAccess(getAfter(/databases/$(database)/documents/scenarios/$(scenario)));
```
sprad, line 22: This turned out to be the key to making this work.  `getAfter` allows the rules to be run within the context of a `batch`.  Without using `getAfter`, I kept getting permissions errors because the newly duplicated scenario did not exist yet and we were trying to add facilities to it.  Note:  all other existing functionality (i.e. non-batched writes like updating cases, changing names, toggling sidebar switches, etc.) still works when using `getAfter`.

src/database/index.tsx
```diff
@@ -257,11 +258,14 @@ export const saveScenario = async (scenario: any): Promise<Scenario | null> => {
 
 export const getFacilities = async (
   scenarioId: string,
-): Promise<Array<Facility> | null> => {
+): Promise<Facility[]> => {
```
sprad, line 261: Just cleaning up this interface to be a little more consistent with our other methods that return arrays

src/database/index.tsx
```diff
@@ -482,3 +486,75 @@ export const deleteFacility = async (
...
+    const batch = db.batch();
+
+    // Duplicate and save the Scenario
+    const scenarioDoc = db.collection(scenariosCollectionId).doc();
+
+    batch.set(scenarioDoc, {
+      name: `Copy of ${scenario.name}`,
+      description: `This is a copy of the '${
+        scenario.name
+      }' scenario, made on ${format(new Date(), MMMMdyyyy)}`,
```
sprad, line 513: Note: this `name` and `description` copy was provided by Andrew

src/database/index.tsx
```diff
@@ -482,3 +486,75 @@ export const deleteFacility = async (
...
+      createdAt: timestamp,
+      updatedAt: timestamp,
+    });
+
+    // Duplicate and save all of the Facilities
+    const facilities = await getFacilities(scenarioId);
+    for (const facility of facilities) {
+      const facilityCopy = Object.assign({}, facility);
+      delete facilityCopy.id;
+      delete facilityCopy.scenarioId;
```
sprad, line 530: Note that I am intentionally deleting the `facilityCopy`'s id so that the copied over facilities will get their own unique id.  However, I am intentionally leaving their original `createdAt` and `updatedAt` timestamps.

src/database/index.tsx
```diff
@@ -482,3 +486,75 @@ export const deleteFacility = async (
...
+      roles: {
+        [userId]: "owner",
+      },
+      createdAt: timestamp,
+      updatedAt: timestamp,
+    });
+
+    // Duplicate and save all of the Facilities
+    const facilities = await getFacilities(scenarioId);
+    for (const facility of facilities) {
```
jessex, line 527: To be clear, we're _not_ overwriting the `updatedAt` or `observedAt` timestamps on `modelVersions` because we want to preserve that chain of observations. Makes sense. Is the reason we are not updating the `facility.updatedAt` timestamp just to make sure we don't lose its history? Makes sense to me but want to be sure.

src/database/index.tsx
```diff
@@ -282,7 +286,7 @@ export const getFacilities = async (
 
     console.error(error);
 
-    return null;
+    return [];
```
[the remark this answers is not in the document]
sprad: Actually, I'll just revert this change so that we don't risk the regression and so that this PR maintains its singular focus

src/database/index.tsx
sprad: Correct, we are not overwriting the `updatedAt` or `observedAt` in `modelVersions` so that we can preserve the observation history.  This is significant so that we don't mess up R(t) calculations in these copies.  There isn't a strict use case for preserving the `facility.createdAt` and `facility.updatedAt` per se, but I chose to preserve them because I thought it would look a bit weird if the duplicated facilities all looked like they were updated at the time of duplication in the Facility Row instead of at the actual time they were last updated (see the image below):

Screen Shot 2020-05-01 at 4 24 18 PM

If that was an errant judgment call, I'm happy to drop these timestamps and generate new ones for the facility copies.

src/database/index.tsx
```diff
@@ -550,7 +547,6 @@ export const duplicateScenario = async (
 
     batch.commit();
   } catch (error) {
-    debugger;
```
sprad: 😳

PR #286, opened 2020-05-01
[Excalibur] Validate response impact data

### LINKED ISSUE #266: [Excalibur] Guard for consistent systemType/stateCode across facilities
On the `ResponseImpactDashboard`, check all facilities to ensure they have the same `systemType` and `stateCode` and issue a warning message to the user if not.

From [PRD](https://docs.google.com/document/d/1_jbpbsObUWkakyitmuIMaNKhq1YVU_fpo4j06ip8ets/edit?ts=5e95eb62#):

- Every facility has a staff AND population count
- All facilities have matching locale information
-- Same facility type (e.g., all are state prison or all are county jail)
-- Same state / county (e.g., will not work if one facility is in CA and the other is in OR)

**Additional context**

jovergaag:
## Description of the change

Validates response impact data to ensure that all facilities have the same `systemType` and `stateCode`. Also verifies that each facility has staff population and total incarcerated population greater than zero.

Displays an appropriate error message with invalid data.

## Type of change

## Related issues

Closes #266 

### Development

### Code review

### DIFF
--- a/src/page-response-impact/ValidDataWrapper.tsx
+++ b/src/page-response-impact/ValidDataWrapper.tsx
@@ -1,16 +1,38 @@
+import { sum } from "d3";
+import { pick } from "lodash";
 import React, { useEffect, useState } from "react";
 import styled from "styled-components";
 
 import Colors from "../design-system/Colors";
-import { Facilities } from "../page-multi-facility/types";
-// const POPULATION_DATA_ERROR_MSG = "Impact could not be generated because one or more of your facilities does not have a staff or incarcerated population count. Please add population numbers and try again."
+import { Facilities, Facility } from "../page-multi-facility/types";
+const POPULATION_DATA_ERROR_MSG =
+  "Impact could not be generated because one or more of your facilities does not have a staff or incarcerated population count. Please add population numbers and try again.";
 const LOCALE_DATA_ERROR_MSG =
   "Impact could not be generated because locale information does not match across facilities";
+
 const ValidDataWrapperContainer = styled.div``;
 const ErrorMessage = styled.div`
-  padding: 10px;
+  padding: 20px;
   color: ${Colors.darkForest};
+  width: 60%;
 `;
+
+function sumAgeGroupPopulations(facility: Facility): number {
+  return sum(
+    Object.values(
+      pick(facility.modelInputs, [
+        "age0Population",
+        "age20Population",
+        "age45Population",
+        "age55Population",
+        "age65Population",
+        "age75Population",
+        "age85Population",
+        "ageUnknownPopulation",
+      ]),
+    ),
+  );
+}
 
 interface Props {
   children?: React.ReactElement<any>;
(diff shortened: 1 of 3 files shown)


src/page-response-impact/ValidDataWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { pick } from "lodash";
+import React, { useEffect, useState } from "react";
+import styled from "styled-components";
+
+import Colors from "../design-system/Colors";
+import { Facilities, Facility } from "../page-multi-facility/types";
+const POPULATION_DATA_ERROR_MSG =
+  "Impact could not be generated because one or more of your facilities does not have a staff or incarcerated population count. Please add population numbers and try again.";
+const LOCALE_DATA_ERROR_MSG =
+  "Impact could not be generated because locale information does not match across facilities";
```
jessex, line 11: Nit: a period at the end of this one for consistency.

jessex: Did we get this copy from @sychang? If not, I'm wondering if users will immediately grasp the meaning of "locale information" or if we should be more explicit?

src/page-response-impact/ValidDataWrapper.tsx
```diff
@@ -0,0 +1,90 @@
+import { sum } from "d3";
+import { pick } from "lodash";
+import React, { useEffect, useState } from "react";
+import styled from "styled-components";
+
+import Colors from "../design-system/Colors";
+import { Facilities, Facility } from "../page-multi-facility/types";
+const POPULATION_DATA_ERROR_MSG =
+  "Impact could not be generated because one or more of your facilities does not have a staff or incarcerated population count. Please add population numbers and try again.";
```
jessex, line 9: Is it confirmed that we don't want to display the impact report if there are no staff numbers? It feels like we should still be able to generate a report given that the incarcerated population is significantly larger than the staff population, but if there are metrics that necessarily require both then that might make sense.

jessex: Alternatively, if this is partially meant to be a forcing function to encourage users to provide staff info, I get that, too!

src/page-response-impact/ValidDataWrapper.tsx
jovergaag: Yes, these came from the [PRD document](https://docs.google.com/document/d/1_jbpbsObUWkakyitmuIMaNKhq1YVU_fpo4j06ip8ets/edit?ts=5e95eb62#) and @sychang. But I agree with you though that we might need to be more explicit about what needs to be updated on the facilities. @sychang Any thoughts?

src/page-response-impact/ValidDataWrapper.tsx
jovergaag: The requirement to validate staff numbers was based on my recommendation, I thought that the model outputs were zero if there was no staff population input. As I'm playing around with it more this morning it looks like that is not the case, the zero outputs I was seeing must have been due to the prison population input. I removed the staff population check for now, @sychang if you want to keep it in we can just revert that commit.

src/page-response-impact/ValidDataWrapper.tsx
```diff
@@ -0,0 +1,71 @@
...
+        facilities[0].systemType !== f.systemType
+      );
+    });
+  }
+
+  function validatePopulationData(facilities: Facilities) {
+    return facilities.some((f) => {
+      return sumAgeGroupPopulations(f) === 0;
+    });
+  }
```
jovergaag, line 43: I chose to iterate over the facilities twice here just for clarify and ability to easily understand what the valid cases are since we are moving quickly and don't have tests. I don't really have a concept of how many facilities these systems have, so if this is causing a performance issue it will be easy to refactor.

src/page-response-impact/ValidDataWrapper.tsx
jessex: Yes that's not a requirement for the model, and poking around Firebase it looks like there are definitely some models where that input is not filled in. I'd suggest removing this particular validation.

PR #293, opened 2020-05-02
don't let initial exposed estimate exceed total population

### LINKED ISSUE #272: Negative Susceptible Population for High Case Counts
**Describe the bug**
When a facility inputs a high % of the population as infected the exposed/susceptible population counts are not initialized properly. The exposed population is computed to be too large (greater than the total population) and the susceptible population becomes negative.

**To Reproduce**

1. Go to the covid facility page
2. Enter in 495 cases for 500 population in any bucket (staff, age unknown, etc)
3. See the graph is weird

**Expected behavior**
First, the cases should be distributed into the compartments. Then only the infectious group (not total cases) should be included in the exposed group calculation. Finally, the exposed pop group can be computed as the min of `infectious * ratioExposedToInfected` and the remaining population.

**Additional context**
The code that needs to be altered is here: https://github.com/Recidiviz/covid19-dashboard/blob/master/src/infection-model/seir.tsx#L292

macfarlandian:
## Description of the change

We were extrapolating the number of initially exposed people from the number of initial cases, but failed to clamp the maximum value at the total population size, which could lead to negative values in the "susceptible" curve. This fixes that! Really it just ports over the fix from Python, where @justkunz had already implemented it.

## Type of change

## Related issues

Closes #272 

### Development

### Code review

### DIFF
--- a/src/infection-model/seir.tsx
+++ b/src/infection-model/seir.tsx
@@ -276,15 +276,19 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
   // assign people to initial states
   zip(ageGroupPopulations, ageGroupInitiallyInfected).forEach(
     ([pop, cases], index) => {
-      const exposed = cases * ratioExposedToInfected;
-      singleDayState.set(index, seirIndex.exposed, exposed);
-      singleDayState.set(index, seirIndex.susceptible, pop - cases - exposed);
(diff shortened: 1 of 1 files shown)


justkunz:
Thanks for the quick fix @macfarlandian! 🚀

PR #316, opened 2020-05-05
[R(t)] UX - Add R(t) number to Scenario View

### LINKED ISSUE #200: R(t) UX - Add R(t) number to Scenario View
We're now ready to start adding dynamic rate of spread! This means showing users not just the projections for their facilities, but (based on the past several days' observations) the current rate of spread in those facilities.

Rate of spread is a number from 0-7ish, where <1 indicates that the virus is spreading too slowly to expand its reach and will extinguish, while >1 indicates it will continue to spread (the larger the number, the faster it spreads). This visual treatment shows the current number for each facility.

**This issue** focuses on adding rate of spread information to the Scenario Details view.

- Note: This issue does not include adding the tabbed behavior to the Scenario Details view shown in the mocks, or the R(t) comparison view in the 'Rate of Spread' tab in the mocks. This issue is only about updating the list of facilities and their charts that exist today.
- Add the new 'Rate of spread' numbers in the upper-right of the individual facility charts on the main Scenario Details view
- Update the rest of the facility row to match the current style in the mocks (may be some mild changes in text wrapping/layout)
- Add tooltips for the rate of spread to explain this new UI feature (see Copy doc below)

**Additional context**
- This issue will depend on model support for R(t) (see issue #62) to launch.
- [Current mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=1636%3A130) - Note that this issue only covers the updates to the 'Scenario Details' mock, and **excludes** the 'Projections / Rate of Spread' tabs at the top.
- [Copy doc](https://docs.google.com/spreadsheets/d/1bnBsQXbtrRogruVkDPiAketsHUrv3H26ibw0ydh94D8/edit#gid=1594361286)

rememberlenny:
## Description of the change

When relevant, adding the latest RT value to the scenario view. Currently RT only displays when there are multiple days of data, so most of the RT values in my example are "?".

Based on discussion with @macfarlandian I'm not showing the RT value if the facilitydata is missing, but am adding a "?" if the facility data returns with a null value.

## Type of change

## Related issues

Closes #200

## Checklists

- [x] Adding tooltip

### Development

### Code review


src/page-multi-facility/FacilityRow.tsx
```diff
@@ -35,6 +38,27 @@ const groupStatus = {
...
+  [RateOfSpreadType.CONTROLLED]: Colors.green,
+  [RateOfSpreadType.INFECTIOUS]: Colors.darkRed,
+  [RateOfSpreadType.MISSING]: Colors.forest,
+};
+
+const pillBorderColor: { [x: string]: string } = {
+  [RateOfSpreadType.CONTROLLED]: Colors.teal,
+  [RateOfSpreadType.INFECTIOUS]: Colors.orange,
+  [RateOfSpreadType.MISSING]: Colors.forest30,
+};
```
macfarlandian, line 51: looks like you either don't need these or they should be part of the pill component

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -109,11 +140,18 @@ const FacilityRow: React.FC<Props> = ({
...
+
   const [facility, updateFacility] = useState(initialFacility);
-  let useRt, facilityRtData;
+  let useRt,
+    facilityRtData,
+    latestRt = null;
   if (useFlag(["useRt"])) {
     useRt = true;
-    facilityRtData = rtData ? rtData[facility.id] : undefined;
+    facilityRtData = rtData ? rtData[facility.id] : null;
```
macfarlandian, line 150: i don't think this is correct ... if the fetch has failed this value will already be null. if it's undefined that generally means a fetch has not been completed yet. Similarly, `rtData` should only really ever be undefined while the app is initializing and we haven't done the initial fetch yet (a "loading" state for practical purposes). 

I know it's not very explicit but we should try to stick to that general convention if possible to avoid making it even more confusing than it already is.

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -109,11 +140,18 @@ const FacilityRow: React.FC<Props> = ({
...
+
+    if (facilityRtData) {
+      latestRt = facilityRtData?.Rt[facilityRtData.Rt.length - 1].value;
```
macfarlandian, line 153: considering the conditional you shouldn't need `facilityRtData?` here?

macfarlandian: also see #273 , a helper function is forthcoming that should replace this assumption that the last value is always most recent (it should be a safe assumption for now but it's obviously brittle)

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -0,0 +1,62 @@
...
+import styled from "styled-components";
+
+import { RateOfSpreadType } from "../constants/EpidemicModel";
+import Colors from "../design-system/Colors";
+import PillCircle from "../design-system/PillCircle";
+
+const pillColors: {
+  [x: string]: {
+    [x: string]: string;
+  };
```
macfarlandian, line 11: would something more specific like 
```
{
  [key in RateofSpreadType]: {
    text: string;
    border: string;
  };
}
```
work here instead? (probably not exactly because I am bad at TypeScript but hopefully you get the general idea)

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -0,0 +1,62 @@
...
+    border: Colors.orange,
+  },
+  [RateOfSpreadType.MISSING]: {
+    text: Colors.forest,
+    border: Colors.forest30,
+  },
+};
+
+const RtValuePill = styled.div<{ spreadType: string }>`
+  .pill-circle {
```
macfarlandian, line 28: can do `${PillCircle} { ... }` instead of having to add this classname

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -0,0 +1,62 @@
...
+    text: Colors.darkRed,
+    border: Colors.orange,
+  },
+  [RateOfSpreadType.MISSING]: {
+    text: Colors.forest,
+    border: Colors.forest30,
+  },
+};
+
+const RtValuePill = styled.div<{ spreadType: string }>`
```
macfarlandian, line 27: require enum membership instead of arbitrary strings?

src/rt-timeseries/RtTimeseries.tsx
```diff
@@ -69,7 +68,7 @@ const Tooltip: React.FC<{
   return (
     <ChartTooltip>
       <TooltipContents>
-        <TooltipValue>{numeral(value).format("0.0")}</TooltipValue>
+        <TooltipValue>{value}</TooltipValue>
```
macfarlandian, line 71: why

src/constants/EpidemicModel.tsx
```diff
@@ -0,0 +1,5 @@
+export const RateOfSpreadType = {
+  CONTROLLED: "Controlled",
+  INFECTIOUS: "infectious",
```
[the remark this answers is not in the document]
rememberlenny: They should be consistent. Updated to all be lowercase here.

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -35,6 +38,27 @@ const groupStatus = {
   infectious: true,
 };
 
+const pillTextColor: { [x: string]: string } = {
```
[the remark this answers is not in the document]
rememberlenny: @jessex I'm looking at this now, and think the problem is fixed. I think there was some old code I must have deleted between this comment. 

Just to double check, this is no longer an issue right?

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: Your typescript > my typescript.

That looks good. I had to change RateOfSpreadType to an enum, but this is better.

src/page-multi-facility/FacilityRow.tsx
rememberlenny: Re: conditional, I thought that passed the typescript error, but I didnt realize that actually also works for the assignment.

rememberlenny: Adding a todo to update the helper when its ready.

src/page-multi-facility/FacilityRow.tsx
rememberlenny: Yep - deleted.

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: I knew something like that worked. Ill update the others too.

src/rt-timeseries/RtTimeseries.tsx
rememberlenny: Not sure how this happened. No changes should be here.

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: That didnt work...

macfarlandian: well that's mysterious ... https://styled-components.com/docs/advanced#referring-to-other-components is my source, possibly something is interfering with this?

src/page-multi-facility/FacilityRow.tsx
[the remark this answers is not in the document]
jessex: Yep! Good to go now.

src/page-multi-facility/FacilityRow.tsx
rememberlenny: Updated

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: Not sure what the issue was but its working now. Thanks!

rememberlenny:
@jessex Just added fixes to the points you mentioned!

src/page-multi-facility/FacilityRowRtValuePill.tsx
macfarlandian: it doesn't look like anything actually was changed here?

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -0,0 +1,111 @@
...
+    border: Colors.teal,
+  },
+  [RateOfSpreadType.Infectious]: {
+    text: Colors.darkRed,
+    border: Colors.orange,
+  },
+  [RateOfSpreadType.Missing]: {
+    text: Colors.forest,
+    border: Colors.forest30,
+  },
```
macfarlandian, line 30: you changed `RateOfSpreadType` t0 a numerical enum, so these are going to be keyed like `0: {}, 1: {}, 2: {}` ... not technically wrong but kind of weird, maybe not what you want?

src/constants/EpidemicModel.tsx
```diff
@@ -0,0 +1,5 @@
+export enum RateOfSpreadType {
+  "Controlled",
+  "Infectious",
+  "Missing",
+}
```
macfarlandian, line 5: mentioned elsewhere but this will create a numeric enum, not a string enum with the same interface as what this type looked like before. not sure this is necessarily what you wanted here

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -0,0 +1,111 @@
...
+    white-space: pre-wrap;
+  }
+`;
+
+interface Props {
+  latestRt: number | null | undefined;
+}
+
+const rtSpreadType = (latestRt: number | null | undefined) => {
+  if (!latestRt) {
```
macfarlandian, line 71: Rt can technically be zero so I think this will give you a false negative in that case

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return RateOfSpreadType.Missing;
+  } else if (latestRt > 1) {
+    return RateOfSpreadType.Infectious;
+  } else {
+    return RateOfSpreadType.Controlled;
+  }
+};
+
+const displayRtValue = (latestRt: number | null | undefined) => {
+  return !latestRt ? "?" : latestRt.toFixed(1);
```
macfarlandian, line 81: similar to above, `latestRt === 0` will trip this up. `numeral(latestRt).format("0.0") | "?"` should work because numeral will output null if it gets a non-numeric argument, we have been using that library for formatting throughout the site

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: Odd. I changed it but may have accidentally reverted through merging with master.

src/constants/EpidemicModel.tsx
rememberlenny: Misunderstood that. Fixed.

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: Fixed.

src/page-multi-facility/FacilityRowRtValuePill.tsx
rememberlenny: Nice. Updated.

src/page-multi-facility/FacilityRowRtValuePill.tsx
```diff
@@ -1,61 +1,112 @@
...
+    <>
+      <RtValuePill spreadType={rtSpreadType(latestRt)}>
+        <PillTooltip>
+          <ChartTooltip>
+            <TooltipContents>
+              <TooltipValue>
+                {!!latestRt ? hasDataTitle : needsDataTitle}
+              </TooltipValue>
+              <TooltipLabel>
+                {!!latestRt ? hasDataBody : needsDataBody}
```
macfarlandian, line 102: I did not flag this in the previous review, but this will also do the wrong thing if `latestRt` is zero

rememberlenny: Make fixing updates.

PR #318, opened 2020-05-05
Disable name editing in facilities list

### LINKED ISSUE #312: Remove editing of facility names in facility list
- We should keep the on-hover behavior (show pencil icon), but extend it to the entire area to the right of the case numbers and make the pencil icon bigger or replace with ‘click to edit’

- When the user clicks, we should take them to Facility Details view (not make the facility name editable in this view). The user will only be able to update facility names in the Facility Details view from now on. 

Part of #306

macfarlandian:
## Description of the change

This removes the textarea from `FacilityRow` and makes it so that hovering anywhere in the row's click target (except for the button that opens the Add Cases modal) makes the edit icon appear. Also makes said icon a bit larger, as suggested in the issue. Overall other than the slightly larger icon, nothing should look different, only the post-click behavior (navigating rather than entering an input field) is changed.

## Type of change

## Related issues

Closes #312 

### Development

### Code review

### DIFF
--- a/src/page-multi-facility/FacilityRow.tsx
+++ b/src/page-multi-facility/FacilityRow.tsx
@@ -109,21 +109,22 @@ const FacilityRow: React.FC<Props> = ({ facility: initialFacility }) => {
       <DataContainer className="flex flex-row mb-8 border-b">
         <div className="w-2/5 flex flex-col justify-between">
           <div className="flex flex-row h-full">
-            <div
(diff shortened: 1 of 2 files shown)


macfarlandian:
@SoumyaMahadevan this might overlap a little bit with the other click target stuff that I didn't quite realize was a separate issue, sorry if I clobbered something you are working on

PR #334, opened 2020-05-07
[Excalibur] Add modal to prompt user for date and total populations for staff & incarcerated

### LINKED ISSUE #296: [Excalibur] Prompt the user to input a date and starting population as a pop-up modal when the user clicks “show my impact”
PRD section: https://docs.google.com/document/d/1_jbpbsObUWkakyitmuIMaNKhq1YVU_fpo4j06ip8ets/edit#heading=h.7zyhlv618xu2

Mocks: https://www.figma.com/file/6z1QQW84a59Ul3c17PgWdV/%F0%9F%9A%80Recidiviz-CJStatus-(excalibur)?node-id=1837%3A652

daschi:
## Description of the change

Adds a modal in front of the Response Impact dashboard that asks users to update staff and incarcerated population numbers before looking at the generated impact report. 

Adds the following property to the scenario model: 
`populations: [
   { date: date, staffPopulation: number, incarceratedPopulation: number },
]`

## Type of change

## Related issues

Closes #296 

### Development

### Code review


src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  function handleSubmit() {
+    saveBaselinePopulations(populations);
+  }
+
+  return (
+    <>
+      <Text>
+        If you skip this step, we will assume a starting incarcerated population
+        of {defaultTotalPopulation}, which is derived from Vera data, and your
```
jessex, line 108: I wonder if it would be more precise/formal to say "which is derived from data from the Vera Institute of Justice"?

jessex: Assuming that copy is from @sychang, Serena have we vetted that with Clem or Andrew? I want to make sure we attribute things in the right way.

daschi:
@jessex I did chat with @sprad about it and talked about some future use cases we might want to prepare for that @sychang mentioned. Namely: 

1. Next time they click "Generate" the modal will still appear but their previously entered population could appear pre-filled.
2. (not in scope for now) In the future, we may want to be able to let them pull up different "versions" of the saved report comparing the impact to different points in time (e.g., before a big release vs. before anything covid related ever happened)
3. (not in scope for now) If they request impact models per facility, we'd give them the ability to enter even more info (original pop per facility) in which case we'd definitely want to save it because it gets to be a lot of information that we don't want them to need to re-enter each time

Looking back at our conversation I do see he recommended using the key name `totalPopulations` instead of just `populations`, so I'll update that name!

macfarlandian:
one concern I have is that there is no guarantee that these arbitrary population inputs are going to align with the population inputs to the curve functions. as long as that we're not going to do anything with this data that assumes they are the same then it should be fine.

daschi:
@macfarlandian I'm not sure I understand, but would it be better to rename the value on the scenario object to make it clearer that these population numbers are supposed to be used for the "original" projection curve charts only (vs the current curve charts which should use the sum of the modeled facilities' populations) ?

macfarlandian:
yeah sorry let me try to clarify ... I guess specifically what I would be worried about would be somehow combining one of these population figures with a set of age-bracketed user inputs from the model version history, or something like that. For example, getting the curve data from some bracket inputs and then reporting it as a percentage of this new population data (because all the curves may not add up to 100%).

src/page-response-impact/hooks.tsx
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  useEffect(() => {
+    if (modelInputs.length === 0) return;
+
+    // Reverse the populations array to get the most recently modelled populations
+    const {
+      staffPopulation: userInputStaffPopulation,
+      incarceratedPopulation: userInputIncarceratedPopulation,
+    } = reverse(scenarioPopulations)[0];
```
rocktacular, line 115: I'm getting an error here since I do not have any `scenarioPopulations` (`scenario.populations`). Do we need to have a check for its `length` or give it a default value? 

Or do we have a migration strategy for this? My facility data is pre scenario-UI.

rocktacular: when I provide some data for `scenarioPopulations`, I get an infinite looping of the the following 3 hooks:

- useSystemWideData
- useOriginalCurveData
- useReductionData

I believe it is due to lodash's `reverse` mutating the array in place https://lodash.com/docs/4.17.15#reverse while `scenarioPopulations` is also in the dependency array for this `useEffect` hook.

daschi:
@macfarlandian I think I understand the concern, but I'm sorry if I'm getting it wrong--For this feature we are comparing it to default/static values for cases for the original projection (though I think that might be changing after what @sychang mentioned in standup today). Would it be helpful to add a note somewhere that says we should not be using these population counts to compare current case counts?

src/page-response-impact/hooks.tsx
daschi: For 1. that's a good point, I will add a guard for that. 
For 2. good catch--I hadn't noticed that locally, I will fix that

Thanks for reviewing!

src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
sychang: Checked with Andrew, let's make it more generic to be "based on public data" for a couple reasons
1. If we change the source of data we don't have to change the copy
2. We want to avoid spotlighting any advocacy-focused groups directly in the tool

macfarlandian:
@daschi personally i'm not worried about that ... it seems fine, I also went and looked at the mocks and I don't really see anything to worry about at this point. just something to keep in mind as we continue to work with this data i guess

src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
sychang: Checked, let's make it more generic to be "based on public data" so if we change the source of data we don't have to change the copy.

PR #340, opened 2020-05-07
[Excalibur] Use current staff and incarcerated population as original projection inputs

### LINKED ISSUE #339: [Excalibur] Use current staff and incarcerated case count for original projection
Use the same staff and incarcerated case count for both the current and original projection on the Impact Dashboard.

**Additional context**
To reduce the number of variables and complexity between projections, the only variable inputs between the two should be population and rate of spread. Therefore, use use the case count sums from the current projection as inputs to the original projection.

jovergaag:
## Description of the change

Use the same staff and incarcerated case count for both the current and original projection on the Impact Dashboard. Previously we were using staff cases = 1 and incarcerated cases = 1.

## Type of change

## Related issues

Closes #339 

### Development

### Code review


src/page-response-impact/responseChartData.tsx
```diff
@@ -51,11 +56,16 @@ export function getCurveInputs(modelInputs: EpidemicModelState[]) {
...
+    staffCases,
+    incarceratedPopulation,
+    incarceratedCases,
+  } = systemWideData;
   return {
-    staffCases: 1,
+    staffCases: staffCases,
     staffPopulation: staffPopulation,
-    ageUnknownCases: 1,
+    ageUnknownCases: incarceratedCases,
```
jovergaag, line 68: @macfarlandian Question for you. Do the curve charts treat the different cases by age group inputs differently? So, would setting input `age20Cases=15, age85Cases=3` as inputs give a different result than `ageUnknownCases=18`? For our original projection inputs I'm just using the sum of incarcerated cases and assigning it to `ageUnknownCases` here but I'm starting to think that this approach is not going to produce a fair comparison to the actual scenario model inputs.

macfarlandian: yes they do .... they have different fatality rates and soon they will have different hospitalization rates as well.

Short of actually getting the age breakdowns for the "original" projection though, putting them all as age unknown is the best bet, since that uses rates that are averaged across all brackets. it's definitely going to be less accurate but it may be the best we can do with the data we have?

src/page-response-impact/ResponseImpactDashboard.tsx
```diff
@@ -117,18 +116,24 @@ const ResponseImpactDashboard: React.FC = () => {
...
+        localeDataSource,
+        modelInputs[0].stateCode,
+      ).totalPrisonPopulation;
+    } else {
+      incarceratedPopulation = getLocaleDefaults(
+        localeDataSource,
+        modelInputs[0].stateCode,
+        modelInputs[0].countyName,
+      ).totalJailPopulation;
+    }
```
daschi, line 132: Ah, is this something I should add to my PR too? If we `getLocaleDefaults` and pass it a countyName for a "State Prison" system type will the `totalPrisonPopulation` be incorrect?

jovergaag: Yes, if there is a `countyName` present but systemType is 'State Prison' then the population coming from the spreadsheet is incorrect. We will need to specify these individually.

src/page-response-impact/responseChartData.tsx
```diff
@@ -22,9 +22,8 @@ const NUM_DAYS = 90;
 const NUM_SEIR_CATEGORIES = 9;
 
 export type SystemWideData = {
+  [key: string]: number;
```
daschi, line 25: Should we make this type more specific and use the properties defined in the EpidemicModelContext?

jovergaag: That's what I had originally but Typescript would then not allow me to access the object via the key on line 97. I'm not sure the best way to approach this, I'm still trying to figure out Typescript.

jovergaag: I suppose I could avoid the Object.keys iteration and just explicitely sum each of the attributes in the `SystemWideData`. That might be easier to read anyway. Let me give that a try to see what it looks like.

daschi: Also, you could use the `Partial` utility like this:
```
export type SystemWideData = Partial<ModelInputsPersistent> & {
  hospitalBeds: number;
  incarceratedPopulation: number;
}

```

jovergaag: Cool! Good tip. I ended up using the `Partial` utility and it seems to be the right solution. I also left the other attributes in the declaration to make it clear what is being summed at the system level. Ready for another look.

daschi:
great! 🍰 🍵 ☕ 🌯

PR #342, opened 2020-05-08
Explicitly sets modelInputs.observedAt on the facility details page

jessex:
## Description of the change

As discussed in [this Slack thread](https://recidiviz.slack.com/archives/G010VACT8F2/p1588900629168100), there is a bug causing the `modelInputs.observedAt` date (and consequently the latest `modelVersions.observedAt` date) to be set to whatever is the most recent `observedAt` date in the existing list of model versions. I don't know if this is the most elegant fix, but it works in local testing. The approach is to eliminate any uncertainty or ambiguity and just set `observedAt` to right now when entering updates in this form, which is the desired behavior.

Local testing:
- Updated the facility through the facility input form before and after the change. The buggy behavior happens before the change and not after. After, the modelInputs and modelVersions observedAt fields are set to right now
- Tested the add cases modal after the change and was still able to successfully create an observation/version for a day in the past. It created a `modelVersions` with the right `observedAt` date and _did not_ update `facility.modelInputs`, all as expected. I think went back to the facility details page and made another update and it successfully updated `facility.modelInputs` and added a new `modelVersions`, both with right now as the `observedAt` date.

## Type of change

## Related issues

N/A

### Development

### Code review

### DIFF
--- a/src/page-multi-facility/FacilityInputForm.tsx
+++ b/src/page-multi-facility/FacilityInputForm.tsx
@@ -120,8 +120,8 @@ const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
 
   const save = () => {
     if (facilityName) {
-      // Set observedAt to right now when updating a facility form from this input form
-      const modelUpdate = model[0];
+      // Set observedAt to right now when updating a facility from this input form
+      const modelUpdate = Object.assign({}, model[0]);
       modelUpdate.observedAt = new Date();
 
       saveFacility(scenarioId, {
--- a/src/page-multi-facility/FacilityInputForm.tsx
+++ b/src/page-multi-facility/FacilityInputForm.tsx
@@ -120,12 +120,16 @@ const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
 
   const save = () => {
     if (facilityName) {
+      // Set observedAt to right now when updating a facility form from this input form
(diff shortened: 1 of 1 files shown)


src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -120,12 +120,16 @@ const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
 
   const save = () => {
     if (facilityName) {
+      // Set observedAt to right now when updating a facility form from this input form
+      const modelUpdate = model[0];
```
sprad, line 124: Do you need to copy the object in order not to accidentally mutate `model[0]` elsewhere?

```javascript
const modelUpdate = Object.assign({}, model[0]);
```

jessex: I hadn't noticed any adverse effect but I could have missed them. This makes great sense.

src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -120,12 +120,16 @@ const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
 
   const save = () => {
     if (facilityName) {
+      // Set observedAt to right now when updating a facility form from this input form
```
sprad, line 123: Nit: remove "form" after "facility"

PR #352, opened 2020-05-08
Add real data to the Rt comparison chart

### LINKED ISSUE #202: R(t) UX - Add new R(t) Comparison view, tabs to Scenario Details view
We're now ready to start adding dynamic rate of spread! This means showing users not just the projections for their facilities, but (based on the past several days' observations) the current rate of spread in those facilities.

Rate of spread is a number from 0-7ish, where <1 indicates that the virus is spreading too slowly to expand its reach and will extinguish, while >1 indicates it will continue to spread (the larger the number, the faster it spreads). This visual treatment shows the current number for each facility.

**This issue** focuses on adding a page where users can compare their current facilities by R(t), to see where they need to focus most attention.

- Implement the 'Projections' and 'Rate of spread' tabs in the Scenario Details view
- Add the large comparison chart + content on the 'Rate of Spread' tab 
- Take note of the 'insufficient data' (?) and 'no spread' (0) base cases 
- Add tooltips (see copy doc below)

**Additional context**
- This issue will depend on model support for R(t) (see issue #62) to launch.
- [Current mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=1636%3A130) - Note that this issue only covers the updates to the 'Facility Details' mock
- [Copy doc](https://docs.google.com/spreadsheets/d/1bnBsQXbtrRogruVkDPiAketsHUrv3H26ibw0ydh94D8/edit#gid=1594361286)

macfarlandian:
## Description of the change

Pipes our fetched Rt data into the comparison chart. Includes functionality to look backward in time at different snapshots of Rt (which I have interpreted as "the most recent Rt value as of X days ago", because that seemed to be the only reasonable implementation of what's in the mocks).

Since the initial load can take a while, rather than waiting to render the entire chart we are rendering it a row at a time as the data comes in, but overlaying the entire chart with a loading spinner until all the fetches are complete.

## Type of change

## Related issues

Closes #202 

### Development

### Code review

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "react-is": "^16.13.1",
     "react-popper": "^2.1.0",
     "react-spinners": "^0.8.1",
-    "react-tabs": "^3.1.0",
     "react-virtualized": "^9.21.2",
     "react-virtualized-auto-sizer": "^1.0.2",
     "regenerator-runtime": "^0.13.5",
--- a/package.json
+++ b/package.json
@@ -154,6 +154,7 @@
     "@types/react-tabs": "^2.3.1",
     "babel-jest": "^25.4.0",
     "jest": "^25.4.0",
+    "jest-date-mock": "^1.0.8",
     "jest-styled-components": "^7.0.2",
     "ts-jest": "^25.4.0"
   }
--- a/setupTests.tsx
+++ b/setupTests.tsx
@@ -1,2 +1,3 @@
 import "@testing-library/jest-dom";
 import "jest-styled-components";
+import "jest-date-mock";
--- a/src/constants/index.tsx
+++ b/src/constants/index.tsx
@@ -1,3 +1,10 @@
+import { Facilities } from "../page-multi-facility/types";
+
 // Date Format Constants
 export const MMMD = "MMM d";
 export const MMMMdyyyy = "MMMM d, yyyy";
+
+export type FetchedFacilities = {
+  data: Facilities;
+  loading: boolean;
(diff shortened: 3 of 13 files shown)


macfarlandian:
yeah i think i have identified the problem (the charts don't like being rendered while they're hidden) ... i have the fix in another branch but maybe I should open that PR first since it's so disruptive for testing. stand by

macfarlandian:
The tab rendering problem, so far as I can tell, was that they behave unpredictably when they are rendered inside a div with `display: none`, which we were doing as part of the tab-switching behavior. 

Rendering the tabs the normal way with `react-tabs` did indeed cause huge performance problems; something about the way it was animating transitions or something. As a result of that we were barely using the functionality of `react-tabs` anyway, so I just ripped it out. Switching tabs with regular old conditional render logic in React performs better than that (it's not great, but that is a more pervasive issue that I think has to do with how all the somewhat expensive curve calculations are done synchronously; hopefully we can live with that for now?)

macfarlandian:
they don't affect anything visually but they're just rendering junk; I meant to fix them and apparently forgot ... give me a few minutes, should be an easy fix

macfarlandian:
actually the ones you found @jessex are different from the ones I was thinking of ... I think they might be related to hot module reloading or something elsewhere in the application because I can't reproduce them any other way. anyway I fixed mine and i'm not worried about the rest

PR #363, opened 2020-05-08
Remove Icon Coupling from the InputNameWithIcon Component

### LINKED ISSUE #350: Remove Icon Coupling from the InputNameWithIcon Component
**Describe the bug**
Our current `InputNameWithIcon` [bakes in a toggle](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/design-system/InputNameWithIcon.tsx#L61) to display a folder icon when the passed in `showIcon` property is set to true.  This creates an issue for the new scenario library because the intention of this icon is to be used as a trigger for presenting the Scenario Library modal.  It doesn't make sense to put the Scenario Modal loading code in this otherwise generic component.

**To Reproduce**

N/A - mostly a code refactoring

**Expected behavior**

Rename the `InputNameWithIcon` to `InputName` and remove its functionality to show or hide the folder icon.  This will make the component itself more generic and will also give it parity with its sister `InputDescription` component.  After that refactor, separately include the folder icon in the Scenario Sidebar and add a click handler to it that triggers the display of the Scenario Library Modal.

**Screenshots**

The folder icon highlighted in red below should no longer be tied to the `InputNameWithIcon` and should instead exist independently such that it can trigger the Scenario Library modal when clicked:

Screen Shot 2020-05-08 at 3 04 01 PM

sprad:
## Description of the change

* Removing the Folder Icon coupling from the InputName in order to provide more UI flexibility in where and how we place this folder icon.  This is particularly relevant for the Scenario Library which needs to be triggered by the folder icon.

## Type of change

## Related issues

Closes #350

### Development

### Code review

### DIFF
--- a/src/design-system/InputName.tsx
+++ b/src/design-system/InputName.tsx
@@ -3,29 +3,24 @@ import styled from "styled-components";
 
 import Colors from "./Colors";
 import EditInPlace, { Props as EditInPlaceProps } from "./EditInPlace";
-import iconFolderSrc from "./icons/ic_folder.svg";
+
+interface NameLabelDivProps {
+  border?: boolean;
+}
 
 const borderStyle = `1px solid ${Colors.paleGreen}`;
 
-const NameLabelDiv = styled.label`
-  align-items: baseline;
+const NameLabelDiv = styled.label<NameLabelDivProps>`
   cursor: pointer;
   display: flex;
   flex-direction: row;
-  justify-content: space-between;
   padding-bottom: 15px;
-  border-bottom: ${borderStyle};
+  border-bottom: ${(props) => (props.border ? borderStyle : "none")};
   font-size: 24px;
   font-family: Libre Baskerville;
   font-weight: normal;
   letter-spacing: -0.06em;
   line-height: 24px;
-`;
-
-const IconFolder = styled.img`
-  display: inline;
-  flex: 0 0 auto;
-  margin-right: 12px;
 `;
 
 const Heading = styled.h1`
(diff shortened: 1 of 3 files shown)


sprad:
I'm not wild about this implementation, but it works and doesn't seem any better or worse than what we already had in place.

Screen Shot 2020-05-08 at 7 03 42 PM

src/design-system/InputName.tsx
```diff
@@ -46,19 +41,18 @@ interface Props
...
   persistChanges,
   setName,
-  showIcon,
+  border,
   ...passThruProps
 }) => {
   return (
-    <NameLabelDiv>
-      {showIcon && <IconFolder alt="folder" src={iconFolderSrc} />}
+    <NameLabelDiv border={border}>
```
sprad, line 55: We no longer embed this Folder Icon within this otherwise generic component.  However we do pass in a new property `border` to optionally display a bottom border if the UI calls for it (i.e. in the `FacilityInputForm` in this case).

src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -112,21 +117,30 @@ const ScenarioSidebar: React.FC<Props> = (props) => {
...
-          <ScenarioLibraryModal
-            trigger={<IconFolder alt="folder" src={iconFolderSrc} />}
+        <ScenarioName>
+          {showScenarioLibrary && (
+            <ScenarioLibraryModal
+              trigger={
+                <IconFolder
+                  style={{
+                    marginTop: "8px",
+                  }}
```
sprad, line 127: A side effect of embedding these trigger components within the modal component itself is that we have to do any custom styling inline.  I think that's ok.

Issue #360, opened 2020-05-08
[Excalibur] Pre-launch QA list


sychang:
**Is your feature request related to a problem? Please describe.**
An ongoing list of issues that come up from QA.

Issues that come up during QA are listed at the bottom of the PRD in this section: 
https://docs.google.com/document/d/1_jbpbsObUWkakyitmuIMaNKhq1YVU_fpo4j06ip8ets/edit#heading=h.ijifhwfwohic

Folks can put their name next to each title to assign to themselves and strikethrough when completed.

PR #379, opened 2020-05-11
[Excalibur] Make /impact layout more responsive

amyrlam:
## Description of the change

Went the "quick fix" route for now so don't have to change the rest of the markup. 

(The column markup is also used on the Add Facility page, which isn't really responsive at all for now, but this fix makes it a little better)

IS:
responsive2

## Type of change

## Related issues

Part of: #360

### Development

### Code review

### DIFF
--- a/src/design-system/PageColumn.tsx
+++ b/src/design-system/PageColumn.tsx
@@ -3,6 +3,10 @@ import styled from "styled-components";
 export const PageContainer = styled.div`
   display: flex;
   flex-direction: row;
+
+  @media (max-width: 700px) {
+    flex-direction: column;
+  }
 `;
 export const Column = styled.div<{ width?: string }>`
   margin: 20px;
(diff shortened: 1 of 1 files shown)


src/design-system/PageColumn.tsx
```diff
@@ -3,6 +3,10 @@ import styled from "styled-components";
 export const PageContainer = styled.div`
   display: flex;
   flex-direction: row;
+
+  @media (max-width: 700px) {
+    flex-direction: column;
+  }
 `;
 export const Column = styled.div<{ width?: string }>`
   margin: 20px;
```
daschi, line 12: I think we should also change the default width of the `Column` on line 13 to be `100%` within that same media query

amyrlam: @daschi good catch, thanks! How about 95% since we lose some of the right margin? (I eyeballed 95%, can see 100% effect in this gif below.)

responsive3

daschi: oh yeah--actually it might work better if you change it to `width: unset` and then it should fit the space I think

daschi: OK - `unset` doesn't work in IE11, but `inherit` does which I think would do the same thing here

amyrlam: 👍 nice, never thought of using inherit for width, that's cool! made the fix and looks good on facility page as well

PR #384, opened 2020-05-11
force sidebar re-render after scenario switch

### LINKED ISSUE #348: Properly Handle Scenario Change State Management
**Describe the bug**

When selecting a scenario within the Scenario library most of the application resets to use the new scenario's information, but the Name and Description in the Scenario Sidebar do not.

**To Reproduce**

0.  Make sure you have created at least two scenarios
1. From the home screen notice the name, description, toggles, facilities, etc. for the Scenario you are concurrently viewing.
2. Click on the Scenario Library folder and select a different scenario from the one that you are currently viewing.
3. Notice that the Scenario sidebar updates correctly for the toggles and last updated at date.  Also notice that the facilities associated with the clicked on scenario are also loaded.  However, notice that the Scenario's name and description in the sidebar are not updated to reflect the currently viewed scenario's information.

**Example**

Viewing my baseline scenario:

Screen Shot 2020-05-08 at 2 41 14 PM

Switching to another (non-baseline) scenario

Screen Shot 2020-05-08 at 2 41 23 PM

Again, notice that all of the expected facilities and attributes in the sidebar are loaded as expected except the name and description.  Note:  The updated date happens to be the same in both scenarios, but I've confirmed that this value does change when navigating through scenarios with different updated dates.

macfarlandian:
## Description of the change

Fixes an issue where the scenario name and description were not being updated in the scenario sidebar after changing scenarios.

The scenario sidebar and its descendants have accumulated some convoluted state management logic that contributed to this bug, and this PR does not do anything to improve that; instead I had to just force its entire component tree to re-render when the scenario changes, which forces all the form inputs to re-initialize.

## Type of change

## Related issues

Closes #348 

### Development

### Code review

### DIFF
--- a/src/page-multi-facility/ScenarioSidebar.tsx
+++ b/src/page-multi-facility/ScenarioSidebar.tsx
@@ -1,6 +1,6 @@
 import { format } from "date-fns";
 import { navigate } from "gatsby";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import styled from "styled-components";
 
 import { saveScenario } from "../database";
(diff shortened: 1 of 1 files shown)


macfarlandian:
yeah I think that would be nice but it would require some significant surgery ... all of the calculations that are currently done synchronously when we render a row would have to become async with some kind of loading state (either per row or per scenario; per row is probably easier). we should make an issue and come back to it

PR #387, opened 2020-05-11
Change firebase query to use composite index for modelVersions. 

### LINKED ISSUE #254: R(t) Calc - handle multiple observations on same day
When this happens we will want to order or filter by this secondary field (confirmation on that TBD, based on calc function requirements)

Likely need to add a composite index for "observedAt" and "updatedAt" fields in Firestore for querying/ordering multiple versions on the same day. 

We are currently doing the filtering post-fetch on the client side by ordering the fetched records as part of the `getRtInputsForFacility` function, as it is required to prevent errors from our Rt calculation function. Once we can perform this ordering and filtering in the database itself, the database function (currently named `getFacilityModelVersions`, though that should probably changed) should be updated to issue a query that returns just one record per day (the most recently updated one) and all of the client-side uniqueness and ordering code in `getRtInputsForFacility` should be removed.

See discussion at https://github.com/Recidiviz/covid19-dashboard/commit/191e4b09ad699b7941fa012b94b1ab9e58362085#r38825212

SoumyaMahadevan:
## Description of the change
- Add a new composite index for modelVersions sub collection.
- Changed getFacilityModelVersions to order by observedAt and updatedAt to account for multiple observations on the same day. observedAt is sorted asc and updatedAt is sorted desc.
- Removed the extraneous orderBy done in getRtInputsForFacility, since this is taken care of in the query itself.

- Verified that getRtInputsForFacility returns the same o/p before and after the change.
- No index errors on console due to using the composite index.

E.g. o/p of getRtInputsForFacility selecting only the most recent updatedAt for every observedAt.

image

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

Closes #254

### Development

### Code review

### DIFF
--- a/src/infection-model/rt.tsx
+++ b/src/infection-model/rt.tsx
@@ -79,15 +79,8 @@ const getRtInputsForFacility = async (
   let modelVersions = await getFacilityModelVersions({
     scenarioId: facility.scenarioId,
     facilityId: facility.id,
+    distinctByObservedAt: true,
   });
-
-  // inputs must contain no more than one record per day.
-  // when we have multiples, filter out all but the most recently updated
-  modelVersions = uniqBy(
-    modelVersions,
-    // make sure time doesn't enter into the uniqueness
-    (record) => record.observedAt.toDateString(),
-  );
 
   const cases: number[] = [];
   const dates: string[] = [];
--- a/src/infection-model/rt.tsx
+++ b/src/infection-model/rt.tsx
@@ -83,16 +83,8 @@ const getRtInputsForFacility = async (
 
   // inputs must contain no more than one record per day.
   // when we have multiples, filter out all but the most recently updated
-  // TODO: it may be possible to do this filtering directly in the query
-  // with a compound index, see #254 and replace this code if possible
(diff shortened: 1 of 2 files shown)


src/infection-model/rt.tsx
```diff
@@ -83,16 +83,8 @@ const getRtInputsForFacility = async (
...
-  // with a compound index, see #254 and replace this code if possible
   modelVersions = uniqBy(
-    orderBy(
-      modelVersions,
-      ["observedAt", "updatedAt"],
-      // uniqBy retains only the first item when it encounters duplicates,
-      // so make sure the most recently updated one is first
-      ["asc", "desc"],
-    ),
+    modelVersions,
```
macfarlandian, line 87: I think I would rather have the uniqueness filter move out of this function as well; it depends on the inputs being in the right order (and that order is somewhat counterintuitive), so having the two transformations be so distant from one another makes me worry this will be prone to breakage.

SoumyaMahadevan:
Updated getFacilityModelVersions to include uniqBy functionality if passed as param.
image

PR #401, opened 2020-05-12
[Excalibur] Update copy for the hospital beds at peak metric to specify incarcerated

### LINKED ISSUE #400: [Excalibur] Hospital beds used numbers inconsistent between charts/card
**Describe the bug**
The number of hospital beds used for the original and current projection charts are inconsistent with the 'Hospital beds used at peak reduced by' in the card at the bottom of the page.

**To Reproduce**

1. On the response impact page, hover over the original curve chart at the peak hospitalized and note the number. Do the same for current.
2. Compute the Hospital beds used at peak 'recued by' percentage: `(orig - curr) / orig`
3. Note that your calculation is different than what is shown in the percentage in the card at the bottom.

**Expected behavior**
These values should match

**Screenshots**

- OS [e.g. Windows 10]:
- Browser [e.g. chrome, edge]:
- Version [e.g. 22]:

**Smartphone (please complete the following information):**

- Device: [e.g. iPhone6]:
- OS: [e.g. iOS8.1]:
- Browser [e.g. stock browser, safari]:
- Version [e.g. 22]:

**Additional context**
I think that the issue lies in the curve charts and not in the calculation for the card. I have traced the issue back to `getCurveChartData` in `responseChartData.tsx`. I believe that the calculation of hospitalized on lines 160 and 161 are not quite right. One way to check this out is to look at the curve chart data for a single facility, and compare it to the curve chart on that facility's edit page to ensure the values match up. Then sum the hospitalized values for each facility per day to make sure we're doing the sum correctly.

The other possible source of the bug is the `getHospitalizedForDay` call on line 129 of `ResponseImpactCardStateUtils.tsx`. Is that function not returning what we think it is?

daschi:
## Description of the change

Update the copy for the hospital beds used at peak card at the bottom of the page so that it is clearer that it is calculated by incarcerated projection numbers (as opposed to the curve charts which sum both incarcerated and staff numbers)

image

image (1)

## Type of change

## Related issues

Closes #400 

### Development

### Code review

### DIFF
--- a/src/page-response-impact/ReducingR0ImpactMetrics.tsx
+++ b/src/page-response-impact/ReducingR0ImpactMetrics.tsx
@@ -36,7 +36,7 @@ const ReducingR0ImpactMetrics: React.FC<Props> = ({ populationImpact }) => {
       <ImpactMetricCard
         title={
           <HospitalBedsTitle
-            title="Hospital beds used at peak"
+            title="Hospital beds used by incarcerated"
             value={populationImpact.hospitalBedsUsed}
           />
         }
@@ -44,7 +44,7 @@ const ReducingR0ImpactMetrics: React.FC<Props> = ({ populationImpact }) => {
         icon={hospitalIcon}
       />
       <ImpactMetricCard
-        title={`Additional staff ${staffAbilityText} to work at peak`}
+        title={`Additional staff ${staffAbilityText} to work`}
         value={formatAbsValue(populationImpact.staffUnableToWork)}
         icon={staffIcon}
       />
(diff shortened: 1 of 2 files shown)


src/page-response-impact/ReducingR0ImpactMetrics.tsx
```diff
@@ -36,15 +36,15 @@ const ReducingR0ImpactMetrics: React.FC<Props> = ({ populationImpact }) => {
...
+            title="Hospital beds used by incarcerated"
             value={populationImpact.hospitalBedsUsed}
           />
         }
         value={`${formatAbsValue(populationImpact.hospitalBedsUsed)}%`}
         icon={hospitalIcon}
       />
       <ImpactMetricCard
-        title={`Additional staff ${staffAbilityText} to work at peak`}
+        title={`Additional staff ${staffAbilityText} to work`}
```
jovergaag, line 47: Are these the latest and greatest copy suggestions? @rocktacular merged a PR this morning adding the 'at peak' copy. Is this the final?

daschi: yeah after chatting with @sychang we were trying to find a way to shorten the card text to include `incarcerated` and she had an idea of moving `peak` to the heading instead. I'm still waiting for Serena to confirm these screenshots

Issue #320, opened 2020-05-05
R(t) and case count state management improvements


jessex:
**Describe the bug**
Updates to case counts, meant to update R(t) calculations, do not update the R(t) calculations and output immediately. Instead, they tend to require a full refresh of the scenario details page (or navigating away and back to a facility details page, since refresh is not supported on facility details) in order to regenerate R(t) output.

The summary is that we should recalculate R(t) output eagerly, when case observations are made, to avoid potentially misleading UX.

**To Reproduce**

Some example flows that are unexpected right now:

1. Click on the case number on a facility row in the scenario details page to open the add cases modal
1. Add a new observed day case count and click save
1. Observe that the R(t) pill for that facility row has stayed the same, and that the R(0) chart (which takes R(t) as input) has stayed the same (when the Rate of Spread tab is built, it will presumably also stay the same, as the surrounding context is currently implemented)
1. Refresh the page and see that the R(t) calculations are re-performed and updated appropriately

Ideally, what would have happened here would be that after clicking save on the add cases modal, the R(t) calculations for that specific facility row would have been re-performed and output in the facility row re-rendered with the updated numbers.

A similar example:
1. Click on a facility row in the scenario details page to open the facility details page for that row
1. Click the add cases button near the top left to open the add cases modal, add a new observed day case count and click save
1. Similarly as before, observe that the R(t) chart and the R(0) chart are both unchanged
1. You might be tempted to refresh the page here to see the updated charts, but refresh on facility details just opens it up as a blank page for creating a new facility
1. Instead, to see the updated output, you would need to 1) navigate back to the scenario details page, and 2) refresh the scenario details page to see the updates (however, before refreshing the details page in sub-step 2, you would have seen the updated case count, just not the updated chart or R(t) pill)

Again, ideally what would have happened is that after clicking save on the add cases modal, even from the facility details page, the R(t) calculation would be re-performed and the two charts re-rendered after clicking save.

**Expected behavior**
As described above, when clicking _save_ on the add cases modal (from either the scenario details or facility details pages) or when clicking save on the facility details page, R(t) calculations should be re-performed and components when display its output should be re-rendered.

PR #408, opened 2020-05-13
Update RT value when new case is added on dashboard or facility page

rememberlenny:
## Description of the change

Currently, when a new day of case data is added, the RT value is not refreshing. To see the updated value, the page needs to be manually refreshed. This causes problems especially on the facility page, where refreshing results in clearing the page store and displays an unexpected empty page.

The updated values received from calculate_rt endpoint are now used to update the FacilityContext, and update the frontend accordingly.

In a possible later change, a "loading" indicator might be worth adding while the updated value is received.

## Type of change

## Related issues

Closes [#320]

### Development

### Code review

### DIFF
--- a/src/infection-model/rt.tsx
+++ b/src/infection-model/rt.tsx
@@ -2,6 +2,7 @@ import { ascending } from "d3-array";
 import { formatISO, fromUnixTime, parseISO } from "date-fns";
 import { mapValues, maxBy, minBy, orderBy, uniqBy } from "lodash";
 
+import { FacilityEvents } from "../constants/dispatchEvents";
 import { getFacilityModelVersions } from "../database";
 import { totalConfirmedCases } from "../impact-dashboard/EpidemicModelContext";
 import { Facility } from "../page-multi-facility/types";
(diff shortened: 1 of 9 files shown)


src/page-multi-facility/FacilityRow.tsx
```diff
@@ -12,6 +12,7 @@ import Tooltip from "../design-system/Tooltip";
 import CurveChartContainer from "../impact-dashboard/CurveChartContainer";
 import { totalConfirmedCases } from "../impact-dashboard/EpidemicModelContext";
 import useModel from "../impact-dashboard/useModel";
+import { getUpdatedFacilityRtData, RtData } from "../infection-model/rt";
```
jessex, line 15: Whoops. Remember to remove `RtData` from this import before merging.

src/infection-model/rt.tsx
```diff
@@ -126,6 +127,21 @@ export const getRtDataForFacility = async (
   }
 };
 
+export async function getUpdatedFacilityRtData(
```
macfarlandian, line 130: nitpicking but I think a function that starts with "get" implies that it will return something; would just call this "updateFacilityRtData" or something like that

rememberlenny: Updated

Issue #372, opened 2020-05-11
Merlin - 5/11 Fit 'n finish polish round 


cawarren:
**Is your feature request related to a problem? Please describe.**
Several minor polish issues to re-align with mocks.

**Describe the solution you'd like**
Please
- Identify an issue from [the doc](https://docs.google.com/document/d/15oREP0hXoQSnzL-mMp3fTYRLyomyhskLD38-l2t1v_E/edit)
- Put your name next to it in the doc
- File an issue referring back to this one to work out of
- Cross off the item in the doc when you're done

If you have any questions at all, ping @cawarren or @sychang 

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

sychang:
Moved issue back to "To Do" so other folks can continue hoping into the doc and grabbing issues

amyrlam:
@sychang ah makes sense now, my bad!

amyrlam:
I worked on a couple things in the Google Doc. For clarity, subtasks of this issue are open to anyone and I'll assign myself to something in the Google Doc if move back to this!

amyrlam:
Took a new task from the Google Doc

PR #418, opened 2020-05-14
[Merlin] Make loading spinner appear in center of row

amyrlam:
## Description of the change

To achieve this, make row full-height on load and account for padding

WAS:
spinner-was

IS:
spinner

👉 Please let me know if this is not the right approach. I tried adding a conditional prop to `LoadingContainer` [here](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/design-system/Loading.tsx#L7-L13) but got stuck / couldn't achieve the same result. If not the right approach, any similar components here I could look at? 

FWIW it's good on mobile as-is, seems like the height of the chart is constant.

## Type of change

## Related issues

Part of https://github.com/Recidiviz/covid19-dashboard/issues/372 cc @cawarren @sychang 

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/CurveChartContainer.tsx
+++ b/src/impact-dashboard/CurveChartContainer.tsx
@@ -1,23 +1,11 @@
 import isEmpty from "lodash/isEmpty";
 import React from "react";
 import { useEffect, useState } from "react";
-import BounceLoader from "react-spinners/BounceLoader";
-import styled from "styled-components";
 
-import Colors from "../design-system/Colors";
+import Loading from "../design-system/Loading";
 import { MarkColors } from "./ChartArea";
 import CurveChart, { ChartData } from "./CurveChart";
 import { useEpidemicModelState } from "./EpidemicModelContext";
-
-const LoadingContainerCustom = styled.div`
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  width: 100%;
-  height: 100%;
-  min-height: 177px;
-  padding-bottom: 32px;
-`;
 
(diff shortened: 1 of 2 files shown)


src/impact-dashboard/CurveChartContainer.tsx
```diff
@@ -48,7 +60,9 @@ const CurveChartContainer: React.FC<Props> = ({
   }, [groupStatus, curveData]);
 
   return !curveDataFiltered ? (
-    <Loading />
+    <LoadingContainerCustom>
```
jovergaag, line 63: For consistency sake, I think it would be best to keep the `<Loading />` component here and just pass in some custom styles. For an example of how to do this you can look at `InputButton.tsx`. The `margin-top` is a conditional style based on the props. You can do something like that for the min-height and padding-bottom here. Does that make sense

cawarren:
This looks really good, @amyrlam - thanks!

src/impact-dashboard/CurveChartContainer.tsx
amyrlam: @jovergaag thanks for the example! Will give that a try and report back

amyrlam: @jovergaag thanks that was super helpful, learned something! 🙏

I was trying to follow https://github.com/Recidiviz/covid19-dashboard/blob/master/src/design-system/Spacer.tsx or https://styled-components.com/docs/basics#extending-styles before 

let me know what you think

Issue #404, opened 2020-05-13
Viewing historical case data on the facility detail page


sychang:
**Is your feature request related to a problem? Please describe.**
Currently, users have no way of viewing the historical case and population data they have entered for a facility. This causes problems because it is difficult to debug a faulty R(t) and hard to know where they are missing data. We also want to encourage them to give us a complete set of data so we can give them the most accurate projection.

**Describe the solution you'd like**
Because they add cases on a per-facility level, we want to add a chart that allows them to view the case data they have inputted historically and easily identify where they are missing data.

![image](https://user-images.githubusercontent.com/6633887/81773381-c3981b00-94ad-11ea-9e79-a442cf68ca9c.png)

![image](https://user-images.githubusercontent.com/6633887/81773410-cdba1980-94ad-11ea-93aa-928e97e8a907.png)

Components

1. Bar chart by default shows the most recent 30 days. 
2. Clicking “Previous 15 days” scrolls the chart to be able to show the previous 15 days of data (if there is an easier chart library to handle scrolling back in time, feel free to use that)
3. Each bar is a click target that opens the “add historical data” modal with the date of the bar pre-selected
4. When there’s no data, the bar is red and stretches the full y-axis length of the chart (On click, opens the “Add historical data” modal with the date pre-selected to the date of the bar. On hover, color gets darker to show that the bar is a click target. Clicking on the hover text should also open the modal)
5. When there is data for a day, the bar is green and has a height of the # of confirmed cases (On click, opens the “Add historical data” modal with the date pre-selected to the date of the bar, with data pre-populated from what the user previously entered). On hover, shows the date, case count, and a CTA to click to change the case count. Click on the hover text should also open the modal)
6. P2: If a case count is 0 or less than 50% between the first y-axis line and 0, display the count above the bar
7. Button at the top right “+ Add Historical Data” to open the “add historical data” modal with yesterday’s date pre-selected (or the default we are currently using)

Mocks: https://www.figma.com/file/6z1QQW84a59Ul3c17PgWdV/%F0%9F%9A%80Recidiviz-CJStatus-(excalibur)?node-id=2206%3A6804

**Describe alternatives you've considered**
We considered showing a table or calendar widget that indicates historical data and data missing. A bar chart lets them see the trend over time. A line chart was also considered, but it is more challenging to interact with.

**Additional context**
[...]

PR #420, opened 2020-05-14
Show missing and historical data in Add Cases modal

### LINKED ISSUE #395: Update adding cases modal calendar widget to visually show when data has been added or not
We currently don't allow users to see where they have previously inputted historical case data. We want to update the calendar picker to be more informative about what data is missing where.

1. Make all the days the same color (currently, Sat and Sun are red but that will become confusing when we start introducing colors that actually indicate something).
image

2. Update the background color for each day with red if they have not manually entered data for that day. No background color if they have inputted data for that day. 

3. Upon selecting a day that they have entered data for, populate the modal with the data they previously entered to allow them to change the data. 

Edit: Some updated mocks, to discuss with @macfarlandian today:
image

Showing more complex set of information on the actual calendar itself to indicate what information is missing - opting for keeping the calendar design simple for now.

**Additional context**
Out of scope for now, but we may introduce a new color when we have a "best guess" at the data for a day given public data that we want them to confirm.

macfarlandian:
## Description of the change

Adds some data and UI cues to the date picker in the Add Cases modal to alert you to which days you are missing data for, and to pre-fill the form with historical data when selecting a day for which some data already exists. To do this I had to replace the existing diffing logic because it didn't really work for this new data handling strategy; it was partially complicated by some overly greedy type definitions, so factoring out the age brackets into their own subtype helped with that somewhat.

Screen Shot 2020-05-14 at 12 26 55 PM

Includes a reusable hook for fetching the model history versions (as discussed with @daschi as #404 will need this data also). This hook doesn't use any persistent storage to prevent redundant fetches, but it does at least encapsulate the fetch logic. Fetches shouldn't be super expensive anyway so hopefully this won't be a big deal in practice. (I was going to try to refactor the Rt fetch function to use it, but it didn't play nicely with the conditional fetch behavior in there so I left that alone for now.)

## Type of change

## Related issues

Closes #395 

### Development

### Code review

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -3,4 +3,3 @@
 # (actual deployment builds will use Now configuration)
 
 GATSBY_SHOW_IMPACT_BUTTON="false"
-GATSBY_SHOW_SCENARIO_LIBRARY="false"
--- a/src/add-cases-modal/AddCasesModal.tsx
+++ b/src/add-cases-modal/AddCasesModal.tsx
@@ -10,8 +10,8 @@ import InputButton from "../design-system/InputButton";
 import InputDate from "../design-system/InputDate";
 import Modal, { Props as ModalProps } from "../design-system/Modal";
 import {
-  EpidemicModelUpdate,
-  persistedKeys,
+  ModelInputsPopulationBrackets,
+  populationBracketKeys,
 } from "../impact-dashboard/EpidemicModelContext";
 import { AgeGroupGrid } from "../impact-dashboard/FacilityInformation";
 import useModel from "../impact-dashboard/useModel";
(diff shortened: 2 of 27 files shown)


macfarlandian:
@jessex I think those "quirks" are also both bugs ... the expected behavior, for better or worse, is that a) it should always reset to the latest date for which there is data, and b) it should use that day's data to pre-fill any date without data when you select it. 

the state management here is quite convoluted ... this seemed to be working at some point but I may have caused a regression when merging master or something. i'll let you know when these cases are resolved.

macfarlandian:
@jessex I believe all these issues should be resolved now (they happened to share a root cause, which was inadvertently setting historical data to the facility's "current" inputs even if they were older than what was previously considered current).

PR #421, opened 2020-05-14
Filter first and duplicate records out of Rt function

### LINKED ISSUE #292: R(t) Timeseries: don't plot 0.0 value that always appears on the first day
currently the first day's value returned from the Rt function is not really "valid" since the values are based on difference from the previous day, and there is no previous day. The best solution is probably to simply not return this value from the function; confirm this approach with PM/DS  before proceeding

macfarlandian:
## Description of the change

Makes two changes to the Rt function requested by @justkunz: 
- drop the first day's Rt value because it is not meaningful and will always be 0 
- drop duplicate case counts in consecutive input records

The difference this produces can be quite dramatic – I tested it with some real-ish data from our aggregated dataset.

Current prod:
Screen Shot 2020-05-14 at 4 02 54 PM

After dropping the first day (same values but much more readable):
Screen Shot 2020-05-14 at 3 32 02 PM

After filtering out duplicates:
Screen Shot 2020-05-14 at 3 55 02 PM

No frontend changes were necessary, the Rt timeseries component already handles this and ensures there are at least two data points returned from the server before attempting to display the chart. (The other Rt features aren't affected by this because they only care about one value at a time.)

## Type of change

## Related issues

Closes #292 

### Development

### Code review

### DIFF
--- a/backend/python-functions/realtime_rt.py
+++ b/backend/python-functions/realtime_rt.py
@@ -139,5 +139,6 @@ def compute_r_t(historical_case_counts):
         raise ValueError('Unable to compute R(t) with the provided data')
 
     most_likely = posteriors.idxmax().rename('ML')
+    # the first day is not a valid value because it has no priors; exclude it
     result = pd.concat([most_likely, hdis], axis=1).iloc[1:]
     return result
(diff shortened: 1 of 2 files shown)


macfarlandian:
not sure but I will gladly wait for him to do so here

PR #427, opened 2020-05-15
System summary statistics - MultiFacilityPage

### LINKED ISSUE #415: Add summary stats to the homepage
Currently, our state engagement teams runs a daily manual report for a handful of states that requires painstaking manual analysis and aggregation. We want to do 3 things with this change:
1. Remove a layer of manual work that our team has to do daily.
2. Encourage states to keep their data up to date and make it very obvious when data has gone stale. If the system-level summary statistics are off, this will prompt the researcher to investigate which facility has stale data and update the data so the data is up-to-date with reality.
3. Start pulling some impact analysis into the forefront of view, also as a way to encourage users to add more data.

On the facility library view, add 5 summary statistics to the top of the page (note, all these stats are for **the current day**)
1. Total resident population: This is the sum of all the incarcerated individuals across all age brackets across all facilities that are currently modeled.
2. Total staff population: The sum of all staff across all the facilities that are current modeled
3. Total resident cases: The sum of all incarcerated cases across all age brackets across all facilities that are currently modeled.
4. Total staff cases: The sum of all staff cases across all the facilities that are current modeled.
5. Number of facilities with confirmed cases: The count of facilities that currently have a non-zero case count, displayed relative to the total (e.g., 5 of 10)
6. Number of facilities with R(t) < 1 currently: The count of all facilities with an R(t) < 1 out of all facilities, even those who do not have an R(t) value calculated (e.g., 1 of 10). This starts pulling in excalibur's impact analysis more into the forefront of view and encourages people to add cases to get an R(t) because if they don't, it will say "0 of 10"

Mock:
image

There is an alternate placement in the left side bar. However, we anticipate more information being added to the summary projection over time (e.g., projections and historical cases for the whole system) and the left panel has too little screen real estate to allow us to add much more information.

**Additional context**
This is a piece of 2 workstreams: 1) Automating the daily reports we currently build manually, 2) encouraging states to come back and enter up-to-date information.

jovergaag:
## Description of the change

Adds the 'System Summary' section at the top of the multi-facility page. 

-Calculates system wide population, case, and Rt metrics
-moves `isRtData` check to `infection-model/rt.tsx`
-adds `Metric` component in the design system

Screen Shot 2020-05-15 at 1 21 36 PM

## Type of change

## Related issues

Closes #415 

### Development

### Code review


src/page-multi-facility/SystemSummary.tsx
```diff
@@ -0,0 +1,129 @@
...
+      label: "Total resident cases",
+    },
+    staffCases: {
+      value: 0,
+      label: "Total staff cases",
+    },
+    facilitiesWithRtLessThan1: {
+      value: `${rtStats.numFacilitiesWithRtLessThan1(
+        facilitiesRtRecords,
+      )} of ${numFacilities}`,
```
jessex, line 82: So just to be sure, if I have 8 facilities and 4 have Rt less than 1, and 2 have Rt greater than 1, and 2 have no Rt value at all yet, then we would say "4 of 8", right? We wouldn't say "4 of 6" or "6 of 8"?

src/design-system/Metric.tsx
```diff
@@ -0,0 +1,51 @@
...
+  label?: string;
+}
+
+const Metric: React.FC<Props> = (props) => {
+  const { value, label } = props;
+
+  return (
+    <MetricContainer>
+      <Value>
+        {typeof value === "number" ? numeral(value).format("0,0") : value}
```
jessex, line 44: Curious: does this `typeof` check work for the function call from `facilitiesWithCases` below? It evaluates the return type and not the function type itself?

jovergaag: The function call returns a string so `typeof value` check works for that metric as well.

Screen Shot 2020-05-15 at 3 06 27 PM

src/page-multi-facility/SystemSummary.tsx
jovergaag: Great question, I need to check on that.

jovergaag:
@jessex Thanks for the review! You're right, it was a bit different than the mocks. I failed to update the metric labels once the new mocks were available yesterday evening. 

I'm looking into that bug. Up to this point I have only had one scenario so it hasn't been an issue but I just duplicated my scenario and confirm the bug so I'll play around with it this afternoon. I need to run an errand now, back to look into this in a bit.

src/page-multi-facility/SystemSummary.tsx
jovergaag: Confirmed it is "6 of 8" We want to include all facilities in this total, even if they don't have an rT value.

jessex: I'm willing to accept that, as long as we are aware that 2 of those 6 are facilities for which we do not yet have enough information to classify.

sychang: Yes! One way to potentially help clarify this to users is updating the copy to say "facilities with a classifiable slowing rate of spread" to demonstrate that the number is representing only the facilities where an r(t) can be calculated _and_ have r(t) < 1.

jessex:
@jovergaag I think maybe this bug predates your work. For example, the R(t) chart in the bottom left of the impact report page will behave similarly (in production, no less). If you log into the app and generate an impact report for your baseline scenario, that chart will sum up to the right number. Go back to scenario details, then select a different scenario. Then switch back to baseline and generate an impact report again. Now the denominator in that chart will include the number of facilities from the other scenario. It seems like as we switch between scenarios, we are appending to some collection of state and not resetting it. So anything that involves that state is going to be impacted, including your work here.

CC @sychang

daschi:
@jovergaag You may have already found this, but since I had also worked on the rt summary stats I was curious and looked into it. The bug is coming from the `rtData` from the `FacilityContext`, it looks like it is collecting the rtData for all facilities across all scenarios that it has been requested for.

src/page-multi-facility/SystemSummary.tsx
jessex: @sychang if we made that copy change, then that would mean we want the calculation to be the "4 of 8" scenario I outlined above, not the "6 of 8", since there would only be 4 that have a calculable R(t) which is less than 1. If we are doing the 6 of 8, then our text should represent that the number is representing facilities where an R(t) is less than 1 _or_ where R(t) has not been calculated yet.

Does that make sense? What do we want for this?

jessex:
Thanks for the investigation, @daschi! Going to copy that over to the issue we filed for this, #444. That's at the top of the backlog as a production-facing bug, for anyone who has bandwidth.

src/page-multi-facility/SystemSummary.tsx
sychang: 4 of 8 in this example!

jovergaag:
@jessex @sychang This is ready for another look. Joshua - I just confirmed that these styles are matching the current mocks.

Screen Shot 2020-05-18 at 2 42 46 PM


I am still seeing one bug. When a user updates the number of cases in a facility row using the modal, that case count change is not being reflected in the summary counts. That is because the case count change does not prompt the dashboard to fetch facilities again.  Fetching the facilities again will cause the facility rows to reload which would be very disruptive. Once the user reloads the page or navigates to another page and back, the case counts are correct. 

I need to go focus on ESH for the rest of the day but I can look into how to trigger that updated case count tomorrow. Let me know if we want to move forward with this as is, or addressing the updating case count issue is a merge blocker.

jessex:
Thanks, @jovergaag, for the refactor and the bug report!

@sychang, I think that, since changes made through the add cases modal from a facility row could impact a number of these metrics, we should wait until Jen gets the chance to ensure that updates impact the summary stats immediately, instead of pushing through up to production today. Does that sound good to you?

sychang:
Yes, good to hold off. Thanks both!

szhu:
@jovergaag I have a question about this feature. Does this summary info UI depend on all the network calls that caluclate Rt? The code seems suggest that it does (SystemSummary takes in `facilitiesRtData`), but when I run it, I see that the summary info shows up before any of the network calls complete.

szhu:
@jovergaag I have a question about this feature. Does this summary info UI depend on all the network calls that caluclate Rt? The code seems suggest that it does (SystemSummary takes in `facilitiesRtData`), but when I run it, I see that the summary info shows up before any of the network calls complete.

jessex:
@jovergaag can chime in to correct me if I am wrong, but _yes_ at least one of the summary stats depends on the R(t) calculations for all of the facility rows being returned (the one about slowing rate of spread, the bottom right stat). It does not show a spinner or any other cue to indicate it's still waiting besides the fact that the rows themselves are still loading, but the numbers do update with each successful R(t) calculation response.

jovergaag:
@jessex Yes, you are correct, one of the summary stats (bottom right) depends on the R(t) calculations. The numbers are updated as the R(t) calculations finish and updated `rtData` is passed into the System Summary.

The loading spinner is only rendered until the cases/population calculations are finished based on data from the `getFacilities()` call in `MultiFacilityImpactDashboard`. At that point the summary statistics are rendered, including the R(t) metric, which initially renders '0 of 0' and continues to update as R(t) calculations finish. This is the same behavior as on the `ResponseImpactDashboard`.

@szhu Does that answer your question?

szhu:
Yes this answers my question! I need to know for my PR whether we in fact need to load all the Rt values, and it looks like we do.

In that case, just a suggestion, do you think it's appropriate to render a loading spinner or a value of "—" in that cell until all the Rt values are loaded? This way users aren't misled by the value of 0. (Or maybe I'm not understanding the use case and we don't need to do that.)

jessex:
I think that may be a nice little UX tweak, @szhu!

@sychang, Sean raised the idea of showing a loading spinner or some other visual cue to indicate that the slowing rate of spread stat (which relies on the asynchronous R(t) calculations) is still loading until all R(t) calculations have returned. If you agree, feel free to file an issue with whatever design you prefer!

PR #457, opened 2020-05-19
[Excalibur] Change default date

amyrlam:
## Description of the change

WAS:
image

IS:
image

---

~~Question: @daschi Serena mentioned something about different behavior the second time a user generates a report, but from looking at the code and clicking around, I'm not sure what she meant?~~ 

## Type of change

## Related issues

Was mentioned at standup then pinged @sychang when was looking for new task!

### Development

### Code review

### DIFF
--- a/src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
+++ b/src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
@@ -89,10 +89,11 @@ const BaselinePopulationForm: React.FC<Props> = ({
   defaultIncarceratedPopulation,
   defaultStaffPopulation,
   saveBaselinePopulations,
+  defaultDate,
   setPage,
 }) => {
   // Default Original Date 12/31/2019
-  const defaultDate = new Date(2019, 11, 31);
+  // const defaultDate = new Date(2019, 11, 31);
 
   const [populations, setPopulations] = useState({
     staffPopulation: defaultStaffPopulation,
--- a/src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
+++ b/src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
@@ -91,8 +91,8 @@ const BaselinePopulationForm: React.FC<Props> = ({
   saveBaselinePopulations,
   setPage,
 }) => {
-  // Default Original Date 2/1/2020
-  const defaultDate = new Date(2020, 1, 1);
+  // Default Original Date 12/31/2019
+  const defaultDate = new Date(2019, 11, 31);
 
   const [populations, setPopulations] = useState({
     staffPopulation: defaultStaffPopulation,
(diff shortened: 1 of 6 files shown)


daschi:
@amyrlam If a user has already visited the impact page then they'll save a `baselinePopulation` (`{ date: Date, staffPopulation: number, incarceratedPopulation: number }`) to the `baselinePopulations` array on the scenario.

Then, in the `useSystemWideData` function called in `ResponseImpactDashboard`, it sets the latest population numbers in the `baselinePopulations` array if there are any (lines 28 and 36), otherwise it sets the population numbers to the public data from the locale spreadsheet that's available from `getLocaleDefaults`. 

At the end of the function on lines 61-69 you'll see that it sets the `systemWideData` internal state to use the user input numbers if available, or the default locale data. Then those numbers are passed to the BaselinePopulationModal to display on line 108 of the ResponseImpactDashboard.

Lemme know if that doesn't help with your question or if there was something else!

amyrlam:
@daschi thanks for your help! I think I understand what you mean - I think you're saying that once I've visited `/impact` and changed the default population inputs, those numbers persist if I visit `/impact` a second time. 

So if I visit `/impact` and change the date when generating a report, you also want the date to persist if I visit `/impact` a second time? I'll take a stab at that!

daschi:
@amyrlam yes, I think that should still be the case with the date too, since the baselinePopulation that is saved includes the date and it's being passed to the `InputDate` on the form.

amyrlam:
Let's have this reviewed / merged as-is, and I'll focus on having the `defaultDate` persist (as `defaultStaffPopulation` and `defaultIncarceratedPopulation`) in a separate PR.

* `defaultDate` should start as `12/31/2019` instead of `2/1/2020` on first visit to /impact, as it does in this PR
* If the user visits `/impact` once and saves a different date e.g. `1/15/2020`, `defaultDate` should be overridden with `1/15/2020`
  * Meaning that, if the user visits `/impact` a second time, the datepicker in the `/impact` model should default to what the user inputted, `1/15/2020`. The date can be changed again

src/page-response-impact/BaselinePopulationModal/BaselinePopulationForm.tsx
```diff
@@ -89,10 +89,11 @@ const BaselinePopulationForm: React.FC<Props> = ({
   defaultIncarceratedPopulation,
   defaultStaffPopulation,
   saveBaselinePopulations,
+  defaultDate,
   setPage,
 }) => {
-  // Default Original Date 2/1/2020
-  const defaultDate = new Date(2020, 1, 1);
+  // Default Original Date 12/31/2019
```
jessex, line 95: We can get rid of both of these commented lines now, right?

src/page-response-impact/hooks/useSystemWideData.tsx
```diff
@@ -58,8 +60,12 @@ export function useSystemWideData(
       modelInputs,
     );
 
+    // Default Original Date 12/31/2019
```
jessex, line 63: I think people will look at these two lines in succession and assume there's an off by one error. Can you comment why this is correct?

amyrlam:
Thanks @daschi for all of your help! Here we are with the date persist, we are good to go:

impact-date

PR #458, opened 2020-05-19
Add authenticated users to Firestore 

### LINKED ISSUE #431: Create a saveUser Function to Store User Information

To support the user-to-user sharing work, we will need a way to store User related information in Firestore.  Create a `saveUser` function to accomplish this goal.

The method signature for this function will likely look something like this:

`saveUser(userAttributes: any): Promise<void>`

The following information should be stored along with each user document:  `auth0Id`, `name`, `email`, `createdAt`, `updatedAt`

### LINKED ISSUE #432: Store the User’s Auth0 Attributes Upon Signup

To support the user-to-user sharing work, we will need a way to store User related information for each new user who signs up for the app.

When a user signs up for the app, pull the `email`, `name`, and `auth0Id` and use the [saveUser function](https://github.com/Recidiviz/covid19-dashboard/issues/431) to store this information in Firestore.  This information is provided by Auth0 and can be accessed from the `user` state of this context: https://github.com/Recidiviz/covid19-dashboard/blob/master/src/auth/react-auth0-spa.tsx

### LINKED ISSUE #433: Update the Name Field for All Existing Users in Auth0

To support the user-to-user sharing work, we will need to know the name of each of the users within our system.

This is a manual process where someone with Auth0 access and knowledge of our user base will need to update each user that is currently tracked in our Auth0 instance.

**Note:**  This work should not be started until https://github.com/Recidiviz/covid19-dashboard/issues/419 is completed so that we don't have gaps in our data if users sign up after this manual data entry process.

### LINKED ISSUE #435: Create a getUser Function to Retrieve User Information

To support the user-to-user sharing work, we will need to be able to retrieve user information from Firestore.

Create a `getUser(auth0Id: string): Promise<User | null>` function to return a User based on an Auth0 id.

This work also implies some additional work that will need to be performed:

* Add a `User` type to our [Type definition file](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/page-multi-facility/types.tsx).
* Add a `buildUser` [type transform](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/type-transforms.tsx) that converts a Firestore user document into a plain JavaScript object of type `User`

macfarlandian:
## Description of the change

Creates fetch and save functions for documents in the new Firestore `users` collection. Upon successful authentication, uses these new functions as needed to create or update the corresponding document to store the name and email provided by Auth0. (This is a slightly different approach than what was outlined in the linked tickets, because there is not, as far as I could tell, any way to detect and respond to a new user registration from the Auth0 web client. On the other hand, this also gets us a bonus feature of keeping our records in sync with Auth0 as people revisit the site.)

I also had to update the Firestore security rules, because it turns out there was a flaw in my testing previously and it wasn't actually possible to create a new user document for oneself. These rules (tested in the Firestore console and locally) should allow for that and also forbid users from changing their Auth0 ids (whether inadvertently or for malicious reasons, whatever those might be). The rules are already live — I needed them in place to test this, and no prod code refers to the collection yet anyway, so it should be safe.

## Type of change

## Related issues

Closes #432, closes #431, closes #435 

### Development

### Code review

### DIFF
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -50,7 +50,8 @@ service cloud.firestore {
       }
 
       allow read: if userIsAuthenticated();
-      allow write: if userIsSelf(resource);
+      allow update, delete: if userIsSelf(resource) && userIsSelf(request.resource);
+      allow create: if userIsSelf(request.resource);
     }
   }
 }
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -50,7 +50,7 @@ service cloud.firestore {
       }
 
       allow read: if userIsAuthenticated();
-      allow update, delete: if userIsSelf(resource) && userIsSelf(request.resource);
+      allow update: if userIsSelf(resource) && userIsSelf(request.resource);
       allow create: if userIsSelf(request.resource);
     }
   }
--- a/src/auth/react-auth0-spa.tsx
+++ b/src/auth/react-auth0-spa.tsx
@@ -77,6 +77,7 @@ export const Auth0Provider = ({
     setUser(user);
   };
 
+  // This hook is specific to our application and not derived from Auth0 example code
   useEffect(() => {
     async function updateStoredUser() {
       if (user !== undefined) {
--- a/src/auth/react-auth0-spa.tsx
+++ b/src/auth/react-auth0-spa.tsx
@@ -1,7 +1,10 @@
 // @ts-nocheck
 
 import createAuth0Client from "@auth0/auth0-spa-js";
+import { isEqualWith, pick } from "lodash";
 import React, { useContext, useEffect, useState } from "react";
+
+import { getUser, saveUser } from "../database";
 
 const DEFAULT_REDIRECT_CALLBACK = () =>
   window.history.replaceState({}, document.title, window.location.pathname);
@@ -73,6 +76,42 @@ export const Auth0Provider = ({
     setIsAuthenticated(true);
     setUser(user);
   };
+
+  useEffect(() => {
+    async function updateStoredUser() {
+      if (user !== undefined) {
+        const savedUser = await getUser(user.sub);
+        if (!savedUser) {
+          // this is expected, as auth0 users predate this feature.
+          // we can lazy-migrate them into our user database by
+          // creating a record now.
+          saveUser({
+            name: user.name,
+            email: user.email,
+            auth0Id: user.sub,
+          });
+        } else {
+          // if anything has changed, update our records.
+          // this anoints Auth0 as the ultimate source of truth
+          // for user info, synced at the beginning of a session
+          if (
+            !isEqualWith(
+              savedUser,
+              user,
+              (a, b) => a.name === b.name && a.email === b.email,
+            )
+          ) {
+            saveUser({
+              ...savedUser,
+              ...pick(user, ["email", "name"]),
+            });
+          }
+        }
+      }
+    }
+    updateStoredUser();
+  }, [user]);
+
   return (
     <Auth0Context.Provider
       value={{
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -382,7 +382,7 @@ export const saveUser = async (userData: UserToSave): Promise<void> => {
       db.collection(usersCollectionId).add(buildCreatePayload(userData));
     }
   } catch (e) {
-    console.error("Encounter error while trying to save user:\n", e);
+    console.error("Encountered error while trying to save user:\n", e);
   }
 };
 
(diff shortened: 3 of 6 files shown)


backend/firestore.rules
```diff
@@ -50,7 +50,8 @@ service cloud.firestore {
       }
 
       allow read: if userIsAuthenticated();
-      allow write: if userIsSelf(resource);
+      allow update, delete: if userIsSelf(resource) && userIsSelf(request.resource);
```
jessex, line 53: Can you briefly explain this change?

src/auth/react-auth0-spa.tsx
```diff
@@ -73,6 +76,42 @@ export const Auth0Provider = ({
     setIsAuthenticated(true);
     setUser(user);
   };
+
+  useEffect(() => {
```
jessex, line 80: Can we add a comment just above this hook explaining that this is a custom addition to the file that was _not_ copied in from Auth0 docs?

src/auth/react-auth0-spa.tsx
```diff
@@ -73,6 +76,42 @@ export const Auth0Provider = ({
...
+            saveUser({
+              ...savedUser,
+              ...pick(user, ["email", "name"]),
+            });
+          }
+        }
+      }
+    }
+    updateStoredUser();
+  }, [user]);
```
jessex, line 113: So, if I understand correctly, the net effect of this is that the hook is only called when the `user` object changes, which should only be when the user logs into a new session, right?

src/auth/react-auth0-spa.tsx
```diff
@@ -73,6 +76,42 @@ export const Auth0Provider = ({
...
+          // we can lazy-migrate them into our user database by
+          // creating a record now.
+          saveUser({
+            name: user.name,
+            email: user.email,
+            auth0Id: user.sub,
+          });
+        } else {
+          // if anything has changed, update our records.
+          // this anoints Auth0 as the ultimate source of truth
```
jessex, line 95: I'm in love with any sincere usage of the word _anoints_ in our codebase. :+1:

src/database/index.tsx
```diff
@@ -328,6 +336,56 @@ export const getFacilityModelVersions = async ({
...
+        throw new Error(
+          `Missing required field "auth0Id" in object ${JSON.stringify(
+            userData,
+          )}`,
+        );
+      }
+      db.collection(usersCollectionId).add(buildCreatePayload(userData));
+    }
+  } catch (e) {
+    console.error("Encounter error while trying to save user:\n", e);
```
jessex, line 385: Typo: "Encountered"

src/database/type-transforms.tsx
```diff
@@ -81,3 +87,9 @@ export const buildScenario = (
 
   return scenario;
 };
+
+export const buildUser = (document: firebase.firestore.DocumentData): User => {
+  const data = document.data();
+
+  return { ...pick(data, ["name", "email"]), id: document.id };
```
jessex, line 94: I might be making an ultra-obvious mistake, but shouldn't we also be picking `"auth0Id"`? Why is that field not on the user type, more generally?

backend/firestore.rules
macfarlandian: yes, this verifies that the authenticated user's ID matches both the `auth0Id` field in the existing document and the `auth0Id` field that would result from the incoming change.

Looking at this now, though, I believe there is still a mistake in here, because delete commands won't have an incoming change. (We're not currently deleting user records so this was another oversight in my testing.) It's possible this whole thing is overkill anyway; the critical thing is the new change rule, so I can restore this to its previous state if you think that's best.

src/auth/react-auth0-spa.tsx
macfarlandian: it'll run every time we authenticate them, which is essentially on every page load (after login or immediately if they are already logged in)

src/database/type-transforms.tsx
macfarlandian: I left it out because a) we don't need it and b) we don't want it to leak. In this situation users are only looking up themselves, but once they start looking up others (i.e. when we implement the actual sharing feature) we don't want them having access to everyone's auth0 credential. 

We do need it to be in the database, which is why it's required for new document creation, but other than that we shouldn't be passing it around.

backend/firestore.rules
jessex: Thanks for the explanation. What if we just remove the `delete` action from here for now and loop back if and when we need explicit user deletion support?

Issue #261, opened 2020-04-30
Fix frontend performance issues


szhu:
Edit: List of issues in priority order, [as discussed below](https://github.com/Recidiviz/covid19-dashboard/issues/261#issuecomment-622157870):

*   [x] Lazy loading the facilities in the facilities page (#460)
*   [x] Debouncing form inputs for a short duration (~0.3s~ 1s) so that people who type decently fast will get a good experience (#474)
*   [x] Committing form inputs only on enter/unfocus (#474)
*   [ ] Stop using react-spinners
*   [x] Debounce chart and table renders -- probably not needed if inputs are debounced (#474)
*   [ ] Memoize expensive-to-render stuff that has siblings that do constantly need to re-render
*   [x] Consider using an alternative to current GlobalStyles approach (#476)

---

**Is your feature request related to a problem? Please describe.**

A few key interactions on the site are slow, including:

*   Loading the list of facilities is very slow.
*   Loading the list of facilities blocks the UI. The user cannot go to the homepage and immediately load click Add Facility.
*   When editing a facility, typing any text in any of the fields blocks the UI for a bit. It should not block the UI at all.

**Describe the solution you'd like / Describe alternatives you've considered**  
I did some proof of concepts and did some React profiling. These are some solutions that could help. (All need futher investigation/validation.)

Both pages:

*   Consider using an alternative to styled-components GlobalStyles or consider using it a different way. The React profiler suggests that GlobalStyles rendering (at least the way we're currently using it, which is to place it on every page rather than at the root) takes a significant amount of time.

Scenario page:

*   Consider using an alternative to the react-spinners. The React profiler suggests that the emotion css framework that it requires is taking a significant amount of time to render.
*   Consider only rendering the facilities graphs are currently visible in the browser window.
*   Consider adding a short delay (1s?) before fetching any data, so that users who want to load the home page and immediately add a facility will not be blocked.

Facility page:

*   Consider debouncing form field inputs (only commit the input after a short delay).
*   Consider aggressively memoizing components that don't change from render to render (otherwise **every** form field re-renders when editing **any** form field).
*   Consider debouncing the chart and table renders.

I will edit this list as we come up more fixes/if we discover some of these fixes don't result in significant improvement. Please use the comment below to suggest other fixes and other ways to determine what's slow!

---

cc @jessex

jessex:
[...]

PR #460, opened 2020-05-20
Speed up performance on multi-facility page

szhu:
## Description of the change

Speeds up the current home page to reduce the time it takes for someone to load the page and immediately add a facility. In total, these changes reduces the time the user has to wait before being able to add a new facility from

- On modern browsers: 5s -> 2.5s
- On Internet Explorer: ~3 minutes -> ~1 minute

**Hoist auth call to beginning and remove duplicate call:**

Loading any user data is bottlenecked on the initial calls to Auth0. (No data even starts being loaded until the Auth0 call completes.) Previously, the Auth0 call is made twice:
- Once by `<Auth0Provider>` in `src/auth/react-auth0-spa.tsx` when the React tree mounts.
- Once by `authenticate()` in `src/database/index.tsx` when the various components start to request data.

This PR makes these two fixes:
- The call is made asap rather than when components start to get mounted.
- The call is now made just once instead of twice.

**Impact:** On modern browsers, reduces the time to displaying the "Add facility" button from 3.5 to 2.5 seconds (1 second faster, 30% faster).

**Lazy load facility rows:**

We only show the content of the facility rows when they are visible in the browser window.

The reason this is important is because even though the loading for facility rows is mostly asynchronous, some parts of the logic (like rendering the chart) appear to be synchronous and block the UI from being interacted for fractions of a second to a few seconds at the time.

NOTE: This original PR had code to also lazy-load the Rt calculations, but this conflicts with #427, so the merge commit reverts this change.

**Impact:** On modern browsers, this reduces the time to being able to interact with the "Add facility" button and navigate to the "Add facility" page by a few seconds.

On Internet Explorer, this provides a much more incredible boost. In Internet Explorer, much more of the code is blocking, and nothing on the page can be interacted with until all the charts load.

**Impact:** On Internet Explorer, testing on a page with 12 facilities, this reduces the time to being able to interact with the "Add facility" button and navigate to the "Add facility" page from 171 seconds to 51 seconds (120 seconds faster, 70% faster).

| Before    | After    |
| ----------| -------- |
| ![before] | ![after] |

[before]: https://user-images.githubusercontent.com/1570168/82450597-beb10980-9a61-11ea-8c40-f4ddb5972ef2.gif
[after]: https://user-images.githubusercontent.com/1570168/82450592-bd7fdc80-9a61-11ea-95d5-b0c06e27bc98.gif

## Type of change

## Related issues

Partially addresses #261 

### Development

### Code review


src/page-multi-facility/FacilityRowPlaceholder.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { useRef, useState } from "react";
+import useVisibility from "react-use-visibility";
+
+import Loading from "../design-system/Loading";
+import { Facility } from "./types";
+
+interface Props {
+  facility: Facility;
```
macfarlandian, line 9: it doesn't seem like you are actually using this prop?

src/auth/AppAuth0ClientPromise.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Creating the Auth0 client was originally done in
+ * src/auth/react-auth0-spa.tsx, but the non-React-aware code in
+ * src/database/index.tsx needs access to this client as well.
+ *
+ * TODO: Refactor the database code to accept an Auth0 client, or something like
+ * that.
+ */
+const AppAuth0ClientPromise: Promise<Auth0Client> =
+  typeof window === "undefined"
+    ? new Promise(() => undefined)
```
jessex, line 33: What are the cases in which we may reach this undefined state, and what is the impact on app behavior?

src/page-multi-facility/FacilityRowPlaceholder.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children: React.ReactElement;
+}
+
+const FacilityRowPlaceholder: React.FC<Props> = (props) => {
```
jessex, line 13: Can we add a comment somewhere in this file that explains its purpose? I think future contributors to the project may not get it immediately.

src/auth/AppAuth0ClientPromise.tsx
szhu: All the times when window is `undefined` are during static render. When the code is loaded by the browser, window is always defined! 

This check is done a few times in the app. Let me know what a good way to document this is. I thought about abstracting it into a function like `isStaticRender`, but I'm not sure if such a function would make it less clear why we need the check. I can also just add a short comment (`// window is undefined during static rendering`) every time we have this check.

src/page-multi-facility/FacilityRowPlaceholder.tsx
szhu: Yes good idea!

src/auth/AppAuth0ClientPromise.tsx
jessex: Yeah I think another layer of indirection on that would probably harm legibility, I agree. Just that short comment would probably suffice, but we may not even need it. I was more curious than anything.

src/page-multi-facility/FacilityRowPlaceholder.tsx
szhu: Oh good point. I originally was, then I reverted the change.

src/auth/AppAuth0ClientPromise.tsx
szhu: Fixed in 6b1bf93

szhu:
Note: After adding in the System Summary from #427, it's even faster in IE now since less rows are initially visible! Now it takes only about 20 seconds to get to the screen to add a new facility, which seems so much more reasonable compared to before.

szhu:
> this is super cool @szhu ... there is one question i have beyond the minor issue i have flagged here. One of my suspicions (never fully investigated) was that the major blocker in the facility rows was not the chart rendering itself, but rather the curve function, which is executed synchronously in the `ChartArea` render function when the components are first mounted. I'm just wondering whether that's something you considered here and whether you think it would make an appreciable difference to their performance?
> 
> (I think rendering lots of charts is also probably not super performant regardless of where their data comes from, so I'm not sure which has the greater influence in more extreme cases and for IE users)

@macfarlandian and I discussed this a little more on Slack. I think it's out of scope for this PR, but I'll make a note of this in #261 so we can track it there.

PR #466, opened 2020-05-21
use opacity on datepicker show/hide

### LINKED ISSUE #259: Date picker in model input is broken
**Describe the bug**
The date picker for Planned Releases fails to populate the date field. (The field can still be updated by typing in a date manually.)

**To Reproduce**

1. Go to the facility page for any facility
2. Click on an empty date field for planned reductions under "Mitigation effors"
3. Click on a date in the calendar popover
4. See error: date field remains blank
5. (alternatively: do the same for an already populated date field and observe that the field value does not change)

**Expected behavior**
Date field should be populated with the selected date

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

- OS [e.g. Windows 10]: macOS Catalina
- Browser [e.g. chrome, edge]: Firefox
- Version [e.g. 22]: latest

**Smartphone (please complete the following information):**

- Device: [e.g. iPhone6]:
- OS: [e.g. iOS8.1]:
- Browser [e.g. stock browser, safari]:
- Version [e.g. 22]:

**Additional context**
Bug is seen on latest `master` development builds but does not currently occur in prod

glen0071:
## Description of the change

DatePicker selection not saving on Firefox.

DatePicker component with `.react-date-picker__calendar--closed` class was set to `display: none`. In Firefox, maybe element is removed from DOM before onChange method triggers? 

Using opacity to conceal/show DatePicker seems to work on Firefox (Chrome worked either way - even without InputDate.css rules on `closed`/`open`, I guess Chrome works well with `react-date-picker`'s out-of-the-box styling).

In giph, Firefox on left, Chrome on right
fix-datepicker
 
## Type of change

## Related issues

Closes #259

### Development

### Code review

### DIFF
--- a/src/design-system/InputDate.css
+++ b/src/design-system/InputDate.css
@@ -211,7 +211,10 @@
   z-index: 1;
 }
 .react-date-picker__calendar--closed {
-  display: none;
+  opacity: 0;
+}
+.react-date-picker__calendar--open {
+  opacity: 1;
 }
 .react-date-picker__calendar .react-calendar {
   border-width: thin;


src/design-system/InputDate.css
```diff
@@ -211,7 +211,10 @@
   z-index: 1;
 }
 .react-date-picker__calendar--closed {
-  display: none;
+  opacity: 0;
```
macfarlandian, line 214: this is mysterious to me since the date picker still works in Firefox in other places (e.g. the add cases modal), but if it does fix the issue I don't necessarily have an issue with it. However, one note - because this CSS file is copied out of the package to work around a bug (see comment at the top) I would rather we didn't make the change here, but rather did an override in the styled component inside `DatePicker.tsx` instead.

We might not even need this file anymore because it was an issue specific to Parcel, so I wouldn't want us to cause a regression when we get around to removing it.

glen0071:
Ok. I'm going to dig into this some more and find a better fix for it. And implement it in the DatePicker's styled components.

glen0071:
@jessex @macfarlandian 
Seems like react-date-picker does not work with GatsbyJS. After ruling out a lot of other possibilities, I spun up a starter Gatsby app and plopped react-date-picker into the middle it. Same issue in Firefox.

I think it works on the AddCasesModal because the Modal attaches to the DOM independent from the rest of the app.

Not sure if we can configure Gatsby to fix this. Should I adapt the InputDate component (which AddCasesModal also relies on) to attach to the DOM?

macfarlandian:
@glen0071 I'm not sure I totally follow the logic ... both the modal and the form page are React components and therefore part of the Gatsby app. I'm not seeing the causal link here, maybe you can explain further?

glen0071:
@macfarlandian Good point. I was probably over confident in that last comment. FWIW, I ran a number of tests, adding/removing different components in different combinations. The only thing that seemed to have any affect in Firefox is when the reac-date-picker was in a component mounted to the DOM via 
```
return ReactDOM.createPortal(
    <BackgroundAside onClick={handleOnClick}>
      <ModalContainer ref={ref} height={height} width={width}>
        <ModalTitle title={title} closeModal={closeModal} />
        {children}
      </ModalContainer>
    </BackgroundAside>,
    document.getElementById("___gatsby") as HTMLElement,
  );
```

So, I have to admin I don't know why that is or what that means. I can continue thinking/digging down that path. It made me feel like the easiest fix might be to attach the react-date-picker's parent component to some DOM id. But I'd love to hear other ideas, or else I'll dig around a bit later tonight.

macfarlandian:
that's interesting ... I guess I would feel better if we could identify a root cause ... that still just feels like a mysterious and possibly brittle workaround to me

glen0071:
@macfarlandian I've done some more digging, but not much progress.

ReactDOM docs sorta clarify why that can fix it:
```
The react-dom package provides DOM-specific methods that can be used at the top level of your app and as an escape hatch to get outside of the React model if you need to. Most of your components should not need to use this module.
```

We know it doesn't work, and it doesn't throw any errors in Firefox/Safari. And since the same problem arises in a brand new Gatsby project, but not a create-react-app project, I suspect it's something unique to react-date-picker + Gatsby + Firefox/Safari.

So I guess the next step is to dig into react-date-picker's implementation for clues? Or focus on Gatsby docs? (I fished around there a bit, but I don't think I saw anything relevant, so I thought react-date-picker source code might be a better route.)

macfarlandian:
@glen0071 yeah probably react-date-picker ... if you can identify a bug in there they might just fix it, I believe it is actively maintained

glen0071:
@macfarlandian I opened an issue on react-date-picker, hoping to get some input from its author on how to fix this. Also wondering if it's best to close this PR since the fix will probably come from that library?

macfarlandian:
@glen0071 yeah I think it's fine to close this ... might be good to note the related issue on react-date-picker in our ticket so we can follow up later

glen0071:
Opened an issue on the react-date-picker package, https://github.com/wojtekmaj/react-date-picker/issues/274

Issue #440, opened 2020-05-15
Update the getScenarios Function to Consider the viewer Role Type


sprad:
**Is your feature request related to a problem? Please describe.**

To support the user-to-user sharing work, we will need to return all Scenarios a user has access to regardless of whether or not they are the Scenario’s `owner` or if they are just a `viewer`.

**Describe the solution you'd like**

I’m hoping this is as simple as adding `viewer` to the [list of roles](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/index.tsx#L197) in the `where` clause of the `getScenario` query.

Issue #438, opened 2020-05-15
Create a removeScenarioUser Function to Remove a User from a Scenario


sprad:
**Is your feature request related to a problem? Please describe.**

To support the user-to-user sharing work, we will need to be able to remove users from a Scenario.

**Describe the solution you'd like**

Create a `removeScenarioUser(scenarioId: string, auth0Id: string): Promise<void>` function to remove the given `auth0Id` from the `roles` map of the Scenario with the given `scenarioId`.

Issue #437, opened 2020-05-15
Create an addScenarioUser Function to Add a User to a Scenario


sprad:
**Is your feature request related to a problem? Please describe.**

To support the user-to-user sharing work, we will need to be able to add users to a Scenario.

**Describe the solution you'd like**

Create an `addScenarioUser(scenarioId: string, email: string): Promise<User>` function to save the given `auth0Id` (as looked up by the given `email`) into the `roles` map of the Scenario with the given `scenarioId`.  For now, this user’s role can be hardcoded to `viewer`.

Issue #436, opened 2020-05-15
Create a getScenarioUsers Function to Retrieve a List of Users who can Access a Scenario


sprad:
**Is your feature request related to a problem? Please describe.**

To support the user-to-user sharing work, we will need to be able to retrieve a list of users associated with a given Scenario from Firestore.

**Describe the solution you'd like**

Create a `getScenarioUsers(scenarioId: string): Promise<Array<User> | null>` function to return an array of users belonging to the given `scenarioId`.

This will require gathering all of the Auth0 IDs from the Scenario’s `roles` map and then using the [getUser function](https://github.com/Recidiviz/covid19-dashboard/issues/435) to “convert” those IDs into `User` objects.

PR #469, opened 2020-05-21
Scenario Sharing DB Functionality

### LINKED ISSUE #461: Investigate Security Rule Updates to Limit Potential Data Leakage from the User Collection

At the moment, our firestore.rules allow anyone who is authenticated to `read` from the `users` collection.  This technically allows any authenticated user with the right know-how (i.e. the ability to open a dev tools console and to write some simple code) to access a list of every user in our system along with their attributes (email, Auth0 id, etc.)

Investigate whether or not switching the `read` rule on the `users` collection to a `get` rule would still allow our application to function as desired.  This would eliminate the `list` functionality implied within `read` (which is shorthand for `get` and `list`).  Without `list` in place, users would not be able to simply read the contents of the entire `users` collection.

There are two pieces of functionality we will want to test after making this change:

1.  Invitation functionality.  This should only work if the user knows the email address associated with a `users` document.

2.  Displaying the users in the shared scenario.  For the listing of users in the scenario, it’s TBD whether or not the `where` clause in the firestore query is smart enough to apply the security rules individually to each of the Auth0 users ids provided in the roles map of the Scenario.  If not, we may have to make a query for each user document individually (which is a tradeoff of performance for security).

**A clear and concise description of what you want to happen.**

Conduct the investigation above.  If the results allow us to achieve the desired functionality while also reducing our data exposure, then implement the rules in production.  Otherwise, investigate alternate approaches.

sprad:
## Description of the change

This work includes:

* Investigating Security Rule Updates to Limit Potential Data Leakage from the User Collection
* Creating an addScenarioUser Function to Add a User to a Scenario
* Creating a getScenarioUsers Function to Retrieve a List of Users who can Access a Scenario
* Creating a removeScenarioUser Function to Remove a User from a Scenario
* Updating the getScenarios Function to Consider the viewer Role Type

## Type of change

## Related issues

Closes #461 #437 #436 #438 #440

### Development

### Code review

### DIFF
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -50,7 +50,7 @@ service cloud.firestore {
       }
 
       // These rules limit read access to the users collection by enforcing that read queries
-      // must proivide either an email address or Auth0 User ID.  This prevents the potential
+      // must provide either an email address or Auth0 User ID.  This prevents the potential
       // of a malicious client-side user from accessing every single user document at once.
       // This works because Firestore query constraints must match the same constraints
       // specified by the Firestore rules:
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -49,8 +49,16 @@ service cloud.firestore {
         return user.data.auth0Id == request.auth.uid;
       }
 
-      allow read: if userIsAuthenticated();
-      allow update: if userIsSelf(resource) && userIsSelf(request.resource);
+      // These rules limit read access to the users collection by enforcing that read queries
+      // must proivide either an email address or Auth0 User ID.  This prevents the potential
+      // of a malicious client-side user from accessing every single user document at once.
+      // This works because Firestore query constraints must match the same constraints
+      // specified by the Firestore rules:
+      // https://firebase.google.com/docs/firestore/security/rules-query#secure_and_query_documents_based_on_a_field
+      allow read: if userIsAuthenticated() && resource.data.email == resource.data.email;
+      allow read: if userIsAuthenticated() && resource.data.auth0Id == resource.data.auth0Id;
+
+      allow update, delete: if userIsSelf(resource) && userIsSelf(request.resource);
       allow create: if userIsSelf(request.resource);
     }
   }
--- a/backend/firestore.rules
+++ b/backend/firestore.rules
@@ -58,7 +58,7 @@ service cloud.firestore {
       allow read: if userIsAuthenticated() && resource.data.email == resource.data.email;
       allow read: if userIsAuthenticated() && resource.data.auth0Id == resource.data.auth0Id;
 
-      allow update, delete: if userIsSelf(resource) && userIsSelf(request.resource);
+      allow update: if userIsSelf(resource) && userIsSelf(request.resource);
       allow create: if userIsSelf(request.resource);
     }
   }
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -329,17 +329,12 @@ export const getFacilityModelVersions = async ({
   }
 };
 
-const getUserDocument = async (params: {
-  email?: string;
-  auth0Id?: string;
-}): Promise<firebase.firestore.DocumentData | never> => {
+const getUserDocument = async (
+  params: { email: string } | { auth0Id: string },
+): Promise<firebase.firestore.DocumentData | never> => {
   const db = await getDb();
 
   const [key, value] = Object.entries(params)[0];
-
-  if (!key || !value) {
-    throw new Error(`Invalid user access attempt with input: ${key}:${value}`);
-  }
 
   const userResult = await db
     .collection(usersCollectionId)
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -329,19 +329,36 @@ export const getFacilityModelVersions = async ({
   }
 };
 
+const getUserDocument = async (params: {
+  email?: string;
+  auth0Id?: string;
+}): Promise<firebase.firestore.DocumentData | never> => {
+  const db = await getDb();
+
+  const [key, value] = Object.entries(params)[0];
+
+  if (!key || !value) {
+    throw new Error(`Invalid user access attempt with input: ${key}:${value}`);
+  }
+
+  const userResult = await db
+    .collection(usersCollectionId)
+    .where(key, "==", value)
+    .get();
+
+  if (userResult.docs.length > 1) {
+    throw new Error(`Multiple users found matching ${key} ${value}`);
+  } else if (userResult.docs.length === 0) {
+    throw new Error(`No users found matching ${key} ${value}`);
+  }
+
+  return userResult.docs[0];
+};
+
 export const getUser = async (auth0Id: string): Promise<User | null> => {
   try {
-    const db = await getDb();
-    const userResult = await db
-      .collection(usersCollectionId)
-      .where("auth0Id", "==", auth0Id)
-      .get();
-    if (userResult.docs.length > 1) {
-      throw new Error(`Multiple users found matching id ${auth0Id}`);
-    } else if (userResult.docs.length === 0) {
-      throw new Error(`No users found matching id ${auth0Id}`);
-    }
-    return buildUser(userResult.docs[0]);
+    const userDocument = await getUserDocument({ auth0Id });
+    return buildUser(userDocument);
   } catch (e) {
     console.error("Encountered error while attempting to retrieve user: \n", e);
     return null;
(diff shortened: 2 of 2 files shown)


backend/firestore.rules
```diff
@@ -49,8 +49,16 @@ service cloud.firestore {
...
+      // These rules limit read access to the users collection by enforcing that read queries
+      // must proivide either an email address or Auth0 User ID.  This prevents the potential
+      // of a malicious client-side user from accessing every single user document at once.
+      // This works because Firestore query constraints must match the same constraints
+      // specified by the Firestore rules:
+      // https://firebase.google.com/docs/firestore/security/rules-query#secure_and_query_documents_based_on_a_field
+      allow read: if userIsAuthenticated() && resource.data.email == resource.data.email;
+      allow read: if userIsAuthenticated() && resource.data.auth0Id == resource.data.auth0Id;
+
+      allow update, delete: if userIsSelf(resource) && userIsSelf(request.resource);
```
sprad, line 61: @macfarlandian - I noticed the extra `delete` on this rule in the Firestore console, but not in our current firestore.rules file.  I'm assuming you added this manually?  Let me know if we don't actually want this included.

macfarlandian: oh yeah that rule is invalid, i think I just didn't redeploy after figuring that out. it should just be `allow update: ...`

backend/firestore.rules
```diff
@@ -49,8 +49,16 @@ service cloud.firestore {
...
-      allow read: if userIsAuthenticated();
-      allow update: if userIsSelf(resource) && userIsSelf(request.resource);
+      // These rules limit read access to the users collection by enforcing that read queries
+      // must proivide either an email address or Auth0 User ID.  This prevents the potential
+      // of a malicious client-side user from accessing every single user document at once.
+      // This works because Firestore query constraints must match the same constraints
+      // specified by the Firestore rules:
+      // https://firebase.google.com/docs/firestore/security/rules-query#secure_and_query_documents_based_on_a_field
+      allow read: if userIsAuthenticated() && resource.data.email == resource.data.email;
+      allow read: if userIsAuthenticated() && resource.data.auth0Id == resource.data.auth0Id;
```
sprad, line 59: I find that these rules read pretty oddly, but they do prevent this query from running by throwing a permissions error: `db.collection("users").get()`

backend/firestore.rules
sprad: Ok, I'll get rid of the `delete` 👍

backend/firestore.rules
```diff
@@ -49,7 +49,15 @@ service cloud.firestore {
...
-      allow read: if userIsAuthenticated();
+      // These rules limit read access to the users collection by enforcing that read queries
+      // must proivide either an email address or Auth0 User ID.  This prevents the potential
+      // of a malicious client-side user from accessing every single user document at once.
+      // This works because Firestore query constraints must match the same constraints
+      // specified by the Firestore rules:
+      // https://firebase.google.com/docs/firestore/security/rules-query#secure_and_query_documents_based_on_a_field
+      allow read: if userIsAuthenticated() && resource.data.email == resource.data.email;
+      allow read: if userIsAuthenticated() && resource.data.auth0Id == resource.data.auth0Id;
+
```
sprad, line 60: Also, these rules have not yet been pushed to production.  I tested them doing that hacky thing I do by creating my own `*_test` collections and only applying the rules to those separate collections.  I'll deploy the rules in a little bit (i.e. ~6pm EDT) once other development work slows down.  I don't want to confuse other in progress work if something is wrong with these rule updates.

src/database/index.tsx
```diff
@@ -348,6 +348,46 @@ export const getUser = async (auth0Id: string): Promise<User | null> => {
...
+      // https://firebase.google.com/docs/firestore/query-data/queries#in_and_array-contains-any
+      const user = await getUser(auth0Id);
+
+      if (!user) {
+        throw new Error(
+          "Error retrieving role for ${auth0Id} in scenario {scenarioId}",
+        );
+      }
+
+      users.push(user);
```
macfarlandian, line 378: won't this issue the `getUser` queries serially? I think you could do something like `const users = await Promise.all(auth0Ids.map((auth0Id) => getUser(auth0Id))` if you wanted to fetch them concurrently

src/database/index.tsx
```diff
@@ -379,6 +419,89 @@ export const saveUser = async (userData: UserToSave): Promise<void> => {
...
+      .where("email", "==", email)
+      .get();
+
+    if (userResult.docs.length > 1) {
+      throw new Error(`Multiple users found matching email ${email}`);
+    } else if (userResult.docs.length === 0) {
+      throw new Error(`No users found matching email ${email}`);
+    }
+
+    const userDocument = userResult.docs[0];
```
macfarlandian, line 440: since this is basically the same as what `getUser` is doing, I might just factor it out of `getUser` into an unexported function and just leave `getUser` to pass the result through `buildUser` (I am assuming that was the motivation here since you need the raw Firestore document)

backend/firestore.rules
macfarlandian: this is absolutely breaking my brain but I can't think of with anything better that wouldn't require a bunch of backend work. magical incantation

backend/firestore.rules
```diff
@@ -49,7 +49,15 @@ service cloud.firestore {
         return user.data.auth0Id == request.auth.uid;
       }
 
-      allow read: if userIsAuthenticated();
+      // These rules limit read access to the users collection by enforcing that read queries
+      // must proivide either an email address or Auth0 User ID.  This prevents the potential
```
jessex, line 53: Typo: "provide"

backend/firestore.rules
jessex: Same. 🤷

src/database/index.tsx
```diff
@@ -348,6 +348,46 @@ export const getUser = async (auth0Id: string): Promise<User | null> => {
...
+
+    if (!scenario) {
+      throw new Error(`No scenario found matching id ${scenarioId}`);
+    }
+
+    const auth0Ids = Object.keys(scenario.roles);
+
+    const users = [];
+    for (const auth0Id of auth0Ids) {
+      // We have to get the users one at a time because Firestore limits the number
```
jessex, line 365: Lame, but okay!

src/database/index.tsx
sprad: Oh, super dope.  I didn't even know this was a thing.  I tried it out locally, and after wrangling in TypeScript, everything still functions.  Thanks for helping me level up! 💪

backend/firestore.rules
sprad: giphy

src/database/index.tsx
sprad: I agree that there's likely an opportunity to refactor some things and remove duplication in this file, but I'm not sure if I'm understanding what's being requested here.  The query and checks here are very similar to `getUser`, but we're looking up the user by `email` here since that's what will be passed in from the frontend (vs. `auth0Id` in `getUser`).  Is the suggestion to create an unexported function that is smart enough to deal with either having an `email` or `auth0Id` passed in as a parameter?

macfarlandian: oh whoops, I didn't even catch that. it's a little more involved than what I imagined when I originally commented but I would still say ... yes. (it could accept, e.g., an object with optional keys `email` and `auth0Id` and require that at least one of them has a value.) because the lookup logic is so finicky I think there is enough value in not having it duplicated that i still wouldn't punt on it

sprad: Ok, cool.  I'll knock it out!

sprad:
@macfarlandian - All feedback has been addressed and this PR is ready for final review

src/database/index.tsx
```diff
@@ -329,25 +329,74 @@ export const getFacilityModelVersions = async ({
   }
 };
 
+const getUserDocument = async (params: {
+  email?: string;
+  auth0Id?: string;
+}): Promise<firebase.firestore.DocumentData | never> => {
```
macfarlandian, line 335: slightly pedantic (and possibly incorrect) but I think if you type this as `{email: string} | {auth0Id: string}` you will enforce only one of these being set at a time; could avoid confusion since we are just arbitrarily picking whichever key JavaScript decides should be first. You might not even need to throw an error in that case because the static analysis would cover it? (fine with me if you just ignore this though since I'm not 100% sure it will work anyway)

src/database/index.tsx
```diff
@@ -379,6 +428,77 @@ export const saveUser = async (userData: UserToSave): Promise<void> => {
   }
 };
 
+export const addScenarioUser = async (
+  scenarioId: string,
+  email: string,
+): Promise<User | null> => {
+  const db = await getDb();
```
macfarlandian, line 435: looks like you don't need this anymore

src/database/index.tsx
sprad: Seems to have worked 👍

PR #474, opened 2020-05-21
Debounce inputs to make text easier to enter

szhu:
## Description of the change

Using the "react-debounce-input" package, we now debounce inputs with a debounce time of 1s.

Specifically, using this package, inputs don't pass up their new values unless:
- 1s has passed without any user interaction, or
- the focus has left the input (the user has clicked or tabbed away), or
- the user presses the Enter key.

(For future reference, the last two triggers can be easily turned off if need be. For example, if we use this package with textareas in the future, we probably want to disable having Enter key immediately register the input. (See readme for more info about this https://www.npmjs.com/package/react-debounce-input)

**Impact:** On the facilities page, you can now input an entire number instead of typing it digit by digit, waiting for the graph on the right to update before you can enter the next number.
- On modern browsers, there's only a few fractions of a second saved, but the interface feels much less laggy.
- On IE, each new text input took ~2 seconds to update. This reduces the time it takes to enter a 4-digit number from 8s (often longer, because sometimes you're not sure how long to wait before pressing the next key) to 2s!

(This is hard to demonstrate with a GIF because GIFs themselves are kind of laggy. See for yourself with the Now preview link!)

Note: This PR is fairly straightforward, but it took a few tries to figure out the best approach! Previously I experimented with writing debouncing code manually, or by using other libraries.

## Type of change

## Related issues

Addresses part of #261 

### Development

### Code review

### DIFF
--- a/src/design-system/InputDate.tsx
+++ b/src/design-system/InputDate.tsx
@@ -11,7 +11,7 @@ import DatePicker from "react-date-picker/dist/entry.nostyle";
 import styled from "styled-components";
 
 import Colors from "./Colors";
-import { InputBaseProps, useInputValue } from "./Input";
+import { InputBaseProps } from "./Input";
 import InputLabelAndHelp from "./InputLabelAndHelp";
 
 const InputContainer = styled.div`
@@ -84,7 +84,6 @@ type Props = InputBaseProps<Date> & {
 
 const InputDate: React.FC<Props> = (props) => {
   const { labelAbove, labelHelp, onValueChange } = props;
-  const inputValue = useInputValue(props);
 
   return (
     <InputContainer>
@@ -109,7 +108,7 @@ const InputDate: React.FC<Props> = (props) => {
         }
         tileClassName={props.tileClassName}
         tileContent={props.tileContent}
-        value={inputValue}
+        value={props.valueEntered}
         yearPlaceholder="yyyy"
       />
     </InputContainer>
--- a/src/design-system/InputText.tsx
+++ b/src/design-system/InputText.tsx
@@ -55,7 +55,9 @@ interface Props extends InputBaseProps<string> {
   onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void;
   onKeyDown?: (event: React.KeyboardEvent) => void;
   focus?: boolean;
+  maxLength?: number;
   placeholder?: string;
+  required?: boolean;
   style?: object;
 }
 
@@ -79,6 +81,8 @@ const InputText: React.FC<Props> = (props) => {
           value={props.valueEntered ?? ""}
           headerStyle={!!props.headerStyle}
           placeholder={props.placeholder ?? placeholder}
+          maxLength={props.maxLength}
+          required={props.required}
           onChange={(e) => props.onValueChange(e.target.value)}
           onBlur={props.onBlur}
           onKeyDown={props.onKeyDown}
--- a/src/design-system/InputText.tsx
+++ b/src/design-system/InputText.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useRef } from "react";
 import styled from "styled-components";
 
 import Colors from "./Colors";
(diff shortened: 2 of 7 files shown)


src/design-system/InputText.tsx
```diff
@@ -68,18 +72,17 @@ const InputText: React.FC<Props> = (props) => {
...
         <WrappedInput
           type={props.type}
-          ref={nameInput}
-          value={inputValue ?? ""}
+          inputRef={nameInput}
+          value={props.valueEntered ?? ""}
           headerStyle={!!props.headerStyle}
           placeholder={props.placeholder ?? placeholder}
-          maxLength={props.maxLength}
-          required={props.required}
```
macfarlandian, line 79: why have we lost these?

szhu: Oh oops, forgot to add a self-comment here! The library doesn't appear to support them. But also, I checked and it doesn't look like we use them either.

macfarlandian: that seems ....... bad? questionable at best? these are pretty standard HTML input attributes, are they not?

szhu: Okay, I did some investigating and yeah you're right, that made absolutely no sense, there is no error. Adding these back!

I'm not sure how this happened, but my best guess as to why I thought these needed to be removed was that when I was in the process of figuring out the correct props interface for CustomDebounceInput, these props weren't working because I maybe used the type for a different, non- `<input>` HTML element.

szhu: To view the updated diff of just the last two commits, you can go [here](https://github.com/Recidiviz/covid19-dashboard/pull/474/files/7a864ed0a17ba06b350de970468746284fd0cfda..4d1f13c0225ba45d952a726d61de614716830eeb).

src/design-system/InputText.tsx
```diff
@@ -17,15 +17,17 @@ const TextInputContainer = styled.div`
 `;
 
 const InputWrapper = styled.div`
-  ${StyledInput}
```
szhu: This is where my previous misunderstanding was that caused the bug.

I thought that including this would bring in the styles from `StyledInput`, but all it did was bring in the auto-generated class name!

The correct way to do this is to use `styled.css` to create a stylesheet (in the fixed version, named `InputStyle`). This stylesheet can then be brought in this way.

szhu: This is where my previous misunderstanding was that caused the bug.

I thought that including this would bring in the styles from `StyledInput`, but all it did was bring in the auto-generated class name!

The correct way to do this is to use `styled.css` to create a stylesheet (in the fixed version, named `InputStyle`). This stylesheet can then be brought in this way.

szhu: Oops, looks like I double-clicked submit on this comment. Marking this duplicate as resolved.

src/design-system/InputDate.tsx
```diff
@@ -109,7 +108,7 @@ const InputDate: React.FC<Props> = (props) => {
         }
         tileClassName={props.tileClassName}
         tileContent={props.tileContent}
-        value={inputValue}
+        value={props.valueEntered}
```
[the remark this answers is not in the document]
szhu: No, the new component does not replace `defaultValue`. I have just removed the defaultValue functionality since it wasn't being used.

Would we need `defaultValue` in the future? I'm pushing for never using `defaultValue` in the future, since `defaultValue` has some behaviors that are easy to overlook. For example, what happens when `value` is also defined? What if `value` is the empty string?

I think it makes more sense for people who want `defaultValue` behavior in the future to build it themselves using `useState` and `useEffect` so that the complexity of the behavior is apparent, and so it is easier to make changes to the default behavior. But anyway, I don't think anyone will want to use `defaultValue`, since its behavior is very nuanced.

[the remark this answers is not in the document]
jessex: Thanks for the explanation, @szhu!

PR #477, opened 2020-05-22
Add a Read-Only Banner When a User with a viewer Role is Visiting a S…

### LINKED ISSUE #442: Add a Read-Only Banner When a User with a viewer Role is visiting a Scenario

To support the user-to-user sharing work, we will need to inform users with read-only access to a given Scenario that they can not modify that Scenario.

Using the visuals provided by the [mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=2485%3A0), implement a banner experience to let users know that they have read-only access to a Scenario when applicable.

Using a combination of the [ScenarioContext](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/scenario-context/ScenarioContext.tsx) and the [Auth0Context](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/auth/react-auth0-spa.tsx) should provide us with the information we need to decide whether or not to display the banner.

sprad:
…cenario

## Description of the change

Shows a read-only banner when a user is viewing a scenario that they do not own.

## Type of change

## Related issues

Closes #442 

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -725,6 +725,10 @@ export const duplicateScenario = async (
     }
 
     await batch.commit();
+
+    const duplicatedScenario = await scenarioDoc.get();
+
+    return buildScenario(duplicatedScenario);
   } catch (error) {
     console.error(
       `Encountered error while attempting to duplicate scenario: ${scenarioId}`,
--- a/src/page-multi-facility/FacilityPage.tsx
+++ b/src/page-multi-facility/FacilityPage.tsx
@@ -8,6 +8,7 @@ import useScenario from "../scenario-context/useScenario";
 import SiteHeader from "../site-header/SiteHeader";
 import { FacilityContext } from "./FacilityContext";
 import FacilityInputForm from "./FacilityInputForm";
+import ReadOnlyScenarioBanner from "./ReadOnlyScenarioBanner";
 
 const FacilityPageDiv = styled.div``;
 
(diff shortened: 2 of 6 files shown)


src/database/index.tsx
```diff
@@ -725,6 +725,10 @@ export const duplicateScenario = async (
     }
 
     await batch.commit();
+
+    const duplicatedScenario = await scenarioDoc.get();
+
+    return buildScenario(duplicatedScenario);
```
sprad, line 731: Turns out we weren't ever actually returning this 🤦

src/page-multi-facility/FacilityPage.tsx
```diff
@@ -27,6 +28,12 @@ const FacilityPage: React.FC = () => {
           localeDataSource={localeDataSource}
         >
           <FacilityPageDiv>
+            {scenario.data && (
+              <ReadOnlyScenarioBanner
+                scenario={scenario.data}
+                dispatchScenarioUpdate={dispatchScenarioUpdate}
+              />
```
sprad, line 35: These properties were available here so I passed them into this new component, but I could have just as easily used `useScenario` within the component itself.  It's still not entirely clear to me when to favor passing properties into a component vs. accessing them via a more global context.

sprad:
**When Viewing a Scenario Not Owned by the User**

Screen Shot 2020-05-22 at 7 09 33 PM

**During Scenario Duplication**

Screen Shot 2020-05-22 at 7 09 45 PM

**After Scenario Duplication**

Screen Shot 2020-05-22 at 7 10 28 PM

src/page-multi-facility/FacilityPage.tsx
jessex: I also don't have a great and handy heuristic, but I think this feels like an appropriate use of props. I _think_ in general we want to favor property injection versus the use of hooks for elements that are static, like this banner.

PR #478, opened 2020-05-23
Catch permissions errors and toast them

### LINKED ISSUE #429: Display a Toast Message Upon Firestore Permission Errors

When a user attempts to read from or write to a Scenario that the do not have permission to access the app will throw the following error:

`Uncaught (in promise) FirebaseError: Missing or insufficient permissions.`

When this happens, we should display a Toast message that informs the user that they do not have permission to perform this action.

I haven’t tried this out yet, but I’m hopeful we can make use of React’s [Error Boundary](https://reactjs.org/docs/error-boundaries.html) concept to globally catch errors of this type and subsequently display the Toast message.  It does seem that by default async errors aren’t caught by Error Boundaries, but [there may be a way to work around this limitation](https://medium.com/trabe/catching-asynchronous-errors-in-react-using-error-boundaries-5e8a5fd7b971).  If this work-around feels too hacky, we’ll need to find another way to capture and alert on these errors.

To further clarify, there are two kinds of Toast Messages / Error Handling Situations:

1. If you somehow got access to a scenario for which you have no permissions whatsoever, then the actual read operation should fail with the toast (and the page should ideally fail to load).  In this case, you ideally be re-routed to your Baseline scenario before seeing the Toast.
2. If you tried to perform a write on a scenario for which you are only a viewer, then you should get a toast telling you that that's not permitted

macfarlandian:
## Description of the change

Implements two limited forms of error handling on database fetches:
- illegal writes, which trigger a toast message but otherwise leave the UI untouched
- illegal reads of scenario or facility data, which trigger a toast and return the user to their baseline scenario

To accomplish this I am detecting and selectively re-throwing the errors we want to surface in the UI. We then have to catch them in different ways depending on the desired behavior. To just display a toast, a simple try-catch block is sufficient. I implemented this in a hook to reduce boilerplate. 

I suppose I could have done something similar to catch and redirect on more severe failures, but for better or worse I did not. Instead I created an ErrorBoundary that can be used at the page level and catches whatever bubbles up. In order for these async errors to bubble up the React tree, they have to be rethrown inside the render loop; I also added a hook for doing that. I do think this is sensible because it lets us render some intermediate state while we prepare the baseline scenario data for redirection. Lacking any mocks, I just used a loading spinner for now. 

The unfortunate thing about this part of the feature is that we have a tendency to throw more than one permissions error at a time, because several parts of the page may be doing concurrent fetches and we do not clean them up when the components are unmounted. This can result in a bunch of noisy React errors, which don't break the app itself but do take over your screen in dev mode, which is annoying. (Prod builds are unaffected by this other than some console logs.) Not great but hopefully good enough for now? (We can fix this by cleaning up our side effects but that seemed out of scope for this PR, as it happens all throughout the app.)

The relative straightforwardness of this code belies what an absolute nightmare it was for me to figure out! Async React error handling is Not Fun.

This is also pretty challenging to test at the moment, without the full sharing feature built out. I had to manually edit my scenario roles in Firestore to simulate the use cases we are targeting here. I will not further bloat this description with a detailed accounting of how to do that but contact me if you want some pointers to guide your own QA efforts.

## Type of change

## Related issues

Closes #429 

### Development

### Code review


src/hooks/useRejectionToast.tsx
```diff
@@ -0,0 +1,20 @@
+import { useCallback } from "react";
+
+import { useToasts } from "../design-system/Toast";
+
+const useRejectionToast = () => {
```
jessex, line 5: Can we add a brief comment explaining how this hook works?

src/page-multi-facility/MultiFacilityErrorBoundary.tsx
```diff
@@ -0,0 +1,48 @@
...
+    () => {
+      getBaselineScenario().then((baselineScenario) => {
+        addToast(
+          "Something went wrong. You will be returned to your baseline scenario.",
+        );
+        if (baselineScenario) {
+          dispatchScenarioUpdate(baselineScenario);
+        }
+        navigate("/");
+        resetErrorBoundary();
```
jessex, line 26: What happens if an error happens in the calls before this one within the `then` block? Are we left in some broken state where any page wrapped by this error boundary will fail because we didn't reset the boundary? I'm wondering if we should be doing something like a try-finally to ensure the reset is _always_ called.

src/hooks/useFacilityModelVersions.tsx
```diff
@@ -2,21 +2,28 @@ import { useCallback, useEffect, useState } from "react";
...
+      const modelVersions = await getFacilityModelVersions({
+        facilityId: facility.id,
+        scenarioId: facility.scenarioId,
+        distinctByObservedAt: true,
+      });
+      setFacilityModelVersions(modelVersions);
+    } catch (e) {
+      rethrowSync(e);
+    }
+  }, [facility, rethrowSync]);
```
sprad, line 26: Do we no longer need `setFacilityModelVersions` in this array?

src/page-multi-facility/MultiFacilityErrorBoundary.tsx
macfarlandian: good question; I did some digging on this. 

If an error happens _inside_ the fallback error recovery page, we are entering catastrophic failure territory. As it currently stands, if we fail to execute the `navigate` call you'll just be looking at the loading spinner forever. We could also toast that error, but we can't do anything more useful than that unless we come up with a second fallback state that is less likely to raise an error. (For what it's worth, I think an error inside this effect is quite unlikely, but it's certainly not impossible, especially if you are suddenly unable to reach firestore for some reason.)

If we naively reset the error boundary without fully executing our fallback behavior, we'll just trigger an infinite error loop of some sort (of the original permissions error, and possibly also of whatever error is happening in the fallback component, if it happens every time).

The best solution in that case may just be to render something more useful than a loading spinner when we find ourselves unable to recover from error we've caught.

jessex: Got it. It sounds like this may be a scenario that is so rare that we should not engineer for it straight-away. If you disagree and think we should create some fallback copy that we display indicating a "catastrophic" failure has occurred, feel free to pursue that. Otherwise, I think we can proceed?

macfarlandian: agreed, i'm not too worried about it at this point

src/hooks/useFacilityModelVersions.tsx
macfarlandian: we don't, the functions returned by `useState` are guaranteed to have stable references and don't need to be included here

src/hooks/useRejectionToast.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const { addToast } = useToasts();
+
+  return useCallback(
+    async (p: Promise<any>) => {
+      try {
+        return await p;
+      } catch (e) {
+        addToast(e.message);
```
SoumyaMahadevan, line 13: Since this is an error toast, we probably want to change the color from green to red? You can pass an "appearance" param here and Toast will take care of that.

src/database/index.tsx
```diff
@@ -613,6 +641,10 @@ export const saveFacility = async (
   } catch (error) {
     console.error("Encountered error while attempting to save a facility:");
     console.error(error);
+    throwIfPermissionsError(
+      error,
+      "You don't have permission to edit this facility.",
+    );
```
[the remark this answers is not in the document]
macfarlandian: turns out that deleteScenario is actually accessible in the UI now so good call on both counts

src/hooks/useRejectionToast.tsx
macfarlandian: oh good note, totally missed that!

macfarlandian:
@jessex I added this handler to the Excalibur population modal too ... basically what happens in that case is that it rejects any changes you may have made to the baseline population and just continues on with whatever defaults or previous values it has to fall back on.

PR #481, opened 2020-05-24
[Cleanup] Remove unused files and consolidate files (components/providers/contexts)

rememberlenny:
## Description of the change

There are a number of unused test files. This removes them.

Also there is no consistency for how components, providers, and context files are organized. There are a number of files that are being used in different pages (`/facility` or `/scenario`), but they are organized in the same root `page-*` folders. 

To clean things up, the components are starting to be moved in a `/components` folder. Also the providers and contexts are going to be moved to the `/providers` and `/contexts` folders.

## Type of change

### Development

### Code review


rememberlenny:
Adding @sprad and @macfarlandian in case they want to give it a look.

rememberlenny:
Great points @szhu - I'll incorporate some of these changes. Feel free to follow up after this. 

The main purpose of the original PR is to move away from the dependence on Gatsby's file structure. Since the application began, the files here would benefit from not being structured alongside the `page-*` folder structure. The routing that happens on the client side is becoming more complex.

Feel free to follow up on this after its submitted as well. I'll consider this the PR to set the skeleton and then parts can continue moving.

szhu:
The main purpose of the original PR is to move away from the dependence on Gatsby's file structure.

The only file structure Gatsby has is to place pages in `src/pages`, and we haven't even really been doing that. There is no current existing dependence on Gatsby's file structure.

> the files here would benefit from not being structured alongside the `page-*` folder structure

What is the benefit? Currently when I look the directory listing of `src`, it gives me a good sense of what parts of this app there are.

> The routing that happens on the client side is becoming more complex.

Just to clarify, by "routing' do you mean page routing? I think the current routing is fairly simple, since there are 3 pages. If by "routing" you mean the component dependency graph, then yes I agree!

> Feel free to follow up on this after its submitted as well. I'll consider this the PR to set the skeleton and then parts can continue moving.

My concern is that in the current state of this PR, the new file structure _could_ for some use cases be more confusing than the current file structure. Also after this PR, is would not be obvious that there is a reorganization in progress. I think it might be worth to ask someone who has recently added new features whether it is easy to find things under the new structure.

Also, since reorganizations can move a lot of files around and require everyone with open PRs to do some rebase/merge, consider making all of the necessary reorganization changes under a single PR rather than incrementally over time.

rememberlenny:
@szhu - It seems like you have a lot of opinions here, so Im going to actually close this PR and submit a separate one to delete the unused files. I won't have the time for the back and forth, but I do strongly think that the file structure has outgrown the app from how it started in March.

Since you originally started this, it might be a good task for you to layout some thoughts.

As someone who came into the app in May, I noticed that there is a fair amount of duplication due to redundant/unclear file structure.

szhu:
Hey @rememberlenny that makes a lot of sense, thanks for providing context.

I'm going to open #482 to keep track of this! I just don't want to rush the timeline since this change affects a lot of developers, it ideally should be reviewed by some of the developers who will be impacted by this day-to-day.

Just a note, even though more work is needed, the work you've already done is likely more important than the remaining work to get this merged -- it's super helpful have a real proposal with a concrete set of changes to discuss. So thank you again for making this PR.

Question: If/when I get to polishing up the reorg, is it okay if I reuse the same PR? Or would you like me to open a new one? (The main difference is that this affects the attribution on the merge commit that gets created.)

rememberlenny:
Feel free to do what ever is easiest!

—
Leonard Bogdonoff
RememberLenny.com | 949-322-8496

PR #485, opened 2020-05-26
Support "population" chart header in historical cases graph on facility details page.

### LINKED ISSUE #405: Viewing historical population data over time on the facility detail page
Currently, users have no way of viewing the historical case and population data they have entered for a facility. We want to encourage users to enter accurate historical population data because if we have their per-facility historical population data, we will be able to run Excalibur on a per-facility basis.

This is an addition to letting users see the historical case counts, we want them to be able to see the population data they have inputted over time on the facility detail page.

image

image

Components

1. Users by default see case and population data overlaid on top of one another. They can turn off the case data to just view the population data and vice versa.
2. Bar chart by default shows the most recent 30 days-worth of data. 
3. Clicking “Previous 15 days” scrolls the chart to be able to show the previous 15 days of data (if there is an easier chart library to handle scrolling back in time, feel free to use that)
4. Each bar is a click target that opens the “add historical data” modal with the date of the bar pre-selected
5. When there’s no data, the bar is red and stretches the full y-axis length of the chart (On click, opens the “Add historical data” modal with the date pre-selected to the date of the bar. On hover, color gets darker to show that the bar is a click target. Clicking on the hover text should also open the modal)
6. When there is data for a day, the bar is green and has a height of the # of confirmed cases (On click, opens the “Add historical data” modal with the date pre-selected to the date of the bar, on hover, shows the date, case count, and a CTA to click to change the case count. Click on the hover text should also open the modal)
7. P2: If a case count is 0 or less than 50% between the first y-axis line and 0, display the count above the bar
8. Button at the top right “+ Add Historical Data” to open the “add historical data” modal with yesterday’s date pre-selected (or the default we are currently using)
9. Because the “add historical data” modal allows users to be able save case and population data together, we do not anticipate the scenario in which they have case data entered but no population data or vice versa.

Mocks: https://www.figma.com/file/6z1QQW84a59Ul3c17PgWdV/%F0%9F%9A%80Recidiviz-CJStatus-(excalibur)?node-id=2206%3A7614

We considered showing a table or calendar widget that indicates historical data and data missing. A bar chart lets them see the trend over time. A line chart was also considered, but it is more challenging to interact with.

**Additional context**
This information is repeated in this doc, in case there are complications that warrant a longer discussion: https://docs.google.com/document/d/1KRCYPeC3W7KJm7flUpZKPxdhL7U0sRTfisD0cjqQYuM/edit#heading=h.qo0qd95ajfm1

This is deeply related to #404

SoumyaMahadevan:
## Description of the change

- Added chart header for population as a stacked bar chart.
- Changed both bar colors on hover.
- Allowed turning on and off chart headers for cases and population (default on).

Note: I changed the existing hover colors a bit so it looks better.
Kapture 2020-05-25 at 21 40 05

## Type of change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Configuration change (adjusts configuration to achieve some end related t#405 nctionality, development, performance, or security)

## Related issues

Closes #405

### Development

### Code review


src/page-multi-facility/HistoricalCasesChart/BarChartTooltip.tsx
```diff
@@ -47,19 +50,22 @@ const TooltipSubtitle = styled.div`
...
+  const { data } = summaryData;
   return (
     <TooltipContainer>
       {data.missing ? (
         <TooltipTitle>No data</TooltipTitle>
       ) : (
-        <TooltipTitle>{value} cases</TooltipTitle>
+        <>
+          <TooltipTitle>{data.cases} cases</TooltipTitle>
+          <TooltipTitle>{data.population + data.cases} residents</TooltipTitle>
```
jessex, line 61: It's not obvious to me why this part of the tooltip would sum population and cases. Shouldn't this just be `data.population`?

src/page-multi-facility/HistoricalCasesChart/index.tsx
```diff
@@ -23,14 +26,20 @@ function generateBarChartData(
     });
 
     if (existingVersion) {
+      const cases = totalConfirmedCases(existingVersion);
+      // This is necessary since semiotic doesn't have an option to display a stacked bar chart overlay.
+      // So we display population (non-cases) = total population - cases
```
jessex, line 31: I think this behavior is causing strange behavior. One case I have found: if population is not filled in for a given day, the bar will go below the Y-axis. To test, you can add a historical case on a date that looks something like this:

Screen Shot 2020-05-25 at 23 27 49

You'll see the bar look like this, stretching into negative:

Screen Shot 2020-05-25 at 23 26 50

src/page-multi-facility/HistoricalCasesChart/BarChartTooltip.tsx
jessex: Maybe I don't understand exactly what these variables refer to, because it seems to just display the population count in my local testing...

src/page-multi-facility/HistoricalCasesChart/BarChartTooltip.tsx
```diff
@@ -47,19 +50,22 @@ const TooltipSubtitle = styled.div`
...
+        </>
       )}
 
       <TooltipDate>
         <DateMMMMdyyyy date={data.observedAt} />
       </TooltipDate>
-      <TooltipSubtitle>Click to update</TooltipSubtitle>
+      <TooltipSubtitle>Change</TooltipSubtitle>
```
daschi, line 68: I think we might want to keep this as `Click to update` unless something has changed recently...this is different than the mock and comes from a slack chat I had with Serena after we talked about how it might be difficult to click on link within the hovered tooltip, so the idea was to change it to make it clear the user should click on the bar instead to open the modal.

src/page-multi-facility/HistoricalCasesChart/index.tsx
```diff
@@ -23,14 +26,20 @@ function generateBarChartData(
     });
 
     if (existingVersion) {
+      const cases = totalConfirmedCases(existingVersion);
+      // This is necessary since semiotic doesn't have an option to display a stacked bar chart overlay.
```
macfarlandian, line 30: sorry to be pedantic but FWIW Semiotic's api for stacking bars is correct, it is actually we who are abusing the concept slightly by identifying the second bar as "total population" rather than "uninfected population) ... the height of all stacked bars should add up to 100% of whatever the underlying total is, their contents should never overlap. So the calculations you did here are totally sensible and consistent with the chart form we are using (excepting the issue Josh mentions below)

src/page-multi-facility/HistoricalCasesChart/index.tsx
macfarlandian: to be fair, this is really the fault of our failing to validate the form inputs correctly; having a case count larger than a population count (in this case, zero) is nonsense data. It probably ought to be handled before it ever gets to the chart. The chart's behavior is correct when the input data doesn't violate the assumptions of what it's supposed to represent

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -318,6 +318,12 @@ export function getTotalPopulation(
   );
 }
 
+export function residentPopulation(
+  brackets: ModelInputsPopulationBrackets,
+): number {
+  return sum([getTotalPopulation(brackets), brackets.staffPopulation]);
```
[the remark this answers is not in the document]
SoumyaMahadevan: Yeah, was going to do that in a separate PR. Changed to totalIncarceratedPopulation.

src/page-multi-facility/HistoricalCasesChart/index.tsx
SoumyaMahadevan: True, a more appropriate way of stating this is that semiotic doesn't support overlay bars, which is what we want. I'll update the comment.

src/page-multi-facility/HistoricalCasesChart/BarChartTooltip.tsx
SoumyaMahadevan: The design requirement is for an overlay bar chart (i.e. both bars starting from the x axis). Semiotic supports a stacked bar chart where the second bar will start from the top of the first. So in order to get it to work like an overlay, I changed the second bar data to be the population - confirmed cases.
Since the tooltip needs to continuing showing the total population, I need to add these here. I can rename this to be non-cases or something like that to be clearer.

src/page-multi-facility/HistoricalCasesChart/index.tsx
SoumyaMahadevan: Yeah, I figured we should never have a scenario where the total population is less than the case count.
Since we don't have input validation as of now, I'll handle this in code.

src/page-multi-facility/HistoricalCasesChart/BarChartTooltip.tsx
SoumyaMahadevan: Ok, the designs in #405 say "Change" so I updated it. Will leave it as "Click to update".

src/page-multi-facility/HistoricalCasesChart/BarChartTooltip.tsx
jessex: Thanks for the explanation, @SoumyaMahadevan! I understand now.

Issue #407, opened 2020-05-13
Calculate RT breaks when inputs result in large rt value


rememberlenny:
**Describe the bug**
The `/calculate_rt` endpoint returns opaque errors when the case data results in large rt values.

This can be very confusing when someone is looking to test the RT value results.

**To Reproduce**

1. Create a new facility and enter `1000 current cases`, and `10000 total population`. (ie. 5/10/20 with 1000 cases)/
2. Go back to the dashboard and add a previous days data. Values should be 1 day before the current date with `100 current cases`. Don't change total population. Save. (ie. 5/9/20 with 100 cases)
3. Add another previous days data, one week before the last day, with `717 cases`. (ie. 5/2/20 with 717 cases)/
4. There should be no issues. Output from calculate_rt endpoint will be the following:

```
`{“Rt”: [{“date”: “2020-05-03”, “value”: 0.0}, {“date”: “2020-05-09", “value”: 12.0}, {“date”: “2020-05-10”, “value”:
8.63}], “high90”: [{“date”: “2020-05-03”, “value”: 10.81}, {“date”: “2020-05-09", “value”: 12.0}, {“date”: “2020-05-10”,
“value”: 9.3}], “low90”: [{“date”: “2020-05-03”, “value”: 0.0}, {“date”: “2020-05-09", “value”: 11.8}, {“date”:
“2020-05-10”, “value”: 7.930000000000001}]}`
```

Now do the same thing as above, but instead the last day of data should be 718 cases. This will result in error.

1. Create a new facility and enter `1000 current cases`, and `10000 total population`. (ie. 5/10/20 with 1000 cases)/
2. Go back to the dashboard and add a previous days data. Values should be 1 day before the current date with `100 current cases`. Don't change total population. Save. (ie. 5/9/20 with 100 cases)
3. Add another previous days data, one week before the last day, with `718 cases`. (ie. 5/2/20 with 718 cases)/
4. This will result in the following error:

```
{"error": "Traceback (most recent call last):\n File \"/user_code/realtime_rt.py\", line 134, in compute_r_t\n hdis =
highest_density_interval(posteriors, p=.9)\n File \"/user_code/realtime_rt.py\", line 89, in highest_density_interval\n
return pd.DataFrame([highest_density_interval(pmf[col], p=p) for col in pmf],\n File \"/user_code/realtime_rt.py\", line
[...]

PR #487, opened 2020-05-26
Set Rate of Spread panel chart range to 0-5

### LINKED ISSUE #386: R(t) - Set 'Rate of Spread' tab range to max value of at least 5
The 90% confidence intervals are fairly large in the 'Rate of Spread' tab, likely because of the small numbers that many facilities have, but also because the range in the chart only goes up to 2 (whereas real-world facilities may see rates of spread considerably higher than this).

We should make the range 0-5 in all cases except when a facility on the chart goes above 4, at which point we should go up to at least the max value in the chart+1.

None

**Additional context**
None

daschi:
## Description of the change

Set the Rate of Spread panel chart's x-axis from 0-5 for all cases except when the Rt range is greater than or equal to 4, then set it to the max plus 1. 

@macfarlandian While I was testing this I tried to add historical cases to a facility to try and get an Rt calculation that went beyond 4 and while doing it I noticed this error whenever my cases number was very close to the total residents number (i.e. I put 20,000 cases for 25,000 residents) Just wanted to check if this is an expected error or if I should open an issue for it:

image

## Type of change

## Related issues

Closes #386 

### Development

### Code review

### DIFF
--- a/src/rt-comparison-chart/RtComparisonChart.tsx
+++ b/src/rt-comparison-chart/RtComparisonChart.tsx
@@ -265,7 +265,7 @@ const RtComparisonChart: React.FC<{ data: RtComparisonData[] }> = ({
     Math.max(...data.map((record) => record.values.high90 || 0)),
   );
 
-  const maxExtent = maxRtHigh90 >= 4 ? maxRtHigh90 + 1 : 5;
+  const maxExtent = Math.max(maxRtHigh90 + 1, 5);
 
   return (
     <RtComparisonChartWrapper>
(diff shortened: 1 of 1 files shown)


macfarlandian:
I think that error is related to the fact that your case counts go up and down (they should be cumulative, i.e. never decreasing). We aren't validating that properly but there's already an issue for that: #407

src/rt-comparison-chart/RtComparisonChart.tsx
```diff
@@ -261,10 +261,12 @@ export const isRtComparisonData = (
 const RtComparisonChart: React.FC<{ data: RtComparisonData[] }> = ({
   data,
 }) => {
-  const maxExtent = Math.ceil(
+  const maxRtHigh90 = Math.ceil(
     Math.max(...data.map((record) => record.values.high90 || 0)),
   );
 
+  const maxExtent = maxRtHigh90 >= 4 ? maxRtHigh90 + 1 : 5;
```
macfarlandian, line 268: why not just `Math.max(maxRtHigh0 + 1, 5)`?

daschi: Ah yes that much better, I'll update this soon.

Issue #490, opened 2020-05-26
Run script over Firebase to use `facilityOccupancyPct` values to populate `facilityCapacity` 


cawarren:
**Is your feature request related to a problem? Please describe.**
Issue #203 updates the model behavior to take in the actual facility capacity, rather than the current % of that capacity, so that the model does the work to figure out % capacity rather than making the user do it.

As a result, we're going to need to run a script over the current Firestore (or in FE JS when one value is found to exist and the other doesn't, or something else) to populate the new facility capacity field (`facilityCapacity`) using the old data (`facilityOccupancyPct`).

**Describe the solution you'd like**
We should populate `facilityCapacity` for facilities where a) there is no current `facilityCapacity` value, and b) there is a current `facilityOccupancyPct` value.

The new `facilityCapacity` value should be the old `facilityOccupancyPct` value multiplied by the sum of the resident population in the facility for the last data point where both (population and `facilityOccupancyPct`) are present.

**Describe alternatives you've considered**
- We could leave the value blank, effectively wiping this value for all users who've previously entered a `facilityOccupancyPct` value

**Additional context**
- See Issue #203

macfarlandian:
this migration script was run succesfully in concert with #507

PR #497, opened 2020-05-27
FacilityInputForm - add 'Capacity' field and use it calculate `facilityOccupancyPct`

### LINKED ISSUE #203: Update 'Capacity' field to take actual facility capacity
We currently ask the user to enter the % their facility is over or under capacity, forcing them to perform this calculation themselves.

However, the capacity of a facility is a known and static thing, and is easier for them to provide once then forget about. We should be calculating over/under capacity based on that.

Update the 'Capacity' box to ask for the facility's capacity, instead of for how far above or below it they are. Then use that to calculate ourselves from the population they've entered.

**Additional context**
- This is also generally useful information to us
- Easier input for the user / set-and-forget.

jovergaag:
Note: Do not merge: relies on #490 

## Description of the change

Changes the 'Occupancy Rate' input field on the facility input form to 'Capacity'. We are allowing the user to now enter a capacity number, rather than using their known capacity number to calculate the occupancy percentage rate themselves. The occupancy percentage will be calculated based on the `facilityCapacity` input when the curve inputs are calculated.

- Adds a `facilityCapacity` field on the Epidemic Model
- This new `facilityCapacity` field is used to calculate the `facilityOccupancyPct` in the infection model `curveInputsFromUserInputs` and `curveInputsWithRt`
- Removes the `facilityOccupancyPct` from the Epidemic Model since this is now a calculated value.

This PR will have to be merged and deployed in coordination with #490 which will populate the `facilityCapacity` field based on current `facilityOccupancyPct` and population data. Once that data is populated, we need to immediately merge and deploy this PR to ensure those new `facilityCapacity` values do not become stale.

## Type of change

## Related issues

Closes #203

### Development

### Code review


src/database/__tests__/testcases.tsx
```diff
@@ -21,7 +21,6 @@ saveFacility(scenarioId, {
     countyName: "Ascension",
     rateOfSpreadFactor: "moderate",
     usePopulationSubsets: true,
-    facilityOccupancyPct: 0.85,
```
jovergaag, line 24: I don't see anywhere that we're running this test. Should we remove this test file?

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -42,7 +42,7 @@ export interface ModelInputsPopulationBrackets {
 }
 interface ModelInputsPersistent extends ModelInputsPopulationBrackets {
   facilityDormitoryPct?: number;
-  facilityOccupancyPct?: number;
+  facilityCapacity?: number | undefined;
```
macfarlandian, line 45: I think technically you don't need the extra `undefined`, the `?` covers that case. Similarly, on line 64, if the field can be undefined you don't need to add it there at all; that object is really just for making keys required that were optional in the interfaces it inherits from (convoluted, I know ... an artifact of my learning curve with TypeScript).

jovergaag: Thanks for pointing that out. I'm also still trying to learn TypeScript...sometimes it feels like unraveling a big ball of string. I ended up making a couple of other changes here as I really thought about what the different interfaces were being used for and how the undefined default is handled.

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -61,7 +61,7 @@ export interface EpidemicModelInputs extends ModelInputsUpdate {
   rateOfSpreadFactor: RateOfSpread;
   usePopulationSubsets: boolean;
   facilityDormitoryPct: number;
-  facilityCapacity: number | undefined;
+  facilityCapacity?: number;
```
macfarlandian, line 64: you should be able to just remove this entirely; it's just duplicating what's in `ModelInputsPerisistent` (which this extends, by way of `ModelInputsUpdate`

jovergaag: @macfarlandian Thanks for another look at this.

I had thought about removing it but noticed that this interface is described as 'some fields are required for calculations, define them here'. All of the other fields here are also defined in other interfaces inherited through `ModelInputsUpdate` so I figured this interface was more for documentation than actually defining anything new. I'll go ahead and remove it here, just curious about your thought process is about how to chose which fields remain defined in this interface vs via inheritance?

macfarlandian: FWIW, knowing what I know now about TypeScript, I wish I had defined all the inputs directly in `EpidemicModelInputs` and derived `ModelInputsUpdate` and `ModelInputsPersistent` via `Partial<>` and `Pick<>`. Same end result but I think it would be more readable and less error-prone.

PR #498, opened 2020-05-28
#441 update Scenario Library modal

### LINKED ISSUE #441: Update the Scenario Library to Show All Scenarios Accessible by a User

To support the user-to-user sharing work, we will need to update the existing Scenario Library modal to show all Scenarios a user has access to.

Once the updated [getScenarios function](https://github.com/Recidiviz/covid19-dashboard/issues/440) is in place, we will need to update the Scenario Library to show Scenarios that have been shared with that user.  The ability to delete a Scenario that a user is not the owner of should be removed from the options list.  Additionally, the ordering of the Scenarios should be as follows:

* The current user’s baseline scenario is always first.
* The current users remaining Scenarios should be ordered by name
* The Scenarios shared with a user should be visually separated from the users' own Scenarios and should also be ordered by Baseline and name.

sokunthel:
## Description of the change

Update the Scenario Library to Show All Scenarios Accessible by a User

**Before:**
ScenarioLibrary_Before

**After:**
ScenarioLibrary_After

## Type of change

## Related issues

Closes #441

### Development

### Code review

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,7 +36,12 @@
     "@typescript-eslint/no-explicit-any": "off",
 
     // Allow unused variables that start with "_"
-    "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "^_"
+      }
+    ],
 
     // Ensure default exports have the same name as the file.
     "filenames/match-exported": "error",
@@ -65,7 +70,14 @@
     // Because we want imports sorted!
     // The built-in "sort-imports" doesn't seem to be sufficient, and the
     // "eslint-plugin-import" doesn't seem to work.
-    "simple-import-sort/sort": "warn"
+    "simple-import-sort/sort": "warn",
+
+    // Disable warning:
+    // "React Hook useEffect has a missing dependency: ... Either
+    // include it or remove the dependency array"
+    // However, this kind of warning seems to be useful when working with hooks (performance optimizations)
+    // https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
+    "react-hooks/exhaustive-deps": 0
   },
 
   "settings": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -70,14 +70,7 @@
     // Because we want imports sorted!
     // The built-in "sort-imports" doesn't seem to be sufficient, and the
     // "eslint-plugin-import" doesn't seem to work.
-    "simple-import-sort/sort": "warn",
-
-    // Disable warning:
-    // "React Hook useEffect has a missing dependency: ... Either
-    // include it or remove the dependency array"
-    // However, this kind of warning seems to be useful when working with hooks (performance optimizations)
-    // https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
-    "react-hooks/exhaustive-deps": 0
+    "simple-import-sort/sort": "warn"
   },
 
   "settings": {
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -73,7 +73,7 @@ const authenticate = async () => {
   await firebase.auth().signInWithCustomToken(customToken);
 };
 
-const currrentUserId = () => {
+export const currrentUserId = () => {
   const userId = (firebase.auth().currentUser || {}).uid;
 
   if (!userId) {
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -221,7 +221,11 @@ export const getScenarios = async (): Promise<Scenario[]> => {
     // We're sorting in memory instead of in the query above due to a limitation
     // in the way our data is currently modeled. See this issue for more details:
     // https://github.com/Recidiviz/covid19-dashboard/issues/253
-    return orderBy(scenarios, ["baseline", "name"], ["desc", "asc"]);
+    return orderBy(
+      scenarios,
+      ["baseline", (scenario) => scenario.name.toLowerCase()],
+      ["desc", "asc"],
+    );
   } catch (error) {
     console.error(`Encountered error while attempting to retrieve scenarios:`);
     console.error(error);
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -73,7 +73,7 @@ const authenticate = async () => {
   await firebase.auth().signInWithCustomToken(customToken);
 };
 
-export const currrentUserId = () => {
+const currrentUserId = () => {
   const userId = (firebase.auth().currentUser || {}).uid;
 
   if (!userId) {
--- a/src/page-multi-facility/ScenarioLibraryModal.tsx
+++ b/src/page-multi-facility/ScenarioLibraryModal.tsx
@@ -346,8 +346,7 @@ const ScenarioLibraryModal: React.FC<Props> = ({ trigger }) => {
       </ScenarioLibrary>
     );
   };
-  console.log(sharedScenarios);
-  console.log(currentUserId);
+
   return (
     <Modal
       modalTitle="Library"
--- a/src/page-multi-facility/ScenarioLibraryModal.tsx
+++ b/src/page-multi-facility/ScenarioLibraryModal.tsx
@@ -1,12 +1,7 @@
 import React, { useEffect, useState } from "react";
 import styled from "styled-components";
 
-import {
-  currrentUserId,
-  deleteScenario,
-  duplicateScenario,
-  getScenarios,
-} from "../database";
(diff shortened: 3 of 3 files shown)


src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -316,6 +372,47 @@ const ScenarioLibraryModal: React.FC<Props> = ({ trigger }) => {
             </DeleteModalContents>
           </ModalDialog>
         </ScenarioLibrary>
+
+        {sharedScenarios?.data?.length ? <h3>Shared Scenarios</h3> : null}
+        <ScenarioLibrary>
```
sprad, line 377: This `<ScenarioLibrary>` library section is pretty repetitive with the `<ScenarioLibrary>` found above.  Is there a way we can refactor this code to reduce the duplication?

src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -165,11 +182,50 @@ const ScenarioLibraryModal: React.FC<Props> = ({ trigger }) => {
   const [showDeleteModal, setShowDeleteModal] = useState(false);
   const rejectionToast = useRejectionToast();
 
+  const sortScenarios = (list: any) => {
```
sprad, line 185: It should be possible to achieve this same sorting result using lodash's built in `orderBy` function in the same way we're using it here: https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/index.tsx#L224

You may have to tweak it a bit to get case insensitive sorting to work by following this example:  https://stackoverflow.com/questions/37848030/lodash-how-to-do-a-case-insensitive-sorting-on-a-collection-using-orderby

src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -165,11 +182,50 @@ const ScenarioLibraryModal: React.FC<Props> = ({ trigger }) => {
...
+        }
+      });
+
+      setOwnedScenarios({
+        data: oScenarios,
+        loading: false,
+      });
+
+      setSharedScenarios({
+        data: sortScenarios(sScenarios),
```
sprad, line 228: Assuming that insertion order is preserved with `push`, I _think_ we can just keep all of the sorting logic in the `getScenarios` function and just have this function tease apart the `owned` and `shared` scenarios without having to resort them here.

sprad:
I'm seeing the following visual bug in Chrome when I am the owner of 5 or more scenarios:

Screen Shot 2020-05-28 at 8 55 08 AM

To reproduce, just keep duplicating scenarios until you are the owner of 5 or more

.eslintrc.json
```diff
@@ -65,7 +70,14 @@
...
-    "simple-import-sort/sort": "warn"
+    "simple-import-sort/sort": "warn",
+
+    // Disable warning:
+    // "React Hook useEffect has a missing dependency: ... Either
+    // include it or remove the dependency array"
+    // However, this kind of warning seems to be useful when working with hooks (performance optimizations)
+    // https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
+    "react-hooks/exhaustive-deps": 0
   },
```
macfarlandian, line 81: I don't think this is a good idea ... we are violating it all over the place, which is annoying, but it's usually for a reason that involves some tech debt we have incurred by moving so fast. Having the warnings (or explicit overrides to suppress them) is a good thing IMO

src/page-multi-facility/ScenarioLibraryModal.tsx
macfarlandian: Yes I second this; the incoming sort order of `scenariosData` should be preserved by this code

.eslintrc.json
macfarlandian: FWIW the usual suspect for this warning in our codebase is a function that has to be defined within the render function; it turns out that the `useCallback` hook can help with that (https://reactjs.org/docs/hooks-reference.html#usecallback)

There are a few examples of us using it to help with this issue now, you can see it e.g. here: https://github.com/Recidiviz/covid19-dashboard/blob/master/src/hooks/useError.tsx#L6

src/database/index.tsx
```diff
@@ -73,7 +73,7 @@ const authenticate = async () => {
   await firebase.auth().signInWithCustomToken(customToken);
 };
 
-const currrentUserId = () => {
+export const currrentUserId = () => {
```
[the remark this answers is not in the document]
sokunthel: updated!

src/page-multi-facility/ScenarioLibraryModal.tsx
sokunthel: updated!

src/page-multi-facility/ScenarioLibraryModal.tsx
sokunthel: I was trying to tweak the `orderBy` in my first attempt but was not aware of the lodash. Anyways, tried above suggestion and it seems to do the trick. Thanks!

src/page-multi-facility/ScenarioLibraryModal.tsx
sokunthel: suggestion implemented!

.eslintrc.json
sokunthel: reverted, and implemented `useCallback`. Thanks!

sokunthel:
> I'm seeing the following visual bug in Chrome when I am the owner of 5 or more scenarios:
> 
> Screen Shot 2020-05-28 at 8 55 08 AM
> 
> To reproduce, just keep duplicating scenarios until you are the owner of 5 or more

src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -154,32 +163,47 @@ type Props = Pick<ModalProps, "trigger">;
...
+          sScenarios.data.push(scenario);
+        }
       });
 
+      oScenarios.loading = false;
+      sScenarios.loading = false;
+
+      setAllScenarios([oScenarios, sScenarios]);
+
       setBaselineScenario(scenariosData.find((scenario) => scenario.baseline));
```
sprad, line 200: I think there is a subtle bug in this line.  If the user deletes the scenario they are currently viewing, they should be taken back to _their_ baseline scenario.  At the moment, when a user deletes the scenario they are currently viewing, they are taken back to the first baseline scenario found in the `scenarioData` which may or may not be the user's baseline scenario.

src/page-multi-facility/ScenarioLibraryModal.tsx
sprad: I see that we got rid of the duplication in this `ScenarioLibrary` which is great, but I think the ` ind === 0` toggles make things a bit challenging to read because it can be difficult for people reading the code to extract meaning from having to know the significance of a specific index position.

I think keeping the scenarios in separate `owned` and `shared` arrays and looping through them individually like you had before was easier to read.  Admittedly, my comment wasn't very guiding but what I was hoping would be possible would be to extract some of the duplicated markup out into its own component (internal to this file).  Then we'd pass the arrays separately into the new component to generate the correct HTML output.  The tricky bit would be handling the subtle differences like showing the delete option in the dropdown.

If this proves to be too challenging, then I think its ok to leave things as they are.

src/page-multi-facility/ScenarioLibraryModal.tsx
sprad: I _think_ the fix for this is roughly:

```javascript
setBaselineScenario(scenariosData.find((scenario) => scenario.baseline && scenario.roles[currentUserId] === "owner"));
```

src/page-multi-facility/ScenarioLibraryModal.tsx
sprad: Since we'd like to get a release out today, I'd say maybe just timebox this to 30-60 minutes.  If it's taking longer than that, we can merge and address later

src/page-multi-facility/ScenarioLibraryModal.tsx
sokunthel: updated!

src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -239,6 +274,80 @@ const ScenarioLibraryModal: React.FC<Props> = ({ trigger }) => {
...
+              </DeleteButton>
+              <CancelButton onClick={closeDeleteModal}>Cancel</CancelButton>
+            </ModalButtons>
+          </DeleteModalContents>
+        </ModalDialog>
+      </ScenarioLibrary>
+    );
+  };
+  console.log(sharedScenarios);
+  console.log(currentUserId);
```
sprad, line 350: Can we get rid of these `console.logs`?

sokunthel: done!

PR #495, opened 2020-05-27
Verify cases counts increase monotonically in Rt function

macfarlandian:
## Description of the change

The Rt function requires cumulative case counts to produce a valid output; currently it does not validate the inputs for this requirement, and it behaves inconsistently when it is violated. This adds a check for obvious violations (decreases over time) and responds with an error when they are found.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

Closes #494 

### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #499, opened 2020-05-28
Improve Rt function error handling

macfarlandian:
## Description of the change

Makes Rt failures less opaque, with a combination of backend and frontend changes that support more useful error messaging (something we wanted but deferred in the initial Rt release): 

1. Make Python logging more reliable (with traceable Execution Ids and exception logging) so we can respond with a cleaner error message suitable for UI display
1. On the frontend, create an explicit Rt error object that can be propagated through the UI instead of a `null`. Handle this object everywhere that deals with Rt data (this is the bulk of the PR because we pass it around quite a bit)
1. When we have one of these error objects, include its message in any existing "error" or "unknown value" UI elements. @cawarren this is easy to either change or suppress in favor of additional design input so I decided to just take a stab at implementation and ask for your sign-off here before merging. 

New error messages have been added to:

- The tooltips on Rt pills in the Scenario Projections tab
Screen Shot 2020-05-27 at 6 03 41 PM

- The tooltips on the chart in the Scenario Rate of Spread tab
Screen Shot 2020-05-27 at 6 04 57 PM

- The error message for the Rate of Spread timeseries chart on the Facility page
Screen Shot 2020-05-27 at 6 04 13 PM

There are currently three calc-related error messages that users are likely to encounter (copy suggestions welcome, these did not come off one of our official web copy spreadsheets):
- "Case counts must be cumulative and monotonically increasing": just added in #495, hopefully self-explanatory
- "Not enough data to compute R(t); at least two days of non-repeating case counts are required":  slightly tricky but this is for when we don't have enough usable days (<2, after filtering out duplicates; it would be possible to vary this error before and after deduplication but we currently don't)
- "Unable to compute R(t) with the provided data": any other currently unknown error type

## Type of change

## Related issues

Part of #407 

### Development

### Code review

### DIFF
--- a/src/rt-timeseries/RtTimeseriesContainer.tsx
+++ b/src/rt-timeseries/RtTimeseriesContainer.tsx
@@ -5,11 +5,11 @@ import styled from "styled-components";
 import Colors from "../design-system/Colors";
 import HelpButtonWithTooltip from "../design-system/HelpButtonWithTooltip";
 import Loading from "../design-system/Loading";
-import { RtData } from "../infection-model/rt";
+import { getRtStatus, RtData } from "../infection-model/rt";
 import { updateFacilityRtData } from "../infection-model/rt";
 import AddCasesModal from "../page-multi-facility/AddCasesModal";
 import { FacilityContext } from "../page-multi-facility/FacilityContext";
-import { Facility } from "../page-multi-facility/types";
+import { Facility, RtValue } from "../page-multi-facility/types";
 import RtTimeseries from "./RtTimeseries";
 
 const borderStyle = `1px solid ${Colors.paleGreen}`;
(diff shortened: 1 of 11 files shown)


src/rt-timeseries/RtTimeseriesContainer.tsx
```diff
@@ -5,11 +5,11 @@ import styled from "styled-components";
 import Colors from "../design-system/Colors";
 import HelpButtonWithTooltip from "../design-system/HelpButtonWithTooltip";
 import Loading from "../design-system/Loading";
-import { RtData } from "../infection-model/rt";
+import { isRtData, isRtError, RtData } from "../infection-model/rt";
```
jessex, line 8: Just this one lint issue.

macfarlandian:
I got an offline approval from @cawarren via Slack so I'm proceeding with merge, thanks @jessex

PR #500, opened 2020-05-28
test page to run migration script within our auth

### LINKED ISSUE #490: Run script over Firebase to use `facilityOccupancyPct` values to populate `facilityCapacity`

macfarlandian:
## Description of the change

This will not be merged, but can be reviewed before I run it locally; it's a migration function (nestled lovingly within a temporary test page behind our auth wall) that calculates and adds `facilityCapacity` to all facilities and their model versions, where possible (i.e. if they have the two existing fields). This relies on a temporary Firestore rule that gives me "superuser" access from the website, which I will have to activate before running this and then revert, so the code itself is safe (will result in a permissions error if you try to run it without being covered by said superuser rule).

I did a full test run (including writes) against the `scenarios_test` collection and everything looked fine; I have done a dry run against the real `scenarios` collection and it looks like there are ~120 documents (mostly model versions) that won't be updated by this (mostly because they don't have any population counts), which doesn't seem too bad.

## Type of change

## Related issues

Closes #490 

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -878,7 +878,6 @@ export const addFacilityCapacity = async ({
   const newInputs = { ...modelInputs, facilityCapacity };
   const timestamp = currrentTimestamp();
   newInputs.updatedAt = timestamp;
-  delete newInputs.facilityOccupancyPct;
 
   const db = await getDb();
   let docToUpdate = db
@@ -891,10 +890,7 @@ export const addFacilityCapacity = async ({
     docToUpdate = docToUpdate
       .collection(modelVersionCollectionId)
       .doc(modelVersionId);
-    updateData = {
-      ...newInputs,
-      facilityOccupancyPct: firebase.firestore.FieldValue.delete(),
-    };
+    updateData = newInputs;
   }
   try {
     if (write) {
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -909,7 +909,7 @@ export const addFacilityCapacity = async ({
   }
 };
 
-export const removeOccupancyPct = async ({
+export const doFirestoreCleanup = async ({
   scenarioId,
   facilityId,
   modelVersionId,
@@ -920,6 +920,10 @@ export const removeOccupancyPct = async ({
   const timestamp = currrentTimestamp();
   newInputs.updatedAt = timestamp;
   delete newInputs.facilityOccupancyPct;
+  // round off any non-integer values that resulted from the pct calc
+  if (newInputs.facilityCapacity) {
+    newInputs.facilityCapacity = Math.round(newInputs.facilityCapacity);
+  }
 
   const db = await getDb();
   let docToUpdate = db
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -1,3 +1,4 @@
+// @ts-nocheck
 // the core Firebase SDK must be imported before other Firebase modules
 // eslint-disable-next-line simple-import-sort/sort
 import * as firebase from "firebase/app";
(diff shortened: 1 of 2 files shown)


sprad:
All things considered, I dig this pattern and the helpful write up 🙏

src/database/index.tsx
```diff
@@ -838,3 +839,76 @@ export const deleteScenario = async (
...
+    .collection(facilitiesCollectionId)
+    .doc(facilityId);
+  let updateData = { modelInputs: newInputs, updatedAt: timestamp };
+  if (modelVersionId) {
+    docToUpdate = docToUpdate
+      .collection(modelVersionCollectionId)
+      .doc(modelVersionId);
+    updateData = {
+      ...newInputs,
+      facilityOccupancyPct: firebase.firestore.FieldValue.delete(),
```
jessex, line 896: Why do we add this delete when we're talking about a model version?

src/pages/test.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+// eslint-disable-next-line react/display-name
+export default () => {
+  const migrate = async () => {
+    const scenarios = await getMigrationScenarios();
+    scenarios.forEach(async (scenarioDoc) => {
+      const scenarioId = scenarioDoc.id;
+      const facilities = await getMigrationFacilities(scenarioId);
+      facilities.forEach(async (facilityDoc) => {
+        setFacilityCapacity({ facilityDoc, scenarioId: scenarioId });
```
jessex, line 67: Why do you have the explicit key here but not below on line 73?

src/pages/test.tsx
```diff
@@ -0,0 +1,89 @@
...
+  let modelVersionId;
+
+  if (modelVersionDoc) {
+    modelInputs = modelVersionDoc.data();
+    modelVersionId = modelVersionDoc.id;
+  }
+
+  const capacity = calculateFacilityCapacity(modelInputs);
+  if (capacity) {
+    addFacilityCapacity({
```
jessex, line 41: It doesn't matter but it bothers me that the order is not the same as the order listed in the function signature itself. I know it doesn't matter.

src/database/index.tsx
macfarlandian: for the facility, we are replacing the entire value of `facility.modelInputs` with this new input object, whereas for a model version there is no `modelInputs` key so it just merges the objects together. To remove this field I need to send an explicit delete command.

src/pages/test.tsx
macfarlandian: this call updates `facility.modelInputs` (the "current" input version that we store in the facility document itself) ... the one on line 73 updates each of the documents in this facility's `modelVersions` collection.

src/database/index.tsx
macfarlandian: this is now moot anyway because after talking with @jovergaag I realized it was probably premature to delete this field anyway, since it will break stuff in existing deployment while we're waiting for her feature to be released. that should make this migration non-destructive and we can do another pass later to clean up the unused field after we are sure that everything is copacetic post-release

macfarlandian:
both migrations have been run successfully. closing this without merge.

PR #501, opened 2020-05-28
Add functionality to duplicate a facility

### LINKED ISSUE #204: Add 'Duplicate facility' functionality
We currently enable users to set up facilities, but some states have dozens of the things. Setting up your baseline scenario could take a long time if you're going through and manually selecting all of the same policy inputs / etc. for each one.

Add a 'Duplicate facility' button on the Facility Details view, which copies your inputs into a new facility and puts you in the Facility Details view of the new one. This let's a user fill out facility info once, then make copies and just update name / population info.

Note: There's no expectation of any meaningful relationship between the original and duplicated facilities. This is solely a convenience feature to copy inputs from an already-entered facility into a new one.

**Additional context**
- [Mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=1636%3A130) (see Duplicate Facility button at the top of the Facility Details view)

daschi:
## Description of the change

This adds a `Duplicate` option in the popup menu for the facility input page. It saves a new facility with all of the current inputs copied over (and no historical / version data), and sets the facility name to be `Copy of "facility name"`. I also added a toast for when the page is finished re-rendering. 

I needed to add a `useEffect` to "refresh" the internal state of the `facility` for the `RtTimeseriesContainer` component and the `EditInPlace` which is used for the name and description. This is because updating the `FacilityContext` does not actually re-render all of the children components and the internal state of `facility` stays the same.  

## Type of change

## Related issues

Closes #204 

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -540,20 +540,18 @@ export const removeScenarioUser = async (
   }
 };
 
-const batchSetFacilityModelVersions = async (
+const batchSetFacilityModelVersions = (
   modelVersions: ModelInputs[],
   facilityDocRef: firebase.firestore.DocumentReference,
   batch: firebase.firestore.WriteBatch,
 ) => {
-  modelVersions.forEach((modelVersion: ModelInputs) => {
-    const modelVersionDoc = facilityDocRef
-      .collection(modelVersionCollectionId)
-      .doc();
+  const modelVersionDoc = facilityDocRef
+    .collection(modelVersionCollectionId)
+    .doc();
 
+  modelVersions.forEach((modelVersion: ModelInputs) => {
     batch.set(modelVersionDoc, modelVersion);
   });
-
-  return batch;
 };
 
 /**
@@ -704,7 +702,7 @@ export const duplicateFacility = async (
       scenarioId,
       facilityId,
     });
-    await batchSetFacilityModelVersions(modelVersions, facilityRef, batch);
+    batchSetFacilityModelVersions(modelVersions, facilityRef, batch);
 
     // Delete old facility id and set new createdAt timestamp on payload
     delete facilityCopy.id;
@@ -829,7 +827,8 @@ export const duplicateScenario = async (
         scenarioId,
         facilityId: facility.id,
       });
-      await batchSetFacilityModelVersions(modelVersions, facilityDoc, batch);
+
+      batchSetFacilityModelVersions(modelVersions, facilityDoc, batch);
     }
 
     await batch.commit();
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -540,6 +540,22 @@ export const removeScenarioUser = async (
   }
 };
 
+const batchSetFacilityModelVersions = async (
+  modelVersions: ModelInputs[],
+  facilityDocRef: firebase.firestore.DocumentReference,
+  batch: firebase.firestore.WriteBatch,
+) => {
+  modelVersions.forEach((modelVersion: ModelInputs) => {
+    const modelVersionDoc = facilityDocRef
+      .collection(modelVersionCollectionId)
+      .doc();
+
+    batch.set(modelVersionDoc, modelVersion);
+  });
+
+  return batch;
+};
+
 /**
  * Please note the following usage patterns and provide data to this method
  * accordingly when updating a facility:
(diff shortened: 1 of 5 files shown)


src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -178,7 +187,26 @@ const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
   const closeDeleteModal = () => {
     updateShowDeleteModal(false);
   };
-  const popupItems = [{ name: "Delete", onClick: openDeleteModal }];
+  const onDuplicateFacility = async () => {
+    if (facility) {
+      await rejectionToast(
```
sprad, line 192: Thank you for remembering to take care of the shared scenario case 🙏

daschi:
thanks for taking a look! I'll check in with @cawarren on slack to confirm the data we want to duplicate before merging.

src/page-multi-facility/FacilityInputForm.tsx
daschi: TBH I'm didn't know I had taken care of it 😅  could you explain how it is taken care of? I'm not very familiar with that new feature yet.

sprad: 😄 

By including this `rejectionToast` call, a user can only modify Scenarios and Facilities they own.  So if you're viewing a facility that lives within a Scenario owned by someone else and you try to duplicate that facility, an error toast will be shown.  This is the expected and desired behavior because we don't want people duplicating facilities within someone elses's scenario.

daschi:
@jessex @sprad This is ready for another review! It's been updated to copy all of the versions from the facility being duplicated. I chatted with @sprad  about whether we would want to change the `createdAt` dates for the new versions, and decided to keep them the same to match what we do when we duplicate a scenario.

src/database/index.tsx
```diff
@@ -540,6 +540,22 @@ export const removeScenarioUser = async (
   }
 };
 
+const batchSetFacilityModelVersions = async (
+  modelVersions: ModelInputs[],
+  facilityDocRef: firebase.firestore.DocumentReference,
+  batch: firebase.firestore.WriteBatch,
+) => {
+  modelVersions.forEach((modelVersion: ModelInputs) => {
+    const modelVersionDoc = facilityDocRef
```
sprad, line 549: You should be able to retrieve this modelVersionDoc outside of the `forEach` loop

src/database/index.tsx
```diff
@@ -540,6 +540,22 @@ export const removeScenarioUser = async (
...
+      .collection(modelVersionCollectionId)
+      .doc();
+
+    batch.set(modelVersionDoc, modelVersion);
+  });
+
+  return batch;
```
sprad, line 556: We end up mutating this `batch` object inside of this method so I don't know that there's any reason to return it.

src/database/index.tsx
sprad: Nevermind, we're creating a new `doc` each time

src/database/index.tsx
```diff
@@ -540,6 +540,20 @@ export const removeScenarioUser = async (
...
 };
 
+const batchSetFacilityModelVersions = (
+  modelVersions: ModelInputs[],
+  facilityDocRef: firebase.firestore.DocumentReference,
+  batch: firebase.firestore.WriteBatch,
+) => {
+  const modelVersionDoc = facilityDocRef
+    .collection(modelVersionCollectionId)
+    .doc();
```
sprad, line 550: Sorry, I gave you bad advice.  I see now that this is creating a new modelVersion doc for each model version, which is desired.  With this outside of the loop, we end up only creating a single new modelVersion with whatever the last values are in the modelVersions array.

daschi: Oooh, yeah makes sense! sorry I didn't look at it closely when updating...will change it back

src/database/index.tsx
```diff
@@ -684,7 +740,7 @@ export const deleteFacility = async (
     //
     // * https://firebase.google.com/docs/firestore/manage-data/delete-data#web_2
     // ** https://github.com/Recidiviz/covid19-dashboard/issues/191
-    const facilityDocRef = await scenarioRef
+    const facilityDocRef = scenarioRef
```
sprad, line 743: Do we no longer need to `await` or did we never need to `await` in the first place?

daschi: Yeah we never needed it since `.doc()` doesn't return a promise, and removing it removed a linter warning.

PR #506, opened 2020-05-29
Avoiding Button Nesting Warning

sprad:
## Description of the change

- Previously the app was throwing a warning because we were trying to nest a button within a button.  I've updated the button container to instead be a div.

## Type of change

## Related issues

Closes N/A

### Development

### Code review

### DIFF
--- a/src/scenario-share/ScenarioShareModal.tsx
+++ b/src/scenario-share/ScenarioShareModal.tsx
@@ -100,7 +100,7 @@ const UserList: React.FC<{
   );
 };
 
-const ShareButtonContainer = styled.button`
+const ShareButtonContainer = styled.div`
   align-items: center;
   display: flex;
   justify-content: space-between;
--- a/src/scenario-share/ScenarioShareModal.tsx
+++ b/src/scenario-share/ScenarioShareModal.tsx
@@ -19,7 +19,7 @@ import useRejectionToast from "../hooks/useRejectionToast";
 import { ScenarioUsers } from "../page-multi-facility/types";
 import useScenario from "../scenario-context/useScenario";
 import AvatarCluster from "./AvatarCluster";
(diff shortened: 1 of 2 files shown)


sprad:
Prior to this change, the following warning was always present in the console:

Screen Shot 2020-05-29 at 12 53 04 PM

src/scenario-share/ScenarioShareModal.tsx
```diff
@@ -100,7 +100,7 @@ const UserList: React.FC<{
   );
 };
 
-const ShareButtonContainer = styled.button`
+const ShareButtonContainer = styled.div`
```
sprad, line 103: @macfarlandian - I'm assuming you wanted a `div` here?  Let me know if there's another HTML element you'd prefer.

macfarlandian: oh, thank you ... I didn't notice this warning until after merging that PR and haven't gotten around to fixing it.
Because this whole thing is clickable I think we do actually want this to be a button, but we want `ScenarioShareButton` to .... not be a button. and maybe to not be named that anymore

sprad: For what it's worth this entire `div` is still clickable.  If I add `cursor: pointer;` to the styles, the experience will be identical.

macfarlandian: yeah I know, it's just to keep the markup more semantic. functionally no difference (except possibly for people using screen readers)

sprad: I flipped things around in this latest commit to make the inner button a "link" container:  https://github.com/Recidiviz/covid19-dashboard/pull/506/commits/e141cff4985166796a3f58c96a962f3df67b0a28

PR #512, opened 2020-06-01
Add tests to CI.

teymour-aldridge:
## Description of the change
Runs the test suite against all commits to master and pull requests which makes it easy to check that all commits and pull requests pass.

## Type of change

### Development

- [x] Manual testing has been performed locally (not really applicable/possible)

### Code review


.github/workflows/tests.yml
```diff
@@ -7,8 +7,8 @@ on:
   pull_request:
 
 jobs:
-  lint-all-check:
-    name: yarn lint-all-check
+  test:
```
jessex, line 10: Test coverage is good to add to the CI, but we don't want to remove lint checks. We should have separate jobs for those tasks.

teymour-aldridge: I've left the list coverage in the pre-existing, but I can move both tests into the same file (or workflow) if that's better.

jessex: Oh my apologies, Teymour! I misunderstood what was happening here because 1) I was accidentally looking at only your second commit and thought this was differential instead of additive, and 2) it's early in my timezone. This looks great to me with clearer eyes. Thanks!

teymour-aldridge: It might be more efficient to run lints and tests as part of the same job, because `yarn install` only has to be run once.

teymour-aldridge:
No problem! You look to be doing a really cool (and important) job here, and I'm happy to be able to contribute in some way.

teymour-aldridge:
That test fails locally on my machine as well. I think it's because JSDom (which Jest uses because it runs tests using node.js instead of in a browser) doesn't implement `window.crypto` which causes issues because Auth0 requires `window.crypto` to function. There's a relevant feature request in [the JSDom repository](https://github.com/jsdom/jsdom/issues/1612) but it looks like it's been around for a while and won't be tested.

src/auth/AppAuth0ClientPromise.tsx
```diff
@@ -31,7 +31,7 @@ const finalAuthConfig = {
  */
 const AppAuth0ClientPromise: Promise<Auth0Client> =
   // Note: The window variable is not defined static rendering.
-  typeof window === "undefined"
+  typeof window.crypto === "undefined"
```
teymour-aldridge, line 34: I've updated this if statement (not sure if ifs are statements or expressions in Javascript/Typescript) to match for `window.crypto` instead of `window` because `window` is emulated by Jest (using JSDom) when tests are run but Jest doesn't include code for `window.crypto` which Auth0 relies upon.

teymour-aldridge:
> Thanks, @teymour-aldridge! This looks good in theory. However, I see that there is a test failing and it looks like it has to do with the execution environment afforded by Github vs the expectation of that test. Any idea for a workaround for that particular issue, assuming you can see the failure?

I've come up with a workaround and the tests are now passing!

src/auth/AppAuth0ClientPromise.tsx
jessex: Before going forward with this change, since it's auth-related and high stakes, I'd like @szhu to weigh in and see if he has any concerns.

teymour-aldridge: Cool! (I think `window.crypto === "undefined"` should be true if `window === "undefined"`)

szhu: I think this is fine as long as you document in a code comment:

1.  Why you are testing for `window.crypto`
2.  How the user can manually test that things that this trick depends on still hold true in the future

For example, the existing comment, "The window variable is not defined static rendering." is a statement that can easily be tested in the future. (However, I am missing a statement about why this check is there, which should be something like "The Auth0 client code breaks during static render, but we also don't need it during static render." It would be nice if you can help me add that in as well.)

---

Here are 3 specific things that you can follow to meet the goals above, but if you have another way to address these concerns feel free! Also happy to discuss options here and on Slack.

1.  Why do we need to skip the Auth0 stuff during testing?
2.  How did you check (and how can someone in the future check) that `window.crypto` is undefined during testing?
3.  Please integrate these comments with the comment about static render, such that it makes sense as a whole.

Here's my best guess at what a good final comment will look like, feel free to use this and correct where I'm wrong:

> We need to not use the Auth0 client...
> 
> *   during testing, because it breaks under this environment and because our current test code is not testing auth-related functionality, or
> *   during static render, because it breaks under this environment and because the auth state is different for each visitor anyway.
> 
> We use `typeof window.crypto === "undefined"` because this works to detect both the testing and static render environments. (\`typeof window === undefined\` only detects the static render environment.

szhu:
Not related to the content of this PR, but I see that the Now deploy failed. Clicking details gives me this message:



@jessex I don't really know the context behind this and haven't seen it before, let me know if it's safe to click Authorize. (I'm also fine not clicking it, I'm just curious since I haven't seen this before.)

jessex:
@szhu, I approved the deploy! It happened because this PR was coming from an external fork, I believe.

macfarlandian:
the change to `AppAuth0ClientPromise` discussed here has since been made in another PR (the same issue started to break tests locally as well) so I would like to go ahead and merge this, unless you have any remaining concerns @jessex ?

jessex:
@macfarlandian that sounds fine by me. Sorry for the huge delay on this one, it got away from me.

Let's just resolve that conflict so we're only merging in the yaml configuration and merge it.

macfarlandian:
@jessex yeah i'll take care of that. thanks!

PR #518, opened 2020-06-02
Fix sorting/uniqueness error in model versions

### LINKED ISSUE #515: getFacilityModelVersions doesn't return the most recently updated version
**Describe the bug**
Calling `getFacilityModelVersions` with the `distinctByObservedAt` option should deduplicate multiple observations for the same day by choosing the one with the most recent `updatedAt` value. But it does not do so consistently.

**To Reproduce**

1. Go to the facility page for any facility
2. Save the facility multiple times with different total case counts
3. Open the add cases modal for the facility
4. See error: the form is prepopulated with the case counts you added most recently, but when you open the date picker the case count in the tooltip for today will be wrong, and if you select today then the form will be repopulated with wrong data (the wrong data in this case should be the first value you saved for the current day)

**Expected behavior**
Values should all match in the prepopulated form, the date picker, and the form after selecting the current date, and they should be equal to the most recent saved version. 

This bug manifests in a number of places because it is a bug in the underlying database fetch function; the repro steps above are just the easiest place to see the error in the UI. The fix should make sure the error is corrected in any place that depends on the result of `getFacilityModelVersions` (the Rt pipeline, input form validation, etc)

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

**Additional context**

I'm pretty sure I see where the error is happening: we rely on sort order in the query as defined [here](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/database/index.tsx#L342), expecting it to sort all documents observed on the same day in descending order by `updatedAt`; however, because we do not consistently normalize `observedAt` values by day, observations on the same day may have different timestamps, in which case this query ordering clause does nothing, and what is returned by this function instead is the document with the earliest `observedAt` timestamp.

macfarlandian:
## Description of the change

Detects cases where we have same-day but unequal timestamps among fetched model versions and re-sorts the versions to prepare them for our uniqueness-by-day filter. (Not every facility necessarily has this problem, so we skip the extra work for them.)

I also updated `saveFacility` to normalize all `observedAt` timestamps to the start of the day, which hopefully will avoid introducing this problem for new facilities. Patching the existing data is potentially tricky though, due to access and timezone issues, so I'm not actually doing this here for the moment. (We might want to actually consider changing the underlying datatype as part of that so it's kind of a larger conversation.)

There was also a bug in the recently merged historical case validation that I encountered during this work — it was requiring that _all_ fields in the model increase over time, not just the case counts (lol whoops 😎). I fixed that here too.

## Type of change

## Related issues

Closes #515 

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -4,7 +4,13 @@ import * as firebase from "firebase/app";
 import "firebase/auth";
 import "firebase/firestore";
 
-import { format, startOfDay, differenceInCalendarDays } from "date-fns";
+import {
+  format,
+  startOfDay,
+  differenceInCalendarDays,
+  isSameDay,
+  isEqual,
+} from "date-fns";
 import { pick, orderBy, uniqBy, findKey, maxBy, minBy } from "lodash";
 import { Optional } from "utility-types";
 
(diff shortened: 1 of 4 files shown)


src/database/index.tsx
```diff
@@ -349,13 +350,34 @@ export const getFacilityModelVersions = async ({
...
+        );
+      });
+      if (!normalized) {
+        modelVersions = orderBy(
           modelVersions,
-          // make sure time doesn't enter into the uniqueness
-          (record) => record.observedAt.toDateString(),
-        )
-      : modelVersions;
+          [(version) => version.observedAt.toDateString(), "updatedAt"],
```
jessex, line 369: Is this sorting first by observedAt and then secondarily by updatedAt? Ascending by observedAt and then descending by updatedAt? I might not be thinking clearly but wouldn't we want the two sorts to be in the same direction?

macfarlandian: yeah it's a quirky thing (you will see we are doing the same thing in the query) ... the reason is that `uniqBy` returns the first item in array order that matches its predicate, and that's what we're using to filter duplicate days

jessex: Ah, I understand. Might make sense to add a comment explaining that but I don't feel strongly towards it. It's not unintuitive, it just relies on a bit more context to get.

PR #520, opened 2020-06-02
Reuse curve chart on model inspection page

### LINKED ISSUE #269: Reuse updated chart components on the model inspection page

macfarlandian:
## Description of the change

By parameterizing the minor differences between the two pages (namely, which curves we include and what colors they are), this makes `ChartArea` reusable on both the facility page and the model inspection pages. There should be no change in behavior in multifacility views, but the inspection page comes up to date with the latest chart improvements (e.g. toggling legend, letting hospital beds fall off the Y axis) and should now stay up to date. Redundant chart components for the model inspection page have been removed.

## Type of change

## Related issues

Closes #269 

### Development

### Code review


src/infection-model/seir.tsx
```diff
@@ -56,6 +57,13 @@ export const seirIndexList = Object.keys(seirIndex)
   // these should all be numbers anyway but this extra cast makes typescript happy
   .map((k) => parseInt(seirIndex[k as any]));
 
+// though obscure, this has the effect of filtering out the reverse-mapping keys
```
jessex, line 60: Crazy obscure! I would almost argue for replacing it with a simple function that does the filtering, unless you feel strongly about this.

src/page-multi-facility/FacilityRow.tsx
```diff
@@ -103,6 +97,7 @@ const FacilityRow: React.FC<Props> = ({ facility, onSave }) => {
 
   const chartData = useChartDataFromProjectionData(
     useProjectionData(model, true, facilityRtData),
+    // curvesToShow,
```
jessex, line 100: Remove this line?

src/infection-model/seir.tsx
macfarlandian: i'm not so sure that would be any better .... the alternative to this in typescript-land would be just manually creating a union type of all the enum keys that we would have to keep in sync with the enum by hand. A function that could serve as a type guard would probably still require that, plus we would have a function that we would have to call all over the place (which is not currently the case) rather than just relying on static analysis as here. 

This is super hacky though, so if you would be more comfortable with an explicit lists and loud comments about keeping it in sync, I can get with that

jessex: No worries, we can stick with it as is.

PR #545, opened 2020-06-09
#524 fix overlap issue in Scenario Library Cards

### LINKED ISSUE #524: Owned/Shared Scenario Library Cards Overlap
**Describe the bug**

In the Scenario Library, when too many cards are added, the "owned" scenario cards (i.e. those where my user has the `owner` roll) begin to overlap with the "shared" scenario cards (i.e. those where my user has the `viewer` roll).  See the screenshots below:

**Normal Case - Owned and Shared scenario cards are separate**

Screen Shot 2020-06-03 at 2 01 31 PM

**Error Case - Owned cards overlap with the Shared scenario cards**

Screen Shot 2020-06-03 at 2 01 03 PM

**To Reproduce**

0.  Ensure that your user has access to some shared scenarios (i.e. is a `viewer` on someone else's scenario(s)).
1. Open the Scenario Library Modal
2. Keep duplicating scenarios until the bug presents itself.  Unfortunately, the bug isn't consistent, but I've typically seen it appear whenever I am the owner of 5 of more scenarios.

**Expected behavior**

Regardless of how many scenarios I own and regardless of how many scenarios are shared with me, I should not see the owned and shared scenarios overlap.

- Mac / Chrome - 83.0.4103.61

sokunthel:
## Description of the change

> Fix owned/shared scenario library cards overlap issue

## Type of change

## Related issues

Closes #524

### Development

### Code review

### DIFF
--- a/src/page-multi-facility/ScenarioLibraryModal.tsx
+++ b/src/page-multi-facility/ScenarioLibraryModal.tsx
@@ -23,7 +23,7 @@ const ModalContents = styled.div`
   font-weight: normal;
   justify-content: flex-start;
   margin-top: 30px;
-  height: 100%;
+  height: auto;
 
   h3 {
     border-bottom: 0.5px solid #c8d3d3;
--- a/src/page-multi-facility/ScenarioLibraryModal.tsx
+++ b/src/page-multi-facility/ScenarioLibraryModal.tsx
@@ -23,7 +23,7 @@ const ModalContents = styled.div`
   font-weight: normal;
   justify-content: flex-start;
   margin-top: 30px;
-  height: auto;
+  height: 100%;
 
   h3 {
     border-bottom: 0.5px solid #c8d3d3;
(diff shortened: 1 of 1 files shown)


src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -23,7 +23,7 @@ const ModalContents = styled.div`
   font-weight: normal;
   justify-content: flex-start;
   margin-top: 30px;
-  height: 100%;
+  height: auto;
```
sprad, line 26: This causes a subtle UI degradation that probably isn't worth worrying about, but that we should at least be aware of.  The degradation is that the `Loading` icon is now shifted somewhat awkwardly to the top of the modal during a transition state (i.e. when we're duplicating a scenario).

**Before (height: 100%)**

Screen Shot 2020-06-09 at 9 46 27 AM

**Before (height: auto)**

Screen Shot 2020-06-09 at 9 47 26 AM

The ideal fix for the original issue would be one that didn't also cause this minor UI degradation.  I probably wouldn't spend much time on it, but it's likely worth timeboxing an investigation to see if we can get the best of both worlds.

sokunthel: great catch. I've made a few changes in CSS and DOM structure around where we render the loading component. Please take another look. Thanks!

PR #552, opened 2020-06-10
Simplify state management in Add Cases modal

### LINKED ISSUE #541: Add Cases modal missing input data on first click from Add historical data button on facility page
**Describe the bug**
When opening the Add Cases modal from the `Add historical data` button above the bar chart on the facility page for the first time, the inputs are blank. They should be auto-filled with the latest case and population numbers. When you click on the button a second time the numbers are there. 

**To Reproduce**

1. Go to a facility page through the multi facility dashboard
2. Click on the `Add historical data` button above the bar chart
3. The inputs for cases and populations are blank
4. Close the modal and click on the button again, the inputs are filled in

**Expected behavior**
Expect to see the latest case and population numbers filled in

**Screenshots**
addcasesbug

- OS [e.g. Windows 10]: Catalina 10.15
- Browser [e.g. chrome, edge]: Chrome 
- Version [e.g. 22]: 83

**Additional context**

macfarlandian:
## Description of the change

The bug referenced below was happening due to a discrepancy between initialization and reset of the modal, in the `useAddCasesInputs` hook. If no version matched the provided date, an effect hook would clear all the inputs. However, when the modal was closed, a reset function would set them to some default values rather than clearing them. The slightly convoluted result was that the inputs would only ever be blank the first time you opened the modal, if you didn't open it by clicking on a date in the historical cases chart that already had data (i.e. you clicked on a blank date or the "add historical data" button). 

I don't think we actually _ever_ want to clear the inputs — they should always either have either the values retrieved from the version history, or default values.

While the above issue was the proximate cause, rather than simply patch the issue what I have wound up doing (after some consultation with @daschi) is to clean up the state management in the hook function to make it less redundant and imperative, and thus (hopefully) less bug-prone. There is now a "one-way" cascade of side effects such that the `observedAt` argument triggers a series of state changes: the current date, then the user data for that date, then the desired form input values. Instead of returning a function for callers to change the internally stored date, it now requires a new `observedAt` argument to be passed.

This should fix the bug and also hopefully make this hook a bit easier to work with in general.

## Type of change

## Related issues

Closes #541 

### Development

### Code review

### DIFF
--- a/src/hooks/useAddCasesInputs.tsx
+++ b/src/hooks/useAddCasesInputs.tsx
@@ -1,6 +1,6 @@
 import { startOfDay } from "date-fns";
 import { pick } from "lodash";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 
 import { saveFacility } from "../database";
 import { useToasts } from "../design-system/Toast";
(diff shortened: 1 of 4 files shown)


daschi:
Sorry for the delay! I can take a look at this later this evening

PR #562, opened 2020-06-11
[Report automation] Admin page displaying stats for states and counties

### LINKED ISSUE #531: [Report automation] Admin page to show public stats for all states
We are currently pulling data manually to create weekly reports for a handful of states. The long-term goal is to have an admin page that can automatically generate a report for any state based on the information in the web tool. While #521 is still in progress, this is a step that can be parallelized that moves us toward automation (more context in [PRD](https://docs.google.com/document/d/1WwT52cF1I-Wmn7FomkUkMxFT93yTtvP8XGCRYuKKRQs/edit#heading=h.gvtiq0smp79p))

Create a page at model.recidiviz.org/admin/weeklysnapshot (or other appropriate URL), where the admin can select a state from a dropdown of all 50 states. For that state, the following information is printed on the page (data sources linked in PRD):
- “Number of cases in the state” (Source: NYT)
- “Change in state cases since last week” = +/-N (we have a script that does this, ping serena!)
- "Number of total ICU beds in the state" (Source: our excel model)
- “Counties to watch”: List of up to 4 counties in the state with the greatest percent per capita increase in cases in the past 7 days, as well as the percent per-capita increase. Ordered from highest to lowest (Source: NYT)
- “***” if a DOC facility is located in that county or a neighboring county (Source: facility dataset)

If we do not have a given stat, “?” will be printed instead.

Note: This step saves us 10 minutes or manual work per report - 3.5 hours for all 50 reports!

Example page output, all of the data is fake (no formatting or UX considerations because this page will not be displayed publicly):
image

To-date, we have been pulling the numbers manually and will continue to do so for a handful of states, but this process is not scalable beyond 10 states.

We also considered jumping straight to automation with high-fidelity visuals and formatting, but we expect to get feedback from states in our current manual pilot that alters the content of the final report, so we do not want to spend unnecessary eng effort polishing the visuals until we are confident about the final content.

**Additional context**
As more functionality gets built in the web tool, this page will expand to show more content.
Parent issue: #170

daschi:
## Description of the change

This adds an admin page at `/weekly-snapshot` which loads the states and counties CSV data from the NYT github repo and displays stats for the states and counties. The "***" feature which puts those stars next to counties that have a facility in or near the county is blocked by the facility metadata workstream so is skipped for now:

image

@macfarlandian I used the `useAdminUser` hook at the `pages/weekly-snapshot.tsx` level, but let me know if we want to use it to whitelist users for this page at a different level. 
 
## Type of change

## Related issues

Closes #531 

### Development

### Code review

### DIFF
--- a/src/page-weekly-snapshot/NYTDataProvider.tsx
+++ b/src/page-weekly-snapshot/NYTDataProvider.tsx
@@ -94,13 +94,13 @@ async function fetchNYTData() {
     fetchHistoricalNYTStatesData(),
   ]);
 
-  const dayOne = latestStateData[0].date;
-  const daySeven = subWeeks(dayOne, 1);
+  const daySeven = latestStateData[0].date;
+  const dayOne = subWeeks(daySeven, 1);
   const filteredCountyData = historicalCountyData.filter((d) =>
-    isSameDay(d.date, daySeven),
+    isSameDay(d.date, dayOne),
   );
   const filteredStateData = historicalStateData.filter((d) =>
-    isSameDay(d.date, daySeven),
+    isSameDay(d.date, dayOne),
   );
   let data: { [key: string]: any } = {};
   for (const stateData of latestStateData) {
--- a/src/page-weekly-snapshot/NYTDataProvider.tsx
+++ b/src/page-weekly-snapshot/NYTDataProvider.tsx
@@ -81,6 +81,45 @@ async function fetchHistoricalNYTStatesData() {
   return fetchCSVData(historicalStatesDataURL);
 }
 
+async function fetchNYTData() {
+  const [
+    latestCountyData,
+    historicalCountyData,
+    latestStateData,
+    historicalStateData,
+  ] = await Promise.all([
+    fetchLatestNYTCountyData(),
+    fetchHistoricalNYTCountyData(),
+    fetchLatestNYTStatesData(),
+    fetchHistoricalNYTStatesData(),
+  ]);
+
+  const dayOne = latestStateData[0].date;
+  const daySeven = subWeeks(dayOne, 1);
+  const filteredCountyData = historicalCountyData.filter((d) =>
+    isSameDay(d.date, daySeven),
+  );
+  const filteredStateData = historicalStateData.filter((d) =>
+    isSameDay(d.date, daySeven),
+  );
+  let data: { [key: string]: any } = {};
+  for (const stateData of latestStateData) {
+    if (stateData.state) {
+      data[stateData.state] = {
+        state: [
+          stateData,
+          ...filteredStateData.filter((d) => d.state === stateData.state),
+        ],
+        counties: [
+          ...latestCountyData.filter((d) => d.state === stateData.state),
+          ...filteredCountyData.filter((d) => d.state === stateData.state),
+        ],
+      };
+    }
+  }
+  return data;
+}
+
 export const NYTDataProvider: React.FC<{ children: React.ReactNode }> = ({
   children,
 }) => {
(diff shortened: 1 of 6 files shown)


src/page-weekly-snapshot/WeeklySnapshotPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+  dayOneCases: number | undefined,
+  countyPopulation: number | undefined,
+) {
+  if (
+    (!daySevenCases && daySevenCases !== 0) ||
+    (!dayOneCases && dayOneCases !== 0) ||
+    !countyPopulation
+  )
+    return;
+  return (daySevenCases - dayOneCases) / countyPopulation;
```
jessex, line 69: Just want to shine a light and confirm this is in fact the right calculation for "per capita increase in cases" and that there's no additional need to, for example, multiply by 100 to get to the right percentage. I don't have any explicit reason to think this is not correct, but it's the kind of thing that can be tricky so want to be sure.

src/page-weekly-snapshot/NYTDataProvider.tsx
```diff
@@ -0,0 +1,171 @@
...
+      const latestStatesDataURL =
+        "https://raw.githubusercontent.com/nytimes/covid-19-data/master/live/us-states.csv";
+      return fetchCSVData(latestStatesDataURL);
+    }
+
+    async function fetchHistoricalNYTStatesData() {
+      const historicalStatesDataURL =
+        "https://raw.githubusercontent.com/nytimes/covid-19-data/master/us-states.csv";
+      return fetchCSVData(historicalStatesDataURL);
+    }
```
macfarlandian, line 92: seems like you could define these outside the render loop since they don't depend on any local variables

src/page-weekly-snapshot/NYTDataProvider.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const NYTDataContext = React.createContext<NYTDataContext | undefined>(
+  undefined,
+);
+
+function transformRow(row: DSVRowString): NYTCountyRecord | NYTStateRecord {
+  return {
+    state: row.state,
+    county: row.county,
+    date: row.date ? addDays(new Date(row.date), 1) : startOfToday(),
```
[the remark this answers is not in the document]
daschi: I'm unsure about the best way to do this, but if it looks like `new Date()` always returns a day ahead, so I was seeing this:
```
:> new Date("2020-06-11")
=> Wed Jun 10 2020 17:00:00 GMT-0700 (Pacific Daylight Time)
``` 

..and adding a day returned the correct date. Is there a more standard way to parse the string into the correct date?

src/page-weekly-snapshot/WeeklySnapshotPage.tsx
daschi: It's not super clear but this number does get multiplied to 100 before displaying it on [Line 136](https://github.com/Recidiviz/covid19-dashboard/pull/562/files#diff-117efcb95f817f083faeb4c1c790f02aR136) with the [`numeral().format("0.000 %")`](https://github.com/adamwdraper/Numeral-js/blob/master/src/formats/percentage.js) formatter. If that's too hidden I could multiply it more explicitly and then format it with numeral without using their percentage formatter.

src/page-weekly-snapshot/NYTDataProvider.tsx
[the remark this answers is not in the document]
macfarlandian: I would recommend using datefns.parse instead (https://date-fns.org/v2.12.0/docs/parse) so you can explicitly define the expected format ... parsing dates from strings in JS is notoriously jacked up

Issue #521, opened 2020-06-03
[Shadow baseline] Create shadow baseline scenario that can "sync" to a user's baseline scenario


sychang:
**Is your feature request related to a problem? Please describe.**
[PRD for more context](https://docs.google.com/document/d/1e6yCfUUYIZW0FlTkzzwYg80-wymZ7PkTClo_k7bqUqk/edit?ts=5ed53a10#). We currently ask users to enter complete data about every facility in their system, daily, in order to get a picture of what’s happening in their system. This is an effort to reduce the data input work on their end in order to improve retention and reduce user friction. 

We already pull in county case and death data for every county in the United States, which reduces the end-user’s workload. To further reduce the end-user workload, we will pull in:
- Facility metadata (name, address, county, capacity, population, security level, year built)
- Case, test, and death count per-facility

This data has been collected by our partners and manual collection team over the course of the past month. It is patchy in some places, meaning we need to not promote this sync when the information is mostly unavailable. Data sources to power SBS can be found in the PRD.

**Describe the solution you'd like**
**Creating a Shadow Baseline Scenario**
We will create a Shadow Baseline Scenario (SBS) for each state, which auto-fills based on our data collection efforts. Users will not be able to see or affect the SBS - we will maintain it with external data regardless of any end-user activity.

If the user opts to turn on ‘Sync’ for their Baseline Scenario (BS), we’ll check once a day for updates to the SBS, and if found, propagate them to the user’s Baseline Scenario. Sync is unidirectional - it only applies from the SBS to the BS. 

**Sync behavior**
If the user overwrites a value in their BS that was propagated from the SBS, that value should not be overwritten again by the SBS. 

Examples:
If the user updates facility metadata (e.g. capacity, or the facility name), it will not be overwritten by the SBS moving forward. 
- In spite of this, the syncer should still be able to recognize the facility even if the name is different. TODO: Address UX of this.

If the user updates the case counts for facility ABC on day 6/1/2020 to ‘242’ (where the SBS had inserted ‘241’), that day will not be overwritten again by SBS. The syncer will update values for future days, however.
- In spite of this, for case- and death- data, the syncer should pull the prior day’s value if it’s larger than the value available in SBS (since larger numbers are presumably more recent / accurate)

Note that this means there is a 1:M relationship between the SBS and BSes, since multiple users may exist / have BSes in the same state.

**Describe alternatives you've considered**
[...]

PR #565, opened 2020-06-12
Fetch shadow data for baseline scenarios

macfarlandian:
## Description of the change

Does several things to introduce the "shadow data" feature into our codebase:

1. creates a feature flag
1. adds a new database function for fetching shadow facility data relevant to one's scenario. Currently this is pointed at a hand-made test collection with an agreed-upon shape; we can cut over to the real one once it stabilizes
1. adds a Firestore security rule to make said test collection read-only for all
1. If the flag is enabled and your baseline scenario is selected, fetches the corresponding shadow facilities and stores them in the `FacilitiesContext`. (I am pretty sure this is the behavior we will want for real, but it's easy enough to change if not.)

This should be enough to unblock the UI tasks around mapping your existing facilities to shadow facilities; the data they need should now be available in `FacilitiesContext`. There is still quite a bit more work required to actually display shadow data for mapped facilities, but I plan to move onto that next.

## Type of change

## Related issues

Part of #521 

### Development

### Code review


backend/firestore.rules
```diff
@@ -61,5 +61,9 @@ service cloud.firestore {
       allow update: if userIsSelf(resource) && userIsSelf(request.resource);
       allow create: if userIsSelf(request.resource);
     }
+
+    match /reference_facilities_test/{document=**} {
+  		allow read: if userIsAuthenticated();
+  	}
```
jessex, line 67: Some indentation funkiness here.

src/database/index.tsx
```diff
@@ -41,6 +42,9 @@ const scenariosCollectionId = "scenarios";
 const facilitiesCollectionId = "facilities";
 const modelVersionCollectionId = "modelVersions";
 const usersCollectionId = "users";
+// TODO: when the datasource stabilizes, change this to the real collection
```
jessex, line 45: For TODOs that will not be resolved in the current PR, let's always try to put an issue number with them, e.g. `TODO(XXX): When the datasource stabilizes...`

src/database/type-transforms.tsx
```diff
@@ -109,3 +111,67 @@ export const buildUser = (document: firebase.firestore.DocumentData): User => {
 
   return { ...pick(data, ["name", "email"]), id: document.id };
 };
+
+const buildCovidCase = (
+  doc: firebase.firestore.DocumentData,
+): ShadowCovidCase => {
+  const data = doc.data();
+
+  const observedAt = parse(doc.id, "yyyy-MM-dd", startOfToday());
```
jessex, line 120: Remind me why observedAt is start of today here?

src/database/type-transforms.tsx
```diff
@@ -109,3 +111,67 @@ export const buildUser = (document: firebase.firestore.DocumentData): User => {
...
+  population = population.map(toSimpleTimeseries);
+
+  return {
+    id: facilityDocument.id,
+    state,
+    canonicalName,
+    facilityType,
+    capacity,
+    population,
+    covidCases: facilityCovidCaseDocuments.map(buildCovidCase),
```
jessex, line 174: This is _extreme_ paranoia, but it may be good practice anyway: if we happen to be performing this `map` call while we pass over the midnight line, given that the internal call to `buildCovidCase` invokes `startOfToday()`, we'd have inconsistent dates on some of the results in the collection. It might be wise to pass a date into the function to use as `observedAt` and then initialize that once out here.

src/database/type-transforms.tsx
```diff
@@ -109,3 +111,67 @@ export const buildUser = (document: firebase.firestore.DocumentData): User => {
...
+    capacity,
+    population,
+    ...other
+  } = data;
+
+  // cast known types
+  state = String(state);
+  canonicalName = String(canonicalName);
+  facilityType = String(facilityType);
+  // if these are not arrays then unfortunately they are garbage
```
jessex, line 156: What would they be otherwise, and should we be throwing errors if the types are unexpected?

src/facilities-context/FacilitiesContext.tsx
```diff
@@ -89,6 +97,31 @@ export const FacilitiesProvider: React.FC<{ children: React.ReactNode }> = ({
     }
   }, [scenarioId, state.facilities, state.rtData]);
 
+  // fetch shadow facilities based on user facilities
+  const shouldFetchShadow = useShadowDataEligible();
+  useEffect(() => {
+    // clean up any existing shadow facility data
+    facilitiesActions.clearShadowFacilities(dispatch);
+
+    if (!shouldFetchShadow) return;
```
jessex, line 106: Do we still want to perform the `clearShadowFacilities` call above if this is false?

src/feature-flags/featureFlags.tsx
```diff
@@ -5,6 +5,8 @@ export type FeatureFlags = {
   // TODO: Delete this flag once #202 is done
   showRateOfSpreadTab: boolean;
   showImpactButton: boolean;
+  // TODO: can remove after successful launch
```
jessex, line 8: Same here, let's put an issue number on this TODO.

src/facilities-context/FacilitiesContext.tsx
macfarlandian: we do ... this value is unique per scenario, so it could flip from true to false when, e.g., you change from your baseline scenario to one of your other ones. I can see that it's not particularly intuitive though, I was actually struggling a bit with the best way to do this ... I think I might factor out the cleanup to a different effect that is closer to its cause

src/database/type-transforms.tsx
macfarlandian: there's not really a known reason why they would be something else ... this is just kind of a last-ditch effort to return the proper type if the Firestore document has been mangled somehow. We are not really doing this anywhere else, so it may be more trouble than it's worth? Or maybe, as you say, the more correct solution would be to throw an error (or to simply drop this document from the result set and maybe return an error that can be consumed alongside the valid data ... we don't really have a precedent here other than "assume everything is always fine" so I'm open to whatever

src/database/type-transforms.tsx
macfarlandian: the API for this function is a little weird; in this case we are getting the entire date (year, month, day) from the parsed string and `startOfDay` is a "reference date" that is required to fill in any missing pieces of a full Date object that aren't parsed; in this case we are only using the time portion and the local time zone of "today", which has the effect of normalizing `observedAt` to the start of whatever day it is.

This addresses your Extreme Paranoia as well, since the change of date as you cross the midnight boundary will just be ignored. 

I will grant you that this is rather non-obvious; perhaps an explanatory comment or even a refactor into a function with a more descriptive name would be in order here.

src/database/index.tsx
```diff
@@ -41,6 +42,9 @@ const scenariosCollectionId = "scenarios";
 const facilitiesCollectionId = "facilities";
 const modelVersionCollectionId = "modelVersions";
 const usersCollectionId = "users";
+// TODO (#521): when the datasource stabilizes, change this to the real collection
+const shadowDataCollectionId = "reference_facilities_test";
+const shadowDataCovidCasesCollectionId = "covidCases";
```
sprad, line 47: Annoying Nit:  Do we want to remove references to "shadow" in the codebase?  It seems to me this term may lose its meaning over time.

Also, "Data" here is probably too generic since as far as I know this will always be referring to facilities.

Soooo... because I'm boring, I'd suggest updating these `consts` and the other remaining functions, types, etc.  to "Reference Facility"

`referenceFacilitiesCollectionId`
`referenceFacilitiesCovidCasesCollectionId`
`getReferenceFacilities`
`ReferenceFacility`
`ReferenceCovidCase`

etc, etc...

sprad: If this feels like too much work or if we feel like "shadow" is self-explanatory, feel free to ignore this comment.

src/database/index.tsx
```diff
@@ -41,6 +42,9 @@ const scenariosCollectionId = "scenarios";
 const facilitiesCollectionId = "facilities";
 const modelVersionCollectionId = "modelVersions";
 const usersCollectionId = "users";
+// TODO (#521): when the datasource stabilizes, change this to the real collection
```
sprad, line 45: FYI, if anyone is interested in testing using real data, the `reference_facilities` collection has been populated for ~1300 facilities, each with roughly 90 days of daily covid case data.

src/page-multi-facility/types.tsx
```diff
@@ -64,3 +64,32 @@ export type ScenarioUsers = {
...
+  id: string;
+  state: string;
+  canonicalName: string;
+  facilityType: string;
+  capacity: SimpleTimeseries[];
+  population: SimpleTimeseries[];
+  covidCases: ShadowCovidCase[];
+  // various other metadata that we don't explicitly care about may also be present
+  [key: string]: unknown;
+};
```
sprad, line 95: I think you know this, but just as an FYI, the only data we're going to have out of the gate for reference facilities are:

```js
state
canonicalName
facilityType
covidCases
```

We won't have the other data (i.e. `capacity`, `population`, etc.) until we're at a point where we can import metadata from BigQuery.

That said, I think it's fine to leave these in for now as they represent the idealized future shape of these documents.

src/database/index.tsx
jessex: I agree with this, actually. Reference facility seems like a wiser long-term naming choice in the actual code.

src/page-multi-facility/types.tsx
macfarlandian: yes that's part of the reason why I want to keep using the test collection for now ... although while this typedef does require those fields to be present, it wouldn't object to those arrays being empty. I am assuming that is a case we are going to have to handle in the real world as well

src/database/index.tsx
macfarlandian: I was on the fence about this but I am definitely sympathetic to this argument .... with you weighing in now too I am inclined to just change it so that we don't regret the fact that it has a weird name later

Issue #543, opened 2020-06-09
When adding cases, make adding "ages unknown" and "by age" mutually exclusive


sychang:
**Is your feature request related to a problem? Please describe.**
The current add cases modal creates some confusion between entering cases for "ages unknown" as opposed to broken down by age. If a user has entered facility data broken down by age, we pre-populate the add cases modal with the population age breakdowns but hide that at first. This leads to accidental erroneous case / population inputs.

**Describe the solution you'd like**
We would like to update the add cases modal to reduce the likelihood of erroneous inputs.
1. Remove the arrow at the end of "Add residents and cases by age"
2. When a user clicks on "Add residents and cases by age", hide / remove the row of inputs for "RESIDENT POPULATION (AGES UNKNOWN)" and replace them with the inputs for entering cases and population broken down by age
3. Add "Resident" in front of the ages so users do not misinterpret it as needing staff age breakdowns.
4. Add a CTA to "Add residents and cases without ages," styled the same as the CTA in step 1. When clicked, the rows with age breakdowns disappear and the original "ages unknown" row reappears.

![image](https://user-images.githubusercontent.com/6633887/84103448-aa18bf00-a9d8-11ea-8ca5-e9c4983a7238.png)

5. Recreate this UX on the facility detail page, since that information mirrors what is on the add cases modal.

Default open logic: If a facility previously had cases or population broken down by age, open the modal with the age breakdowns visible by default (step 3). If there is no previous population data broken down by age or a new facility is being created, open the modal with "ages unknown" showing by default.

**Describe alternatives you've considered**
There are some more complex UX alternates, but we are trying to leverage the existing design and interaction as much as possible to minimize change on the user.

**Additional context**
- This change will also make it more straightforward to build projections based on shadow data, which has no cases by age.
- This will also make the initial onboarding less daunting because there appear to be fewer upfront inputs needed to make the model work.

glen0071:
@daschi, In responding to this issue, I was making changes inside the `FacilityInformation` component, importing `FacilityContext`, and then checking `facility.modelInputs` for the data needed to check if case data had been saved on the facility with ages known or unknown.

Now I see some commits you made in the last couple days that replace FacilityContext with a FacilitiesContext paradigm. Makes me wonder if it makes sense to pull the data into FacilityInformation from FacilitiesContext now? Or am I going about this wrongly?

glen0071:
@macfarlandian or @jessex you're welcome to chime in on this ^^

[...]

PR #568, opened 2020-06-12
Conditionally hide age vs unkonwn age input

glen0071:
## Description of the change
Conditionally hides input boxes for age vs unknown age cases and population in AddCases modal and Facility Details page.

Giph from multi-facility view:
multi-view

Giph from Facility Details view:
single-view

## Type of change

## Related issues

Closes [#543 ]

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/FacilityInformation.tsx
+++ b/src/impact-dashboard/FacilityInformation.tsx
@@ -65,6 +65,41 @@ const FormHeaderRow: React.FC<FormHeaderRowProps> = (props) => (
   </LabelRow>
 );
 
+const passedAgesKnown = (model: Record<string, any> | undefined) => {
+  if (model !== undefined) {
+    let keys = Object.keys(model);
+    return keys.some(function (key) {
+      return RegExp(/age\d+/).test(key);
+    });
+  } else {
+    return true;
+  }
+};
+
+const passedAgesUnknown = (model: Record<string, any> | undefined) => {
+  if (model !== undefined) {
+    let keys = Object.keys(model);
+    return keys.some(function (key) {
+      return RegExp(/ageUnknown\w+/).test(key);
+    });
+  } else {
+    return true;
+  }
+};
+
+const collapseAgeInputs = (model: Record<string, any> | undefined) => {
+  if (passedAgesKnown(model)) {
+    return false;
+  } else if (
+    passedAgesKnown(model) === false &&
+    passedAgesUnknown(model) === true
+  ) {
+    return true;
+  } else {
+    return false;
+  }
+};
+
 interface AgeGroupGridProps {
   model: Partial<EpidemicModelState>;
   updateModel: (update: EpidemicModelUpdate) => void;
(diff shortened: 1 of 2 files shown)


src/impact-dashboard/FacilityInformation.tsx
```diff
@@ -80,46 +76,86 @@ export const AgeGroupGrid: React.FC<AgeGroupGridProps> = ({
...
+    ) {
+      return true;
+    } else {
+      return false;
+    }
+  };
+
+  useEffect(() => {
+    setCollapsed(collapseAgeInputs(props.model));
+  }, []);
```
glen0071, line 117: @jessex and @macfarlandian I'm not sure if this is the best place for all this logic to live. Happy to move it around.

src/impact-dashboard/FacilityInformation.tsx
```diff
@@ -80,46 +76,86 @@ export const AgeGroupGrid: React.FC<AgeGroupGridProps> = ({
...
+      return true;
+    }
+  };
+
+  const collapseAgeInputs = (model: Record<string, any> | undefined) => {
+    if (passedAgesKnown(model)) {
+      return false;
+    } else if (
+      passedAgesKnown(model) === false &&
+      passedAgesUnknown(model) === true
```
macfarlandian, line 107: I don't know whether it's in scope but I think we might consider having this be a form validation condition as well as a UI display rule? I don't see anything here that actually prevents you from entering stuff in both fields, although we are certainly making it harder to do so

src/impact-dashboard/FacilityInformation.tsx
macfarlandian: seems right to me but I do think you want to add dependencies here so that the modal reacts to changing user data

src/impact-dashboard/FacilityInformation.tsx
glen0071: I'd be happy to add some form validation, although I'm not sure what you have in mind.

I see the value of keeping this PR smaller/simpler and I could do form validation in the next round. Maybe we should write up an issue on form validation? If you you me more guidance I'll try to write it up.

macfarlandian: yeah that sounds fine. the form validation, as I imagine it, would be pretty similar to what you've already done. e.g. valid would be `passedAgesUnknown (xor) passedAgesKnown` ... invalid would be `passedAgesUnknown && passedAgesKnown`. that way we try to actually enforce mutual exclusivity

src/impact-dashboard/FacilityInformation.tsx
glen0071: @macfarlandian adding `props.model` to the dependencies array was easy enough.

Adding that function call caused a rerender every time I clicked a toggle on show/hide age specific inputs. Does that work for you?

macfarlandian: oh yeah good point. obviously `props.model` is the important one here .... 

looking at the code I think you can probably move all of your new function definitions out of the render loop? It doesn't look like they actually have any dependencies on the local state or props of the component. Then you should be able to leave it out of the effect dependencies without triggering a lint warning and we'll be good to go

glen0071: @macfarlandian ok, I think it's ready for another look

PR #570, opened 2020-06-12
Ingest Covid Case Data on a Daily Basis

### LINKED ISSUE #566: [Shadow baseline] Ingest Covid Case Data on a Daily Basis

In order to drive the functionality specified by the Shadow [PRD](https://docs.google.com/document/d/1e6yCfUUYIZW0FlTkzzwYg80-wymZ7PkTClo_k7bqUqk/edit?ts=5ed53a10#heading=h.yug6v3hbadx2), we will need to import covid case data from public sources on a daily basis.

Create a Google Cloud Function within the `c19-backend` Google Project that is responsible for:

1.  Listening for file creation events on a bucket called `c19-backend-covid-case-data`.
2.  Parsing the Daily Covid Case data CSV file and ingesting its contents into our Firestore instance.

Additionally, this issue should include a full set of instructions on how to set up a "[Data Transfer](https://cloud.google.com/storage-transfer-service)" between the mainline Recidiviz Covid Case storage bucket and the Covid-19 `c19-backend-covid-case-data` storage bucket.

sprad:
## Description of the change

Google Cloud Function used to download, reshape, and persist daily Covid case data.  This function expects that a CSV file will be transferred into a specific Google Storage Bucket each day from the mainline Recidiviz Google Cloud project.  Upon arrival, this function will be triggered and will execute.

## Type of change

## Related issues

Closes #566 

### Development

### Code review

### DIFF
--- a/backend/python-functions/data_ingest.py
+++ b/backend/python-functions/data_ingest.py
@@ -52,7 +52,7 @@ def reshape_facilities_data(file_location):
       "Colorado::Denver Reception & Diagnostic Center": {
         "canonicalName": "Denver Reception & Diagnostic Center",
         "facilityType": "State Prisons",
-        "state": "Colorado",
+        "stateName": "Colorado",
         "covidCases": {
           "2020-06-01": {
             "popDeaths": 0,
@@ -84,12 +84,12 @@ def reshape_facilities_data(file_location):
         reader = csv.DictReader(csv_file, delimiter=',')
 
         for row in reader:
-            state = row["State"].strip()
+            state_name = row["State"].strip()
             canonical_facility_name = row["Canonical Facility Name"].strip()
             facility_type = row['Facility Type'].strip()
             date = row['Date'].strip()
 
-            key = f'{state}::{canonical_facility_name}'
+            key = f'{state_name}::{canonical_facility_name}'
             
             if (key in facilities):
                 # If the facility already exists, append the case counts for a given day.
@@ -100,7 +100,7 @@ def reshape_facilities_data(file_location):
                 facilities[key] = {
                     'canonicalName': canonical_facility_name,
                     'facilityType': facility_type,
-                    'state': state,
+                    'stateName': state_name,
                     'covidCases': {
                         f'{date}': build_covid_case_counts(row)
                     }
@@ -114,9 +114,12 @@ def persist(facilities):
     facilities_collection = db.collection(REFERENCE_FACILITIES_COLLECTION_ID)
                 
     for _key, facility in facilities.items():
+        state_name = facility["stateName"]
+        facility_name = facility["canonicalName"]
+
         facilitiesQueryResult = facilities_collection \
-            .where('state', '==', f'{facility["state"]}') \
-            .where('canonicalName', '==', f'{facility["canonicalName"]}') \
+            .where('stateName', '==', f'{state_name}') \
+            .where('canonicalName', '==', f'{facility_name}') \
             .stream()
 
         facilityDocuments = list(facilitiesQueryResult)       
@@ -141,8 +144,7 @@ def persist(facilities):
             
             batch.commit()
         else:
-            log.error(f'Multiple Documents were returned for ' \
-                f'{facility["canonicalName"]} in {facility["state"]}')
+            log.error(f'Multiple Documents were returned for {facility_name} in {state_name}')
 
 def ingest_daily_covid_case_data(bucket_name, file_name):
     file_location = download_covid_case_data(bucket_name, file_name)
--- a/backend/python-functions/data_ingest.py
+++ b/backend/python-functions/data_ingest.py
@@ -1,8 +1,11 @@
 import csv
 from google.cloud import firestore
 from google.cloud import storage
+import logging
 
 REFERENCE_FACILITIES_COLLECTION_ID = 'reference_facilities'
+
+log = logging.getLogger("cloudLogger")
 
 def build_covid_case_counts(row):
     covid_case_counts = {
@@ -84,11 +87,13 @@ def reshape_facilities_data(file_location):
             state = row["State"].strip()
             canonical_facility_name = row["Canonical Facility Name"].strip()
             facility_type = row['Facility Type'].strip()
+            date = row['Date'].strip()
+
             key = f'{state}::{canonical_facility_name}'
             
-            if(key in facilities):
+            if (key in facilities):
                 # If the facility already exists, append the case counts for a given day.
-                facilities[key]['covidCases'][row['Date']] = build_covid_case_counts(row)
+                facilities[key]['covidCases'][date] = build_covid_case_counts(row)
             else:
                 # If the facility does not already exist, save its metadata along with
                 # the case counts for a given day.
@@ -97,7 +102,7 @@ def reshape_facilities_data(file_location):
                     'facilityType': facility_type,
                     'state': state,
                     'covidCases': {
-                        f'{row["Date"]}': build_covid_case_counts(row)
+                        f'{date}': build_covid_case_counts(row)
                     }
                 }
 
@@ -108,7 +113,7 @@ def persist(facilities):
 
     facilities_collection = db.collection(REFERENCE_FACILITIES_COLLECTION_ID)
                 
-    for key, facility in facilities.items():
+    for _key, facility in facilities.items():
         facilitiesQueryResult = facilities_collection \
             .where('state', '==', f'{facility["state"]}') \
             .where('canonicalName', '==', f'{facility["canonicalName"]}') \
@@ -116,7 +121,7 @@ def persist(facilities):
 
         facilityDocuments = list(facilitiesQueryResult)       
 
-        if len(facilityDocuments) == 0 or len(facilityDocuments) == 1:
+        if len(facilityDocuments) <= 1:
             # Accesses an existing reference facility document or creates a new one if
             # one does not exist.
             facility_id = facilityDocuments[0].id if len(facilityDocuments) == 1 else None
@@ -136,7 +141,7 @@ def persist(facilities):
             
             batch.commit()
         else:
-            print(f'Multiple Documents were returned for ' \
+            log.error(f'Multiple Documents were returned for ' \
                 f'{facility["canonicalName"]} in {facility["state"]}')
 
 def ingest_daily_covid_case_data(bucket_name, file_name):
--- a/backend/python-functions/data_ingest.py
+++ b/backend/python-functions/data_ingest.py
@@ -1,0 +1,145 @@
+import csv
+from google.cloud import firestore
+from google.cloud import storage
+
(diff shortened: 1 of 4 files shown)


sprad:
This is the most production Python code I've ever written.  I'm new to the language so feel free to critique heavily as I'm not yet versed in the patterns and conventions of this ecosystem.  Strictly speaking, this function didn't need to be written in Python, but when I first started implementing this work, I thought it was going to live in one of the Cloud Functions in the [pulse-data](https://github.com/Recidiviz/pulse-data/) project and those functions are written in Python.  Since I already had a chunk of working code, I just rolled with it.

This PR is intended to be complete, functional, and mergeable.  That said, the code isn't as robust as it could be.  I'll call attention to some of the soft spots and we can decide whether or not we want to open separate tickets to address those.

sprad:
@jessex - feel free to hit me up early next week on one of my non-Recidiviz days and I can walk you through the steps to set up the [Data Transfer Service](https://cloud.google.com/storage-transfer-service) to move files from the Recidiviz Google Cloud Project's daily Covid case data bucket into the Covid-19 Google Cloud Project's daily Covid case data bucket.

At the moment, the data transfer triggering options are somewhat limited, but they should work for our needs.  It's possible to either set up a one time transfer or a daily transfer scheduled at a specific time.  There's no cron-like sophistication for scheduling, but it is possible to set up more than one transfer for the same buckets.  By doing so, we could have these two cross-project buckets sync a couple times per day if we'd like.

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+            # Accesses an existing reference facility document or creates a new one if
+            # one does not exist.
+            facility_id = facilityDocuments[0].id if len(facilityDocuments) == 1 else None
+            facility_ref = facilities_collection.document(facility_id)
+            
+            # Remove the Covid case data so that it can be stored separately in its own
+            # sub-collection.
+            covid_cases = facility.pop('covidCases')
+            
+            batch = db.batch()
```
sprad, line 129: For each facility, we write its metadata and its daily case counts in a single batch request.  By performing these database writes in a batch, we limit the number of requests we need to send to match the number of facilities within a CSV file and **_not_** the number of daily case rows.  For example, at the moment this means we'll only be performing ~1350 requests for the entire ~50k line CSV file.  Additionally, this means that even as the file grows each day to add new case counts, we won't be generating more requests.  We'll only add more requests as we add more facilities, which should be must less frequent.

sprad: The batch also speeds things up greatly.  Before performing these operations within a batch, the function could not finish executing within the allotted timeout (9 minutes, the max).  With the batch in place, this function completes in about 2.5 minutes.

backend/python-functions/deploy.py
```diff
@@ -13,14 +13,24 @@
...
 
+trigger_options = {
+    "ingest_covid_case_data": [
+        "--trigger-resource", "c19-backend-covid-case-data",
+        "--trigger-event", "google.storage.object.finalize",
+    ],
+    "calculate_rt" : [
+        "--trigger-http",
+    ]
+}
```
sprad, line 24: I'm not sure how generic this deployment script was intended to be, but I've made it less generic with this function-specific trigger configuration.  If this feels egregious, I can work to update this script to take in more parameters to represent what I'm currently capturing in this dictionary.

I've tested the following commands and they work as expected:

```sh
python deploy.py calculate_rt development
python deploy.py ingest_covid_case_data production
```

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+                    'facilityType': facility_type,
+                    'state': state,
+                    'covidCases': {
+                        f'{row["Date"]}': build_covid_case_counts(row)
+                    }
+                }
+
+    return facilities
+
+def persist(facilities):
```
sprad, line 106: For now, the way this works is that the entire Covid case data CSV file will be reshaped and iterated through in its entirety each day.  We are not doing a daily diff / append.  The significance of this is that if we get a file that changes its case counts for a facility for a given day in the past, those case counts in the past will be updated.  I believe this should actually be the desired functionality.  Our Firestore data will reflect the most recent set of truth as we know it.

It may sound like this would cause a lot of unnecessary database updates, but Firestore seems to be smart enough to not write or update documents unless they are new or have changed.

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+    db = firestore.Client()
+
+    facilities_collection = db.collection(REFERENCE_FACILITIES_COLLECTION_ID)
+                
+    for key, facility in facilities.items():
```
sprad, line 111: `key` isn't used, but when I remove it the function fails 🤷

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+            "popTested": 253,
+            "popTestedPositive": 2
+          }
+        }
+      },
+      ...
+    }
+    """
+    facilities = {}    
+    with open(file_location) as csv_file:
```
sprad, line 80: We're also just assuming that whatever file gets plopped into our bucket is a CSV file in the correct format.  In practice, this should be the case, but it is a naive assumption and some better file validation is likely worthwhile.

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for row in reader:
+            state = row["State"].strip()
+            canonical_facility_name = row["Canonical Facility Name"].strip()
+            facility_type = row['Facility Type'].strip()
+            key = f'{state}::{canonical_facility_name}'
+            
+            if(key in facilities):
+                # If the facility already exists, append the case counts for a given day.
+                facilities[key]['covidCases'][row['Date']] = build_covid_case_counts(row)
```
jessex, line 91: I would pull up the indexing on date to the top of the loop alongside `state` and `facility_type`, e.g. `date = row['Date']`. Then just reference that in both places in this function. It will be easier to read in both places it's being used right now, in my opinion.

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+                
+    for key, facility in facilities.items():
+        facilitiesQueryResult = facilities_collection \
+            .where('state', '==', f'{facility["state"]}') \
+            .where('canonicalName', '==', f'{facility["canonicalName"]}') \
+            .stream()
+
+        facilityDocuments = list(facilitiesQueryResult)       
+
+        if len(facilityDocuments) == 0 or len(facilityDocuments) == 1:
```
jessex, line 119: `if len(facilityDocuments) <=1`

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+            # Accesses an existing reference facility document or creates a new one if
+            # one does not exist.
+            facility_id = facilityDocuments[0].id if len(facilityDocuments) == 1 else None
+            facility_ref = facilities_collection.document(facility_id)
```
jessex, line 123: What happens here if `facility_id` is None?

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+            
+            batch.set(facility_ref, facility)    
+
+            for date, cases in covid_cases.items():
+                covidCasesOnDateRef = facility_ref.collection('covidCases').document(date)
+                batch.set(covidCasesOnDateRef, cases)
+            
+            batch.commit()
+        else:
+            print(f'Multiple Documents were returned for ' \
```
jessex, line 139: Is this indicative of an error we should be raising? I imagine we'd still want to process everything else so this is fine, but want to double-check.

backend/python-functions/data_ingest.py
jessex: Fantastic!

backend/python-functions/data_ingest.py
sprad: I dig it 👍

backend/python-functions/data_ingest.py
sprad: 🤦

backend/python-functions/data_ingest.py
sprad: If `facility_id` is none then we'll create a new facility document.  Essentially lines 122 and 123 determine whether or not we're performing a create or an update.

backend/python-functions/data_ingest.py
sprad: Yes, this represents a bad data case since at the moment we're forced to key off of the combination of the facility's state and name.  In other words, within a state, all facility names must be unique.  If this doesn't hold true for some reason, we wouldn't know which facility we should be updating so instead of guessing, we just log this message.

I've asked the data team to generate globally unique id's for each of the facilities so that we can avoid this and other more serious data issues.  [This appears to be in the works](https://docs.google.com/spreadsheets/d/1v50osSKG-vwmIhjIn6lFKDeY0ARsUT2R4olKsYsba2o/edit?disco=AAAAGnllJaU).

macfarlandian: I would recommend that you use a logger here instead of printing; that way you can give it ERROR status in the logs to make it easier to identify and respond to. (it's super simple, you can see an example in `helpers.py` if you are not familiar with this module in python)

sprad: Great call! 🙌 

I'll update this `print` statement to use a logger.  Thanks for helping me by pointing out an example!

backend/python-functions/data_ingest.py
```diff
@@ -0,0 +1,147 @@
...
+    2020-06-03,State Prisons,Colorado,Denver Reception & Diagnostic Center,253,2,,0,,,,,,,
+    2020-06-04,State Prisons,Colorado,Denver Reception & Diagnostic Center,253,2,,0,,,,,,,
+
+    Would be reshaped to:
+
+    {
+      "Colorado::Denver Reception & Diagnostic Center": {
+        "canonicalName": "Denver Reception & Diagnostic Center",
+        "facilityType": "State Prisons",
+        "state": "Colorado",
```
macfarlandian, line 52: some minor housekeeping but could we call this field `stateName` instead? trying to standardize references to US states versus application state per #584

sprad: Will do.  Are we also intending to update the document schema for the reference facilities in Firestore to use `stateName` instead of `state`?

sprad: Sorry, let me re-phrase.  If we make this change, this will set the document field to `stateName` in the `reference_facilities` collection's documents.  Is that cool?

macfarlandian: we will have to make some minor code changes probably but yes

sprad: Got it.  Will make the update

PR #571, opened 2020-06-15
Merlin UI polish: Prevent y-axis from having dupe ticks

amyrlam:
## Description of the change

Bug fix: If generated tick values = 2 and 2.5, ticks now display as 2 and 2.5, instead of previous duplicated 2 and 2

//

WAS (When the y-axis max = 3, the ticks were duplicating e.g. y-axis ticks were: 1, 1, 2, 2, 3, 3):
image

IS:
image

image

~~👉 Question for @macfarlandian (saw you did a lot of the graph code) or others if you have time lmk here or Slack!~~ 

~~I'm not sure if there is a better way to do this e.g. if y-axis max <= 6, use different `timeFormat` in CurveChart. But I don't see a way to access that in https://github.com/Recidiviz/covid19-dashboard/blob/master/src/rt-timeseries/RtTimeseries.tsx e.g. no max on the yExtent. Any ideas?~~ 

## Type of change

## Related issues

Closes https://github.com/Recidiviz/covid19-dashboard/issues/372, is last issue in [Google doc](https://docs.google.com/document/d/15oREP0hXoQSnzL-mMp3fTYRLyomyhskLD38-l2t1v_E/edit) 

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/CurveChart.tsx
+++ b/src/impact-dashboard/CurveChart.tsx
@@ -35,7 +35,10 @@ const TooltipDatum = styled.li`
   opacity: 0.8;
 `;
 
-const formatThousands = format(",.0f");
+// Note:
+// if tick value = 1,001.00, displays 1001
+// if tick value = 2.5, displays 2.5
+const formatThousands = format(",~g");
 
 interface TooltipProps {
   countTotal: number;


macfarlandian:
the underlying mechanism for determining how many ticks to show and what their values should be is kind of a black box, so I actually think keying this behavior off the max extent is going to be too brittle anyway. I would recommend passing a formatting function that detects fractional values and formats them differently than whole numbers (that way you don't wind up appending `.0` to everything, which would be distracting in like 99% of charts because fractional values are nonsensical here)

amyrlam:
@macfarlandian thanks, that sounds like a good new path 🙏

Will try to have a conditional tickFormat, when the tick is not a `3.0`. Will report back

amyrlam:
@macfarlandian Turns out I was able to solve it in a "D3 way" but your comment was extremely helpful in thinking about the prob differently, thanks again!

Issue #556, opened 2020-06-11
[Shadow sync config] Let users match existing facilities to shadow facilities


sychang:
**Is your feature request related to a problem? Please describe.**
This is a user-facing component of the shadow baseline project to pre-populate user scenarios with publicly collected data to reduce their data entry workload so we can improve retention and reduce user friction.

Once we launch the feature, for users that already have facilities in their model, we need them to do a one-time mapping of their facilities to our shadow data so we can sync the data moving forward.

**Describe the solution you'd like**
Users who already have facilities in the model will create a 1-1 mapping of our shadow baseline scenario (SBS) facilities to their user-entered baseline scenario (BS) facilities. 

In steady state, users can access this modal from a button in the left column according to the [trigger logic](https://docs.google.com/document/d/1e6yCfUUYIZW0FlTkzzwYg80-wymZ7PkTClo_k7bqUqk/edit?ts=5ed53a10#heading=h.efwt1m8okoy2) in the PRD:

![image](https://user-images.githubusercontent.com/6633887/85311745-81ce9d00-b47b-11ea-8f0c-c561fc213d1f.png)

For the modal:
- Key the sync config content based on the "State" and "Type of System" information from the first facility in the user's model (there will be a separate issue to only trigger this feature when state and type of system match across all facilities).
- For SBS facility, users have the option to select an existing BS facility, to "Add as a new facility" (if the facility is not currently modeled), or "Don't sync this facility" if they do not want the SBS facility added to their model.
- A green sync icon will appear next to existing facilities and the “Add as new facility” option, the green sync icon will appear in the input box
- This process should not change the name of their facility
- Save should be disabled until the user makes at least one selection in the dropdown
- A toggle should allow them to turn off the preopulated data entirely. If they were to turn it off and on again, we should remember which facilities they already had synced.

![image](https://user-images.githubusercontent.com/6633887/85311810-9874f400-b47b-11ea-903e-3ab03a558844.png)

![image](https://user-images.githubusercontent.com/6633887/85311969-d114cd80-b47b-11ea-9e3d-83ccaf2bc5cf.png)

When we first launch the feature, this modal will be a promo with a few minor tweaks:
1. Remove the toggle
2. Replace "Cancel" with "Not Now"

![image](https://user-images.githubusercontent.com/6633887/85311855-aaef2d80-b47b-11ea-9a98-beea42ac3354.png)

**Describe alternatives you've considered**
N/A

**Additional context**
[...]

PR #575, opened 2020-06-16
Add reference facility mapping modal when saving new facility

### LINKED ISSUE #559: [Shadow, sync config] When creating a new facility, pick a shadow baseline facility to sync it to
This is a user-facing component of the shadow baseline project to pre-populate user scenarios with publicly collected data to reduce their data entry workload so we can improve retention and reduce user friction.

During steady state, users who have their baseline scenario (BS) synced with the public shadow baseline scenario (SBS) may choose to model a new facility. We want to make sure new facilities created after sync is turned on can be synced to the latest public data.

If sync to public data is turned on, when adding a new facility, on-save, we prompt the user to assign an SBS facility (opens the Sync Config dialog):
image

All facilities begin "unselected" or grayed out. Only one facility may be selected at once.

image

N/A

**Additional context**
Full details in [PRD](https://docs.google.com/document/d/1e6yCfUUYIZW0FlTkzzwYg80-wymZ7PkTClo_k7bqUqk/edit?ts=5ed53a10#)

Data model [design document](https://docs.google.com/document/d/1mvuQpXMPf3V164D4B7waCRArONw9iOWvE1g2SsLh03A/edit?usp=sharing)

Mocks: https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=2758%3A0
TODO: Finalize copy

Backend issue: #521
Related front-end issues: #556, #557, #558, trigger logic issue incoming

daschi:
## Description of the change

This PR adds a new component: `SyncReferenceFacilityModal` 

When a user saves a *new* facility on their baseline scenario and the `enableShadowData` flag is on, they will see this modal and they can either select a reference facility to sync or continue saving the facility without syncing. 

TODO: 

@sokunthel and @macfarlandian I haven't yet added a mapping object like `{ [facilityId]: referenceFacilityId }`, which would need to be added on line 126 in the `SyncReferenceFacilityModal`, because I wasn't sure if something similar was already being done in #556 or elsewhere. I'm wondering if one or both of you would be available to chat about how we want to save the mapping. Also, I'm assuming that this issue doesn't include displaying the "shadow" data on the multi-facility page. Thanks for your help!

## Type of change

## Related issues

Closes #559 

### Development

### Code review


src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -131,10 +133,15 @@ const PageHeaderContainer = styled.div`
...
+  scenario: Scenario;
 }
 
-const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
+const FacilityInputForm: React.FC<Props> = ({ scenario }) => {
+  const scenarioId = scenario.id;
+  const [syncDataModalFacility, setSyncDataModalFacility] = useState<
+    string | null
+  >(null);
+  const useReferenceFacilities = useFlag(["enableShadowData"]);
```
macfarlandian, line 144: there is a hook called `useReferenceFacilitiesEligible` that you probably want to use here instead; there are other criteria besides the feature flag that determine whether the feature should be available.

src/page-multi-facility/SyncReferenceFacilityModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+      setSelectedRefFacility(null);
+    }
+  }
+
+  function handleSave() {
+    if (selectedRefFacility) {
+      deselectFacility();
+      // save a synced composite facility?
+      console.log({ facility, selectedRefFacility });
```
macfarlandian, line 127: I think whatever happens with the saved facility data will be handled within the Facility context; you shouldn't have to interact with that mechanism directly here. All you should have to do, as you said, is update the mapping. (that should just be a new property of the scenario, both you and @sokunthel will need it for your features but it's a simple object (as described in [this section of the design doc](https://docs.google.com/document/d/1mvuQpXMPf3V164D4B7waCRArONw9iOWvE1g2SsLh03A/edit#heading=h.3ez6its1yt01)) so it shouldn't cause major merge conflicts or anything

src/page-multi-facility/SyncReferenceFacilityModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+            referenceFacility={refFacility}
+            onClick={() => handleClick(refFacility)}
+          />
+        ))}
+      </ModalContent>
+      <ModalFooter>
+        <CancelButton
+          onClick={() => {
+            deselectFacility();
+            navigate("/");
```
macfarlandian, line 148: these two lines are repeated several times across files, they should probably be in some kind of "exit this page" function that can be used by the parent component and passed in here as an "onClose" prop or some such

src/page-multi-facility/SyncReferenceFacilityModal.tsx
macfarlandian: oh actually one thing I would add about this -- at some point we are going to have to cut the database over from the test collection to the real collection, which will invalidate any of these mappings we have saved in the meantime. I was thinking we might also want to use a "test" version of the mapping so that we can just change the key at the same time as we change the collection ID and give everyone a fresh mapping. (we can discuss more if that needs clarification)

daschi: Oh okay, do you mean add a test property that we can then drop from the scenario document? Something like:
```
scenario: {
  ...other scenario properties...
  testReferenceFacilityMapping: {
    [facilityId]: referencefacilityId
  }
}
```

macfarlandian: yeah exactly

src/database/index.tsx
```diff
@@ -44,6 +44,8 @@ const usersCollectionId = "users";
 // TODO (#521): when the datasource stabilizes, change this to the real collection
 const referenceFacilitiesCollectionId = "reference_facilities_test";
 const referenceFacilitiesCovidCasesCollectionId = "covidCases";
+// TODO (#559): when we switch to using the real data update this property for the scenario
```
macfarlandian, line 47: not that it hugely matters but i would tag this to #521, as in the comment above

src/database/index.tsx
```diff
@@ -84,17 +86,17 @@ const authenticate = async () => {
   await firebase.auth().signInWithCustomToken(customToken);
 };
 
-const currrentUserId = () => {
+const currentUserId = () => {
```
macfarlandian, line 89: 🙌

src/page-multi-facility/SyncReferenceFacilityModal.tsx
```diff
@@ -0,0 +1,175 @@
+import { navigate } from "gatsby";
```
macfarlandian, line 1: some linting issues in this file

PR #593, opened 2020-06-17
Prevent the App from Continuously Creating Multiple User Documents

### LINKED ISSUE #592: Prevent the App from Continuously Creating Multiple User Documents
**Describe the bug**

There are two separate but related issues in play.  This ticket will discuss them both, but will only attempt to resolve the second.

**First Issue**

There is some condition under which multiple `user` documents will get created for the same user (i.e. all of the document's attributes will be identical, including `auth0Id`).  It's not completely clear what is causing this, but it seems to have something to do with users using the app while also experiencing a poor Internet connection.

So far, this has fortunately only affected two internal users:  Jen and Katrina.  After speaking with Jen, she mentioned the app was displaying odd behavior last while she was having Internet issues.

Katrina offered a similar account and was even able to capture a very suspicious and potentially related error in her browser:

Screenshot 2020-06-09 at 11 18 28
 
Depending on when this issue arises, we may end up creating a separate `user` document if [this code executes](https://github.com/Recidiviz/covid19-dashboard/blob/533901b4a27ef0268e0c87cfceff6589624e7c0e/src/auth/react-auth0-spa.tsx#L90-L99).

**Second Issue**

Once the first issue occurs and a second identical `user` document is created, every time the user logs into or refreshes the app, another `user` document will be created.  This is because:

1.  The call to `getUserDocument` in [getUser](https://github.com/Recidiviz/covid19-dashboard/blob/533901b4a27ef0268e0c87cfceff6589624e7c0e/src/database/index.tsx#L432)...
2.  Will [throw an exception](https://github.com/Recidiviz/covid19-dashboard/blob/533901b4a27ef0268e0c87cfceff6589624e7c0e/src/database/index.tsx#L421-L422) when there are multiple `user` documents for the same `auth0Id`...
3.  Causing the `getUser` function to [return null](https://github.com/Recidiviz/covid19-dashboard/blob/533901b4a27ef0268e0c87cfceff6589624e7c0e/src/database/index.tsx#L434-L436)...
4.  Which causes our authentication code to [create yet another](https://github.com/Recidiviz/covid19-dashboard/blob/533901b4a27ef0268e0c87cfceff6589624e7c0e/src/auth/react-auth0-spa.tsx#L90-L99) `user` document

At the time of writing, there were over 100 errant and duplicated `user` documents for these two users.

**To Reproduce**

I don't know how to reproduce the first issue, but the second issue can be reproduced by manually adding a new `user` document in Firestore with identical attributes to another `user` document.

Upon doing this, the app will continue to appear to function, but under the hood, errors will be thrown and bad data will continue to be generated.

**Expected behavior**

For now, there's not really a graceful way to handle the situation where the database is in a bad state.  As such, we'll just have the app error out.  When this happens, we can clean the data and the app will function correctly again.

sprad:
## Description of the change

See the write up in the [accompanying issue](https://github.com/Recidiviz/covid19-dashboard/issues/592) for full context and a description of this bug.

## Type of change

## Related issues

Closes #592 

### Development

### Code review

### DIFF
--- a/src/database/index.tsx
+++ b/src/database/index.tsx
@@ -420,21 +420,17 @@ const getUserDocument = async (
 
   if (userResult.docs.length > 1) {
     throw new Error(`Multiple users found matching ${key} ${value}`);
-  } else if (userResult.docs.length === 0) {
-    throw new Error(`No users found matching ${key} ${value}`);
   }
 
   return userResult.docs[0];
 };
 
 export const getUser = async (auth0Id: string): Promise<User | null> => {
-  try {
-    const userDocument = await getUserDocument({ auth0Id });
-    return buildUser(userDocument);
-  } catch (e) {
-    console.error("Encountered error while attempting to retrieve user: \n", e);
-    return null;
-  }
+  const userDocument = await getUserDocument({ auth0Id });
+  // It is ok for this method to return null if we do not find a user document
+  // for the given auth0Id.  This will always be the case when a user logs in
+  // for the first time.
+  return userDocument ? buildUser(userDocument) : null;
 };
 
 export const getScenarioUsers = async (


src/database/index.tsx
```diff
@@ -420,21 +420,17 @@ const getUserDocument = async (
 
   if (userResult.docs.length > 1) {
     throw new Error(`Multiple users found matching ${key} ${value}`);
-  } else if (userResult.docs.length === 0) {
-    throw new Error(`No users found matching ${key} ${value}`);
```
sprad, line 424: This isn't actually an error case so we shouldn't throw an error here

PR #603, opened 2020-06-19
Update default information for "Add Historical Data" Modal

### LINKED ISSUE #523: Update 'Add historical data' modal defaults
The modal that appears when a user clicks 'Add historical data' auto-completes the case and population numbers to the most recent date's data for which there is user input.
- This works well when the user is entering new data for a more recent date than that last input
- This does not work well when the user goes to enter historical data prior to their earlier entered data so far.

For example, let's say the user started entering data on 6/1, and has a (cases, population) of (40, 200), (50, 200), and (60, 200) for 6/1-6/3.
- If the user goes to enter 6/4, the defaults for the new entry are (60, 200) which is intuitive - these are likely very close to the new data.
- If the user goes to enter earlier data, however (e.g., 5/31), they _also_ get default values of (60, 200) - even though we should know that the counts have to be at or under the cases and population for 6/1 (40, 200).

This is confusing for users who go to enter historical data, because they're focused on just the values they think changed - so they might update Staff Cases, thinking they can leave Resident Cases alone, because it didn't change between this older date and the date following it.

We should update the dialog default values in cases where the user is entering new data prior to the earliest user-entered data, or between two sets of previously entered data. The rules should be:
- If there is data from a date prior to the date the user is entering information for, the defaults should be for that last date prior for which we have data.
- If there is _not_ data for a date prior to the date the user is entering information for, we should default values to the values of the _next_ forward-looking date for which we have data
- If there is no data backward-looking or forward-looking, there are no defaults / the cells are left blank.  

None

**Additional context**
None

neeharmb:
## Update default information for "Add Historical Data" Modal

> Add existing data to "add historical data" modal inputs depending on if backward-or forward-looking information already exists for the selected date. The specific behavior is detailed in the corresponding issue.

## Type of change

## Related issues

Closes #523

### Development

Testing with scenario in #523:
 
Adding old data from 5/31:
adding old data

Adding new data from 6/4:
adding new data

Adding to facility without existing data:
no data

### Code review

### DIFF
--- a/src/hooks/useAddCasesInputs.tsx
+++ b/src/hooks/useAddCasesInputs.tsx
@@ -18,26 +18,24 @@ export const getBracketData = (modelInputs: ModelInputs) => {
 };
 
 function findMostRecentDate(
-  defaultDate: Date,
+  currentDate: Date,
   facilityModelVersions: ModelInputs[] | undefined,
 ) {
-  let mostRecentDate = defaultDate;
+  let mostRecentDate = currentDate;
   if (facilityModelVersions) {
     // create array of dates with observed data at a given facility
-    let facilityDatesObservedAt = facilityModelVersions.map(
+    const facilityDatesObservedAt = facilityModelVersions.map(
       (observedAt) => observedAt.observedAt,
     );
     if (facilityDatesObservedAt.length > 0) {
-      // filter to dates earlier than (or the same as) the observed date
-      let earlierDates = facilityDatesObservedAt.filter(function (d) {
-        return d.valueOf() <= defaultDate.valueOf();
+      // filter to dates earlier than the current date
+      const earlierDates = facilityDatesObservedAt.filter(function (d) {
+        return d.valueOf() <= currentDate.valueOf();
       });
-      // if there is data for a prior date, use it
+      // if there is data for a prior date, use it, otherwise use the current date
       if (earlierDates.length > 0) {
         mostRecentDate = earlierDates[earlierDates.length - 1];
-      }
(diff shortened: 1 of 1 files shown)


src/hooks/useAddCasesInputs.tsx
```diff
@@ -17,6 +17,31 @@ export const getBracketData = (modelInputs: ModelInputs) => {
...
+      (facility) => facility.observedAt,
+    );
+    // filter to dates earlier than (or the same as) the current date
+    const earlierDates = facilityObservedAtDates?.filter(function (date) {
+      return startOfDay(date) <= startOfDay(observedAtDate);
+    });
+    // if there is data for prior dates, use the most recent one, otherwise use
+    // the next forward-looking date that we have for the facility
+    if (earlierDates && earlierDates.length > 0) {
+      mostRecentDate = earlierDates[earlierDates.length - 1];
```
jessex, line 37: The indexing logic in this conditional block assumes that `facilityModelVersions` has already been sorted by `observedAt`, right? Just making sure that that's a safe assumption?

neeharmb: @jessex from my local testing it looks like it's already sorted, but I'll add an explicit sort to make it more robust

jessex: @macfarlandian to confirm that's a good call and not superfluous. Ian, do you think an extra sort here is called for?

macfarlandian: it should generally be a safe assumption that the dates are already sorted, so I don't think it's really necessary here

PR #606, opened 2020-06-19
Update the Data Model to Support Determining Whether or not a Referen…

### LINKED ISSUE #604: [Shadow, data] Update the Data Model to Support Determining Whether or not a Reference Facility is "New"

In order to support #560 - "When a new data shadow facility is found, enable mapping to existing facility", we will need to keep track a few additional pieces of state to determine whether or not a Reference Facility is "new" from the perspective of a given user.

The proposal will be to update the schema for the `reference_facilities` documents to include a new `createdAt` field.  This field will be set to a timestamp of the time when the document is initially created.  This will require updating the existing Python cloud function to set this new piece of data.  That is the work to be done for this issue.

We will also need to add a new timestamp field to the Baseline `scenario` document called `lastReferenceObservedAt`.  This field will be set each time the user interacts with the Sync Config modal.  We should be able to add and set this field using a "lazy migration" as we've done in the past.

Upon logging into the app, we will need to run the following logic:

1.  Find all of the reference facilities that have a `createdAt` timestamp that is greater than the user's baseline scenario's `lastReferenceObservedAt`.
2.  If any reference facilities are returned, prompt the user to show them the new reference facilities as outlined in #560.

sprad:
…ce Facility is "New"

## Description of the change

- When importing reference facilities from our CSV data source we will need to set a `createdAt` timestamp on each new facility.  This will allow us to know which reference facilities are "new" from the perspective of any given user.  This "newness" determination will be made by comparing a timestamp that the user last observed the reference facility sync modal and this new `createdAt` field.  If the `createdAt` field is greater than the time the user last observed the sync config modal, we know that the reference facility is "new" to that user.  This will be important in driving certain user-facing features.

## Type of change

## Related issues

Closes #604 

### Development

### Code review

### DIFF
--- a/backend/python-functions/data_ingest.py
+++ b/backend/python-functions/data_ingest.py
@@ -125,18 +125,27 @@ def persist(facilities):
         facilityDocuments = list(facilitiesQueryResult)       
 
         if len(facilityDocuments) <= 1:
-            # Accesses an existing reference facility document or creates a new one if
-            # one does not exist.
-            facility_id = facilityDocuments[0].id if len(facilityDocuments) == 1 else None
+            # If a reference facility already exists we will set facility_id to the existing
+            # reference facility's id in order to perform an update on that reference
+            # facility.  Otherwise, we will set the facility_id to None which signals to
+            # Firestore to create a new reference facility document.
+            existing_reference_facility = len(facilityDocuments) == 1
+            facility_id = facilityDocuments[0].id if existing_reference_facility else None
             facility_ref = facilities_collection.document(facility_id)
-            
+
+            # For new reference facility documents, set a createdAt timestamp.  This allows
+            # us to know which reference facilities are "new" from the perspective of a given
+            # user.
+            if (not existing_reference_facility):
+                facility['createdAt'] = firestore.SERVER_TIMESTAMP
+
             # Remove the Covid case data so that it can be stored separately in its own
             # sub-collection.
             covid_cases = facility.pop('covidCases')
             
             batch = db.batch()
-            
-            batch.set(facility_ref, facility)    
+
+            batch.set(facility_ref, facility, merge=True)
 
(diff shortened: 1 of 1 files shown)


backend/python-functions/data_ingest.py
```diff
@@ -125,18 +125,27 @@ def persist(facilities):
...
+
             # Remove the Covid case data so that it can be stored separately in its own
             # sub-collection.
             covid_cases = facility.pop('covidCases')
             
             batch = db.batch()
-            
-            batch.set(facility_ref, facility)    
+
+            batch.set(facility_ref, facility, merge=True)
```
sprad, line 148: I needed to add this `merge` flag or else the `createdAt` field was getting removed for existing reference facility documents upon subsequent imports

daschi: Since we're currently working with test data, can we expect that all reference facility docs will have `createdAt` once this switches over to use the real imported data?

sprad: Yup, all of the `reference_facilities` documents have already been updated to include a `createdAt` field.  I updated the test data to include the `createdAt` field as well.

backend/python-functions/data_ingest.py
```diff
@@ -125,18 +125,27 @@ def persist(facilities):
...
+            # Firestore to create a new reference facility document.
+            existing_reference_facility = len(facilityDocuments) == 1
+            facility_id = facilityDocuments[0].id if existing_reference_facility else None
             facility_ref = facilities_collection.document(facility_id)
-            
+
+            # For new reference facility documents, set a createdAt timestamp.  This allows
+            # us to know which reference facilities are "new" from the perspective of a given
+            # user.
+            if (not existing_reference_facility):
```
jessex, line 139: Nit: These parentheses should be unnecessary.

PR #609, opened 2020-06-24
Prevent duplicate/conflicting age data on case entry

### LINKED ISSUE #572: User can submit known and unknown age data at same time
**Describe the bug**
Users can possibly submit case/population data with conflicting known and unknown ages.

**To Reproduce**
1. In Add Cases Modal or the Facility Details page, 
2. enter cases/population data in the unknown age input boxes, 
3. then toggle to known aged inputs and enter more data. 
4. Click submit/save and it succeeds

**Expected behavior**
Submit should fail and raise Toast message "To prevent duplicate counting, known and unknown ages cannot be entered together. Please clear either the unknown age input or all age inputs."?

**Additional context**
See @macfarlandian suggestion on [PR 568](https://github.com/Recidiviz/covid19-dashboard/pull/568)

glen0071:
## Description of the change

Prevent entry of duplicated age data (known ages and unknown ages) submitted at the same time.

If a user enters more than 0 in both known and unknown age inputs in a form, the save is canceled and an error message appears. Error copy: "To prevent duplicate counting, known and unknown ages cannot be entered together. Please clear either the unknown age input or all age inputs."

Gif demos:
facility-demo

modal-demo

## Type of change

## Related issues

Closes #572 

### Development

### Code review

### DIFF
--- a/src/impact-dashboard/FacilityInformation.tsx
+++ b/src/impact-dashboard/FacilityInformation.tsx
@@ -1,9 +1,11 @@
+import { isUndefined, omitBy, pickBy } from "lodash";
 import numeral from "numeral";
 import React, { useEffect, useState } from "react";
 import styled from "styled-components";
 
 import InputTextNumeric from "../design-system/InputTextNumeric";
 import TextLabel from "../design-system/TextLabel";
+import { useToasts } from "../design-system/Toast";
 import {
   curveInputsFromUserInputs,
   getAdjustedTotalPopulation,
@@ -87,10 +89,35 @@ const pastAgesUnknown = (model: Record<string, any> | undefined) => {
   }
 };
 
+const includesKnownAges = (model: object) => {
+  let definedInputs = omitBy(model, isUndefined);
+  let ageKnownInputs = pickBy(definedInputs, (_, key) => {
+    return RegExp(/age\d+/).test(key);
+  });
+  let total = Object.values(ageKnownInputs).reduce((sum, n) => {
+    return sum + n;
+  }, 0);
+  return total > 0;
+};
+
+const includesUnknownAges = (model: object) => {
+  let definedInputs = omitBy(model, isUndefined);
+  let ageUnknownInputs = pickBy(definedInputs, (_, key) => {
+    return RegExp(/ageUnknown\w+/).test(key);
+  });
+  let array = Object.values(ageUnknownInputs);
+  let total = array.reduce((sum, n) => {
+    return sum + n;
+  }, 0);
+  return total > 0;
+};
+
 interface AgeGroupGridProps {
   model: Partial<EpidemicModelState>;
   updateModel: (update: EpidemicModelUpdate) => void;
   collapsible?: boolean;
+  warnedAt: number;
+  setWarnedAt: (warnedAt: number) => void;
 }
 
 export const AgeGroupGrid: React.FC<AgeGroupGridProps> = ({
@@ -114,6 +141,8 @@ export const AgeGroupGrid: React.FC<AgeGroupGridProps> = ({
 
   useEffect(() => {
     setCollapsed(collapseAgeInputs());
+    // only want to run this once, on initial mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   const ageSpecificCaseCounts = (
@@ -217,11 +246,14 @@ interface AgeGroupRowProps {
   rightKey: keyof EpidemicModelUpdate;
   model: Partial<EpidemicModelState>;
   updateModel: (update: EpidemicModelUpdate) => void;
+  warnedAt: number;
+  setWarnedAt: (warnedAt: number) => void;
 }
 
 const AgeGroupRow: React.FC<AgeGroupRowProps> = (props) => {
   const { model, updateModel } = props;
   const [inputRelativityError, setInputRelativityError] = useState(false);
+  const { addToast } = useToasts();
 
   function checkInputRelativity(
     cases: number | undefined,
@@ -238,6 +270,18 @@ const AgeGroupRow: React.FC<AgeGroupRowProps> = (props) => {
     }
   }
 
+  const checkAgeConflict = (model: object) => {
+    if (Date.now() > props.warnedAt + 10000) {
+      if (includesUnknownAges(model) && includesKnownAges(model)) {
+        addToast(
+          "To prevent duplicate counting, known and unknown ages cannot be entered together. Please clear either the unknown age input or all age inputs.",
+          { appearance: "error" },
+        );
+        props.setWarnedAt(Date.now());
+      }
+    }
+  };
+
   return (
     <FormGridRow>
       <LabelCell>
@@ -251,6 +295,7 @@ const AgeGroupRow: React.FC<AgeGroupRowProps> = (props) => {
           onValueChange={(value) => {
             checkInputRelativity(value, model[props.rightKey] as number);
             updateModel({ [props.leftKey]: value });
+            checkAgeConflict(model);
           }}
         />
       </InputCell>
@@ -261,6 +306,7 @@ const AgeGroupRow: React.FC<AgeGroupRowProps> = (props) => {
           onValueChange={(value) => {
             checkInputRelativity(model[props.leftKey] as number, value);
             updateModel({ [props.rightKey]: value });
+            checkAgeConflict(model);
           }}
         />
       </InputCell>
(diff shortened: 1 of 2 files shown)


src/page-multi-facility/FacilityInputForm.tsx
```diff
@@ -172,6 +172,29 @@ const FacilityInputForm: React.FC<Props> = ({ scenarioId }) => {
...
+    let definedInputs = omitBy(inputs, isUndefined);
+    let ageUnknownInputs = pickBy(definedInputs, (_, key) => {
+      return RegExp(/ageUnknown\w+/).test(key);
+    });
+    let array = Object.values(ageUnknownInputs);
+    let total = array.reduce((sum, n) => {
+      return sum + n;
+    }, 0);
+    return total > 0;
+  };
```
macfarlandian, line 196: since this is repeated more or less verbatim (along with the toast text) I think you should factor it out into a function that you can import into both of these files ... maybe it could throw an error that you could catch in order to interrupt the save

sychang:
Disclaimer: Read and responded to this very quickly, so let me know if I'm not addressing all the points or suggesting something totally egregious, will take a more in-depth look later.

**Should you be able to enter known age cases and unknown age population, or vice versa? In the current implementation, it's all one or the other, but I wonder if some facility admins might, for example, know the ages of confirmed cases but not of the population at large.**
- Yes, that would ideally be possible. This situation may also come up in the opposite manner when we have shadow data, where we do not know the age distribution of cases, but a user may have broken the facility population down by age.

On the rest, I wonder if a couple of UX fixes could resolve all the above 3 bullets:
1. Make all relevant permanently red until the error is fixed (not just the second input)
2. Instead of the error toast appearing on save after the modal closes, is it possible for the toast to appear when the input turns red while the modal is still open, and the user cannot click save until they fix the issue? This way, there's more explanation for what the issue is and also prevents them from losing their data

glen0071:
@jessex I'm not sure which red coloring you see related to known vs unknown age data conflicts.

On an earlier PR I wrote some code that highlights the left (cases) inputs with red if that number exceeds the total population number on the right. That should pop up regardless of whether you're looking at known or unknown cases. 

Is this what you're looking at?
Screen Shot 2020-06-25 at 10 34 31 AM
If that error needs more context, I could add a toast message or something on that. Or if it's just not helpful, I should remove it?

If we are more concerned with duplicated/conflicting data in a single form, I now understand that this might not necessarily be a mistake by the user, so maybe we don't want to suggest this action is an error? Maybe red highlights and error messages should be yellow? Or the toast should have "info" or "success" coloring? 

Regardless of color, it may be more helpful to the user if the toast message (or whatever warning we use) pops up as soon as the conflicting data is entered - rather than waiting for them to press "save". I think that would probably be faster to implement as well.

So the UX would be,
Given a user is filling in both _known_ and _unknown_ case data into a single form, 
When they enter data into a second (possibly conflicting) category of age data
Then a toast message pops up with the copy @sychang provided before. I think that would be sufficient.

Additionally, the toast could be combined with the input fields changing color as well - although turning them red might be confusing/annoying if they should still feel free to enter the data if they're sure it's correct.
And I don't think past data entries would matter at all in this scenario. The messages are only raised for conflicting entries in the current form.

PS: I worry that I, as a volunteer, might be pushing too many ideas on your work, so no hard feelings if you want me to stop raising questions and just implement what you have already learned from the users.

macfarlandian:
At the risk of further derailing this PR I did want to mention that the underlying calculation models do not really support the "ideal" outcome that @sychang describes above at the moment; each population bracket is treated as a discrete group when calculating the projection, so having, e.g., all your cases be "age unknown" while your population is all in age brackets will result in zero output. In other words, cases and population are considered together, bracket by bracket. 

So I think we should either be encouraging people to properly enter data in way that conforms to this requirement (which is mainly what this validation behavior was about originally, I think?) or, if we are actually intending to change that underlying requirement, probably not be too heavy-handed with validation in the meantime

jessex:
Ah sorry about that, @glen0071, you're right: the red coloring I was talking about was because the case counts were higher than the population counts, not that there was a mix of known and unknown inputs! I assumed it was part of your current work because I misread a comment on the original issue in which I thought you were doing _both a toast message and red highlighting_. I see now that that is not the case on the original issue, and I don't think it needs to be the case in the implementation.

So I think your proposed UX sounds right: just display a toast message when the issue appears without closing the modal / navigating away from the page. Just be careful to avoid having that toast message, for example, blink multiple times in a row with each new keystroke that is itself violating. But yes, just continue to skip the red entirely.

@macfarlandian: do you have any concerns about what you're suggesting (continuing to disallow a mix of known cases and unknown population or vice versa, as currently implemented in this PR) with the introduction of shadow data?

And P.S. @glen0071: no need to worry! New ideas from fresh perspectives are always welcome.

macfarlandian:
@jessex with shadow data we are going to have to introduce some validation behavior that allows the history to switch back and forth between age breakdowns and age unknown over time (because cases from shadow data will not have age breakdowns); that's fine for the R(t) calculation because we are using only the total case counts, not the age breakdowns, for that calculation. The future projections will still require the same internal consistency I was talking about for any given day.

glen0071:
@macfarlandian I'm close to done with this one.

However, the toast is triggered in the FacilityInformation component. To prevent it from triggering too often I was hoping to save a state (e.g. `warnedAgeConflcitsAt`), but the FacilityInformation component gets rerendered when toggling between ages known and ages unknown views, which resets the state anyways. So I was hoping to save that state on the EpidemicModelContext. I've figured out how to `write` to it (with updateModel(warnedAgeConflcitsAt: newValue)), but still working how to read from there. Does that sound like a good approach?

macfarlandian:
@glen0071 although it's drifted a little bit over time as our data models have changed, `EpidemicModelContext` is really meant only to represent the data that is needed to serve as input to the curve function. Since this is a piece of state that is really only relevant to the form UI I think it would be better if we could find a reasonable way to keep it out of that context? Other than that I think the general approach sounds good

glen0071:
@macfarlandian 

> @glen0071 although it's drifted a little bit over time as our data models have changed, `EpidemicModelContext` is really meant only to represent the data that is needed to serve as input to the curve function. Since this is a piece of state that is really only relevant to the form UI I think it would be better if we could find a reasonable way to keep it out of that context? Other than that I think the general approach sounds good

That makes sense. I wondered about that. 

I could put the state in a parent component a couple levels up, but that would lead to prop drilling, which I think we want to avoid? I could also create a context for facility information forms which seems overkill for the one piece of state, but maybe it's cleaner, and could eventually house other things?

macfarlandian:
@glen0071 whatever makes sense to you at this point is probably fine ... i'm not overly worried about a little bit of prop drilling here; as long as it's well factored we can always come back and reorganize things later if we need to

glen0071:
@jessex I updated this to trigger the toast on Change.

fac-form
modal

@macfarlandian let me know what you think of the code.

src/impact-dashboard/FacilityInformation.tsx
```diff
@@ -112,9 +126,22 @@ export const AgeGroupGrid: React.FC<AgeGroupGridProps> = ({
...
+    } else {
+      return false;
+    }
+  };
+
   useEffect(() => {
-    setCollapsed(collapseAgeInputs(props.model));
-  }, [props.model]);
+    setCollapsed(collapseAgeInputs());
+  }, []);
```
macfarlandian, line 144: I am guessing the intention here is to only have this run when the component is first mounted; that's fine but you should disable the lint rule on this line and leave a comment to that effect

src/impact-dashboard/FacilityInformation.tsx
```diff
@@ -238,6 +268,17 @@ const AgeGroupRow: React.FC<AgeGroupRowProps> = (props) => {
     }
   }
 
+  const checkAgeConflict = (model: object) => {
+    if (Date.now() > props.warnedAt + 10000) {
+      if (includesUnknownAges(model) && includesKnownAges(model)) {
+        addToast(
+          "To prevent duplicate counting, known and unknown ages cannot be entered together. Please clear either the unknown age input or all age inputs.",
+        );
```
macfarlandian, line 276: should this be an error toast?

glen0071: @jessex should I make this toast red/error colored?

src/impact-dashboard/FacilityInformation.tsx
glen0071: I removed the props.model as a dependency specifically because that was causing the age inputs to toggle unexpectedly. So I'm guessing it's best to just leave it out.

macfarlandian: yeah that seems reasonable to me

src/impact-dashboard/FacilityInformation.tsx
jessex: Agreed, let's make this toast red!

glen0071:
New 📷 
Screen Shot 2020-06-30 at 9 53 01 PM

Issue #558, opened 2020-06-11
[Shadow baseline] Visually distinguishing public data inputs from user-entered data


sychang:
**Is your feature request related to a problem? Please describe.**
This is a user-facing component of the shadow baseline project to pre-populate user scenarios with publicly collected data to reduce their data entry workload so we can improve retention and reduce user friction.

When a user has their baseline scenario synced with a shadow baseline scenario, we need to make it visually clear which inputs are pulled from public data vs. which inputs they have entered themselves or overriden.

**Describe the solution you'd like**
For all individual inputs (whether in the facility details page or add data modal), inputs with a green sync icon signal that particular input is from the shadow public data, whereas an input without the sync icon indicates user data (or no data). When a user overrides a shadow data input, the green sync icon should disappear.

![image](https://user-images.githubusercontent.com/6633887/84345598-63ab9780-ab73-11ea-8671-2504b7c4dfe4.png)

In the scenario library view, if a facility is synced to shadow data, add the sync icon next to the "last updated" date. The last updated date should reflect the date at which the syncer last ran or when the user last modified their data:

![image](https://user-images.githubusercontent.com/6633887/84456642-1e946d80-ac26-11ea-9de1-8eeacc82bd1c.png)

**Describe alternatives you've considered**
N/A

**Additional context**
Full details in [PRD](https://docs.google.com/document/d/1e6yCfUUYIZW0FlTkzzwYg80-wymZ7PkTClo_k7bqUqk/edit?ts=5ed53a10#)

Data model [design document](https://docs.google.com/document/d/1mvuQpXMPf3V164D4B7waCRArONw9iOWvE1g2SsLh03A/edit?usp=sharing)

Mocks: https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=2758%3A0
TODO: Finalize copy

Backend issue: #521
Related front-end issues: #556, #557, trigger logic issue incoming
TODO: We'll add the green sync icon in a couple more places (next to facility in scenario library view, etc.), mocks and issue incoming

philihp:
Hi! I need to take a step back from the project for a sec so I'm removing myself from this ticket. What I had done so far was pushed to the branch https://github.com/Recidiviz/covid19-dashboard/tree/philihp-input-shadow, hopefully it gives the next person a start

PR #610, opened 2020-06-24
#560 Sync reference data when new reference facilities are available

### LINKED ISSUE #560: [Shadow, sync config] When a new data shadow facility is found, enable mapping to existing facility
This is a user-facing component of the shadow baseline project to pre-populate user scenarios with publicly collected data to reduce their data entry workload so we can improve retention and reduce user friction.

Our facility and case dataset will continue to improve over time as we add new facilities. If a user has turned on syncing to public shadow data, when we detect a new shadow facility, we want the user to be able sync to that facility's data.

If a new facility gets added to SBS, the Sync Config will appear the next time the user logs into the model prompting them to map it to a facility in BS or create a new one:
- Only the user baseline facilities that are not yet mapped to SBS facilities should appear as options in the dropdown

image

N/A

**Additional context**
Full details in [PRD](https://docs.google.com/document/d/1e6yCfUUYIZW0FlTkzzwYg80-wymZ7PkTClo_k7bqUqk/edit?ts=5ed53a10#heading=h.yug6v3hbadx2)

Data model [design document](https://docs.google.com/document/d/1mvuQpXMPf3V164D4B7waCRArONw9iOWvE1g2SsLh03A/edit?usp=sharing)

Mocks: https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=2758%3A0
TODO: Finalize copy

Backend issue: #521
Related front-end issues: #556, #557, #558, #559, trigger logic and steady-state component issues incoming

daschi:
## Description of the change

This adds another modal to ask the user to sync reference data when a new reference facility is available. This is still a WIP, there's a couple of todos:

- [x] Set the `referenceDataObservedAt` property after the modals are viewed by the user
- [x] Factor out common/reusable components (started to work on this today but need more time)
- [x] Rename component to something better than "ReSyncRefFacilityModal"

## Type of change

## Related issues

Closes #560

### Development

### Code review


src/page-multi-facility/ReSyncRefFacilityModal.tsx
```diff
@@ -0,0 +1,266 @@
...
+        }),
+      );
+    }
+    onClose();
+  }
+
+  return (
+    <ModalDialog
+      open={open}
+      title={<SyncDataTitle stateName={stateName} systemType={systemType} />}
```
sprad, line 221: `ModalDialog` takes in a `width` property which you'll probably want to use to reduce the size of the modal from its default.

src/page-multi-facility/ReSyncRefFacilityModal.tsx
```diff
@@ -0,0 +1,266 @@
...
+  </SyncDataTitleContainer>
+);
+
+const FacilitiesSelect: React.FC<{
+  value: Facility["id"] | undefined;
+  selections: { [key in ReferenceFacility["id"]]: Facility["id"] };
+  facilities: Facilities;
+  onChange: (value: Facility["id"] | undefined) => void;
+}> = ({ selections, value, facilities, onChange }) => {
+  const disabledOption = (facilityId: Facility["id"]) =>
```
sprad, line 129: Good idea to disable the option for facility references that have already been selected 👍

src/page-multi-facility/ReSyncRefFacilityModal.tsx
```diff
@@ -0,0 +1,266 @@
...
+  } = useFacilities();
+
+  const [selections, setSelections] = useState<{
+    [refFacilityId: string]: Facility["id"];
+  }>({});
+
+  const scenario = scenarioState.data;
+  const syncedRefFacilities = scenario?.[referenceFacilitiesProp] || {};
+
+  if (!open) return null;
```
sprad, line 192: If we make this the first line in the component then we can short circuit lines 179-190 when the modal is not open

src/page-multi-facility/ReSyncRefFacilityModal.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  const facilities = Object.values(facilitiesState).filter(
+    filterSyncedFacilities(syncedRefFacilities),
```
sprad, line 195: `filterSyncedFacilities` isn't actually doing the filtering, right?  If not, maybe it would be better named something like `hasNoSyncedFacilities` since it's more of a check/conditional that ultimately gets used in a filter.

src/page-multi-facility/ReSyncRefFacilityModal.tsx
```diff
@@ -0,0 +1,266 @@
...
+                }}
+              />
+            </Row>
+          );
+        })}
+      </ModalContent>
+      <ModalFooter>
+        <CancelButton onClick={onClose}>Not now</CancelButton>
+        <SaveButton
+          disabled={!Object.keys(selections).length}
```
sprad, line 256: I can't click on the Save button if I don't make any selections, but it might be nice to visually gray the button out as well.

src/database/index.tsx
```diff
@@ -156,6 +156,7 @@ export const SCENARIO_DEFAULTS = {
   dataSharing: false,
   promoStatuses: {},
   baselinePopulations: [],
+  referenceDataObservedAt: undefined,
```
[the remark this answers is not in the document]
daschi: Ah yes, I'm seeing that problem now while trying to save the scenario. Would it make sense then to remove it from the defaults here and add it as an optional property on the Scenario type?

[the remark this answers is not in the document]
sprad: Yes, or otherwise you may be able to get away with just setting it to `null` instead of `undefined`, but you'd have to test that out.

src/page-multi-facility/ReSyncRefFacilityModal.tsx
daschi: I initially tried that, but you would see a react hooks lint error: 
`React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render`

sprad: Ok 🤷‍♂️

sprad:
One other thing, although I'm not strictly sure it's related to this PR.  When I map a new facility in the modal and click 'Save', nothing seems to happen.  I see that the mapping is created in Firestore, but the app doesn't respond in any way.   I was expecting that I'd see case counts on the facility page for the reference data, but the facility instead appears as if it has no data at all and the charts remain empty.

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
```diff
@@ -75,23 +78,43 @@ const ScenarioTab = styled.li<{ active?: boolean }>`
...
   const {
     state: facilitiesState,
     actions: { createOrUpdateFacility, deselectFacility },
   } = useFacilities();
-  const facilities = Object.values(facilitiesState.facilities);
+  const facilities = Object.values(facilitiesState.facilities) || [];
   const rtData = getFacilitiesRtDataById(facilitiesState.rtData, facilities);
+  const systemType = facilities[0]?.systemType;
+  const stateName = facilities[0]?.modelInputs.stateName;
+  const newlyAddedReferenceData =
```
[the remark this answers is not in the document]
daschi: Oh yeah, regarding opening for baseline scenarios--I think it's currently working since it's also checking the value from line 81:
`const referenceFacilitiesEligible = useReferenceFacilitiesEligible();`

which is checking that the `enableShadowData` flag is on and that we are looking at a baseline scenario:
 
`useFlag(["enableShadowData"]) && !!scenario.data?.baseline;`

Re: seeing the modal when all facilities are already mapped, I totally forgot that I talked to @sychang about adding a default option that is `Add as new facility` so I can add that in! Good catch

src/page-multi-facility/ReferenceDataModal/shared/ReferenceFacilitySelect.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface ReferenceFacilitySelect {
+  referenceFacilities: ReferenceFacilityMapping;
+  facilities: Facilities;
+  onChange: (
+    id: ReferenceFacility["id"],
+  ) => (facilityId: Facility["id"] | undefined) => void;
+  selections: ReferenceFacilitySelections;
+}
+
+export const ReferenceFacilitySelect: React.FC<ReferenceFacilitySelect> = ({
```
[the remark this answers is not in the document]
daschi: yes I think so!

daschi:
I'm not positive but I think displaying the reference data might be part of https://github.com/Recidiviz/covid19-dashboard/issues/558

What do you think about adding a toast when a user hits save, so that there is acknowledgment that the scenario mapping was saved?

src/page-multi-facility/ReferenceDataModal/SyncNewFacility.tsx
```diff
@@ -42,7 +42,7 @@ const SyncNewFacility: React.FC<Props> = ({ facilityId, onClose }) => {
     mappedReferenceFacilities,
     referenceFacilities,
   );
-  if (!facilityId) return null;
+  if (!facilityId || isEmpty(unmappedReferenceFacilities)) return null;
```
sprad, line 45: I think you meant to add this to `SyncNewReferenceData`, right?

sprad: Or you can add it as part of the `newlyAddedReferenceData` boolean check

src/page-multi-facility/ReferenceDataModal/SyncNewReferenceData.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const {
+    state: { facilities: facilitiesMapping, referenceFacilities },
+  } = useFacilities();
+  const scenario = scenarioState.data;
+  const mappedReferenceFacilities = scenario?.[referenceFacilitiesProp] || {};
+  const unmappedFacilities = getUnmappedFacilities(
+    mappedReferenceFacilities,
+    facilitiesMapping,
+  );
+  const unmappedReferenceFacilities = getUnmappedReferenceFacilities(
```
sprad, line 64: To clarify my previous comment, we should also return `null` if `unmappedReferenceFacilities` is empty right?

src/page-multi-facility/ReferenceDataModal/SyncNewReferenceData.tsx
```diff
@@ -0,0 +1,104 @@
...
+  <TitleContainer>
+    <TitleText>Prepopulate Data</TitleText>
+    <TitleText>
+      We've found new facility data - select a corresponding facility to
+      autofill with case data
+    </TitleText>
+    <TitleText>
+      State: {stateName || ""}
+      <Spacer />
+      Type of System: {systemType || ""}
```
sprad, line 35: `stateName` and `systemType` are  required, right?  If so, we shouldn't need these `|| ""` fallbacks

src/page-multi-facility/ReferenceDataModal/SyncNewFacility.tsx
daschi: Hmm..good question...I initially didn't add it for `SyncNewReferenceData` because I was thinking that any "newly added" reference facilities that have not been displayed to the user yet would not be synced yet. I'm not sure now if that's a solid assumption. I added it to this modal because this would open whenever the user saves a new facility, and we wouldn't want to show it if all of the reference facilities are already synced.

src/page-multi-facility/ReferenceDataModal/SyncNewReferenceData.tsx
daschi: Yeah I think that makes sense to add here too, will update this.

src/page-multi-facility/ReferenceDataModal/SyncNewFacility.tsx
sprad: Yup, I think this makes sense to include here.

I'm also confusing myself about what the exact conditions are for showing the `SyncNewReferenceData` in the context of handling both mapping unmapped user facilities to new reference facilities and/or adding new reference facilities as user facilities when all of the existing user facilities are already mapped to a reference facility.

src/page-multi-facility/ReferenceDataModal/SyncNewReferenceData.tsx
daschi: It looks like stateName is, but the type for `systemType` on the facility is `string | null` so I was wondering if I had to guard for that. When I remove the fallbacks I don't see any TS issues though, so I'll try that!

src/page-multi-facility/ReferenceDataModal/SyncNewReferenceData.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const unmappedFacilities = getUnmappedFacilities(
+    mappedReferenceFacilities,
+    facilitiesMapping,
+  );
+  const unmappedReferenceFacilities = getUnmappedReferenceFacilities(
+    mappedReferenceFacilities,
+    referenceFacilities,
+  );
+
+  if (!open || isEmpty(unmappedReferenceFacilities)) return null;
```
sprad, line 70: Ok, cool.  So I think after this line, we'll be left with `unmappedReferenceFacilities` that we can either:

1.  Map to existing users facilities if they aren't already mapped

or

2.  Add as an entirely new facility

daschi: Yeah, should I add a new issue to add that new functionality after this merges? Currently they won't yet be able to add as entirely new facility.

sprad: Just got off the phone with Daniela.  We're going to leave this as-is for now so that we can merge since it looks like the issue Sokunthel is working on (#556) may be implementing this functionality.

PR #611, opened 2020-06-24
Enhance Data Visualization in Scenario Library with Projection Graphs

### LINKED ISSUE #347: Enhance the Data Visualization in the Scenario Library

At the moment, our Scenario Library modal shows each Scenario with a simple data visualization that includes the Recidiviz logo (see the image below).  Ideally, we would provide something more visually appealing.  The [original mocks](https://www.figma.com/file/EGJS9njuKHWdFHUS4lrHA4/%F0%9F%9A%80Recidiviz-CJStatus?node-id=910%3A130) show a data visualization that represents the projection graph of the first facility within the scenario.

Screen Shot 2020-05-08 at 12 38 36 PM

Part of the solution is to figure out what the ultimate solution is and to decide if this is something we even want to pursues.  At the moment, here are a few options and some of their advantages and disadvantages:

1.  Leave the Recidiviz Logo visualization in place.
- Pros:
  - The work is already done
  - There aren't any performance implications
  - The case where a scenario does not yet have any facilities is already handled
- Cons:
  - Visually boring

2.  Implement the ideal visualization as defined in the mocks:
- Pros:
  - Looks good
  - Is somewhat relevant to the underlying Scenario
- Cons:
  - Will require additional database code (or updates) to allow for fetching a single facility
  - Will likely have performance implications due to additional data fetching and chart rendering
  - May require untangling some code since at the moment `useModel` (where we translate `modelInputs` into projection curves) must be used within an `EpidemicModelContext` which relies on a `LocaleDataSource`.  Coupling all of these things into this Scenario Library experience seems awkward from a code perspective.
  - Still need to handle the case where a scenario doesn't yet have any facilities

3.  Generate Random Graphs.  I have this semi-working in a local branch on my machine.
- Pros:
  - Looks good
  - Probably easier to implement than Option 2
  - The case where a scenario does not yet have any facilities is handled
- Cons:
  - Could be difficult to implement realistic looking curves
  - Might have to be smart about the curve generation such that generated curves are static for a given scenario.  Otherwise it might be confusing if the curves are constantly changing for the same scenario.

neeharmb:
## Description of the change

> The Scenario Library now displays the projection graph and Rt value for the first facility in a scenario instead of the Recidiviz logo. 

## Type of change

## Related issues

Closes #347

### Development

Basic testing: 
Screen Shot 2020-06-24 at 4 30 53 PM

### Code review

### DIFF
--- a/src/page-multi-facility/FacilityChart.tsx
+++ b/src/page-multi-facility/FacilityChart.tsx
@@ -96,7 +96,7 @@ const FacilityChart: React.FC<{
 
   return (
     <>
-      {firstFacility && firstFacilityRtData ? (
+      {(firstFacility && firstFacilityRtData) ? (
         <EpidemicModelProvider
           facilityModel={firstFacility?.modelInputs}
           localeDataSource={localeDataSource}
--- a/src/page-multi-facility/FacilityChart.tsx
+++ b/src/page-multi-facility/FacilityChart.tsx
@@ -1,0 +1,64 @@
+import React, { useEffect, useState } from "react";
+
+import { getFacilities } from "../database";
+import Loading from "../design-system/Loading";
+import useRejectionToast from "../hooks/useRejectionToast";
+import { EpidemicModelProvider } from "../impact-dashboard/EpidemicModelContext";
+import { getRtDataForFacility } from "../infection-model/rt";
+import { useLocaleDataState } from "../locale-data-context";
+import FacilityRow from "./FacilityRow";
+import FacilityRowPlaceholder from "./FacilityRowPlaceholder";
+import { Facility, RtValue, Scenario } from "./types";
+
+const FacilityChart: React.FC<{
+  scenarioId: string;
+}> = ({ scenarioId }) => {
+  const [firstFacility, setFirstFacility] = useState<Facility | null>();
+  const [
+    firstFacilityRtData,
+    setFirstFacilityRtData,
+  ] = useState<RtValue | null>();
+  const { data: localeDataSource } = useLocaleDataState();
+
+  const handleFacilitySave = (facility: Facility) => {
+    console.log(facility);
+  };
+
+  const fetchFacility = async () => {
+    const facilities = await getFacilities(scenarioId);
+    const firstFacility = facilities?.[0];
+    if (firstFacility) {
+      const firstFacilityRtData = await getRtDataForFacility(firstFacility);
+      setFirstFacility(firstFacility);
+      setFirstFacilityRtData(firstFacilityRtData);
+    }
+  };
+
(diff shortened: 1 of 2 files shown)


src/page-multi-facility/ScenarioLibraryModal.tsx
```diff
@@ -15,6 +15,7 @@ import useCurrentUserId from "../hooks/useCurrentUserId";
 import useRejectionToast from "../hooks/useRejectionToast";
```
jessex, line 15: Another lint check on L8 (I can't directly comment on it) to take care of.

src/page-multi-facility/FacilityChart.tsx
```diff
@@ -0,0 +1,112 @@
...
+`;
+
+interface FacilityChartProps {
+  facility: Facility | undefined;
+  facilityRtData: RtValue | undefined;
+}
+
+const FacilityChart: React.FC<{
+  scenarioId: string;
+  indexOfFacility: number;
```
macfarlandian, line 38: i'm not so sure there's much benefit to having this passed as a prop? if anything we might want to change the index we use _inside_ the component once the data is loaded (e.g. choose the first facility that has valid data, or the one that is most recently updated - not that I'm necessarily suggesting you do that here, just that it's probably a more likely future use case than arbitrarily choosing a different index from outside this component)

src/page-multi-facility/FacilityChart.tsx
```diff
@@ -0,0 +1,112 @@
...
+          setFacility(facility);
+          setFacilityRtData(facilityRtData);
+        }
+      }
+    }
+    fetchFacility();
+    return () => {
+      mounted = false;
+    };
+  }, []);
```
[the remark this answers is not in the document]
macfarlandian: to me it seems sensible to include both of these dependencies; if either of them changed you would need new data to reflect those changes anyway

src/page-multi-facility/FacilityChart.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const [facilityRtData, setFacilityRtData] = useState<RtValue | undefined>();
+  const { data: localeDataSource } = useLocaleDataState();
+
+  useEffect(() => {
+    let mounted = true;
+    async function fetchFacility() {
+      const facilities = await getFacilities(scenarioId);
+      if (facilities) {
+        const facility = facilities[indexOfFacility];
+        const facilityRtData = await getRtDataForFacility(facility);
```
macfarlandian, line 50: I would recommend not "shadowing" the state variable names like this; it doesn't break anything per se but it makes things harder to read

PR #612, opened 2020-06-24
Update Rt calc to handle duplicate cases and smooth Rt result

### LINKED ISSUE #608: Update Rt algorithm to handle infrequent testing
**Describe the bug**
Currently, the Rt calculation assumes testing is performed on a significant population at a regular cadence, and that is unfortunately not what we are seeing in the facility data. Therefore, the facility Rts from the existing algorithm are quite bumpy and are too sensitive to infrequent testing (Connecticut & Wisconsin are top of mind). On the other hand, states with strict testing but few cases are also not being handled properly because we are dropping days with duplicate case counts so Rt never trends towards 0 (Vermont).

**Expected behavior**
An Rt that is consistently high for facilities that are mid-outbreak, and an Rt that is trending down for states that are testing well and have had no outbreak. 

**Additional context**
The proposed fix for this issue is two-fold:
1. Add a 7 day average on top of the Rt alg result in order to smooth the Rts over time and reduce large spikes/drops that are caused by spotty testing
2. Remove the hack that alters the case data and drops duplicate case counts across days
These fixes should be validated against the facilities that we have been working with & monitoring (both ones with and without serious case counts).

justkunz:
## Description of the change

Make the Rt calculation more robust for scenarios with spotty testing that are going through an outbreak, and scenarios that are not. This includes 2 changes: removing the portion that excludes days with duplicate case counts, and smoothing the Rt result to reduce the noise caused by sporadic testing. 

I've tested this in my own research notebook locally, but have not tested this end to end in the covid app. 

## Type of change

- [X] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

Closes #608 

### Development

### Code review

### DIFF
--- a/backend/python-functions/realtime_rt.py
+++ b/backend/python-functions/realtime_rt.py
@@ -132,7 +132,7 @@ def compute_r_t(historical_case_counts):
     # we need at least two days to represent change over time
     if len(smoothed) < 2:
         raise ValueError('Not enough data to compute R(t);'
-            ' at least two days of non-repeating case counts are required')
+            ' at least two days of data are required')
 
     # Note that we're fixing sigma to a value just for the example
     posteriors, _ = get_posteriors(smoothed, sigma=.25)
--- a/backend/python-functions/realtime_rt.py
+++ b/backend/python-functions/realtime_rt.py
@@ -126,10 +126,6 @@ def compute_r_t(historical_case_counts):
         raise ValueError(
             'Case counts must be cumulative and monotonically increasing')
 
-    # we believe duplicates are mostly an artifact of little-to-no testing
-    # and should be excluded from the model
-    case_df = case_df.drop_duplicates(keep='first')
-
     _, smoothed = prepare_cases(case_df)
 
     # Raise an error if there are not enough valid cases to use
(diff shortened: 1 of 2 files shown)


backend/python-functions/realtime_rt.py
```diff
@@ -126,10 +126,6 @@ def compute_r_t(historical_case_counts):
         raise ValueError(
             'Case counts must be cumulative and monotonically increasing')
 
-    # we believe duplicates are mostly an artifact of little-to-no testing
-    # and should be excluded from the model
-    case_df = case_df.drop_duplicates(keep='first')
```
macfarlandian, line 131: Github won't let me comment on it directly but there is an error message below on L134 that makes reference to duplicate case counts; if that requirement is changing we should also change the message to avoid confusion. (I think saying "two days of data" instead of "two days of non-repeating case counts" is probably sufficient?)

justkunz: Good call out! I've updated the error message accordingly and changed the python test to make sure only the first day is dropped in the output.

PR #616, opened 2020-06-29
Add Recovered Cases and Deaths to Case Counts

### LINKED ISSUE #561: Separate recovered/deceased cases from case count for projection
As states are reaching steady state in fighting COVID, lots of incarcerated individuals have recovered and a state's cumulative case count is no longer the best indicator of their projection (because the model currently assume mid-outbreak on the cumulative cases). 

Allow states to tell us how many of their covid cases are "recovered" or "deaths" so we can separate those from the projections.

1. (UI) In the add cases modal, add a column of inputs between "CURRENT CASES (CUMULATIVE)" and population labeled "RECOVERED CASES"
2. (UI) Add a second column to the right of recovered labeled "DEATHS"
3. (UI) Change the label of "CURRENT CASES (CUMULATIVE)" to "CUMULATIVE CASES"
4. (UI) Repeat 1-3 in the facility detail view
5. (UI) Update the copy in the add cases modal to say "Enter the number of cases detected and the total population as of this date."
6. (Model calculation) In creating the projection curve, we will now be projecting based on "active" cases only, which is cumulative - recovered - deaths.

- We could have states directly enter "active" cases but it is less common for states to have that data readily available.
- We could have the model take "new" cases instead of cumulative cases, but may not be a number states readily have available.
- We could not make the change and leave it as cumulative cases, but we risk some states abandoning the model because it does not accurately represent their system at present anymore.

The proposed approach also aligns with our shadow data collection efforts, which are tracking the number of recovered cases.

**Additional context**
There will be a few model updates as a part of this change:
1. Updating the projection to distribute cases across the active and recovered buckets 
2. Update the r(t) calculation so it becomes a smoothed 7-day rolling average
3. Remove the r(t) hack for discarding duplicate counts

neeharmb:
## Add Recovered Cases and Deaths to Case Counts

> Modified UI to allow users to input recovered cases and deaths in the add cases modal and facility detail view. Updated Epidemic Model to include recovered cases and deaths. Altered the placeholder text for numeric inputs because the old placeholder text was consistently being cut-off with more input areas. Added frontend validations to ensure that recovered/death/cumulative cases are monotonically increasing.

## Type of change

## Related issues

Closes #561

### Development

### Code review


src/design-system/InputTextNumeric.tsx
```diff
@@ -49,7 +49,7 @@ const InputTextNumeric: React.FC<Props> = (props) => {
       inputRelativityError={props.inputRelativityError}
       labelPlaceholder={
         props.labelPlaceholder ??
-        (props.type === "number" ? "Enter number" : "Enter a percentage")
+        (props.type === "number" ? "0" : "Enter a percentage")
```
jessex, line 52: Why did we make this particular change?

src/impact-dashboard/EpidemicModelContext.tsx
```diff
@@ -22,22 +22,40 @@ export enum RateOfSpread {
 // fields that we want to store
 export interface ModelInputsPopulationBrackets {
   age0Cases?: number;
+  age0Recovered?: number;
   age0Population?: number;
+  age0Deaths?: number;
```
jessex, line 27: Slight order inconsistency here. Should move this line up one, and the line directly below one, to be consistent in the _cases, recovered, deaths, population_ ordering.

src/infection-model/validators.tsx
```diff
@@ -41,9 +48,13 @@ const validateCasesBySum = (
...
   compareFn: ComparisonFunction,
 ): boolean => {
-  return compareFn(
-    totalConfirmedCases(current),
-    totalConfirmedCases(comparedTo),
+  return (
+    compareFn(totalConfirmedCases(current), totalConfirmedCases(comparedTo)) &&
+    compareFn(
+      totalConfirmedRecoveredCases(current),
+      totalConfirmedDeaths(comparedTo),
```
jessex, line 55: This should be `totalConfirmedRecoveredCases`, no?

neeharmb:
@jessex sorry for the confusion! I was just about to send it out for review and found a bug with the relativity input errors so I kept it in draft mode. If this happens again I'll comment on the PR to avoid any confusion.

src/design-system/InputTextNumeric.tsx
neeharmb: The placeholder text was consistently getting cut off in both the add cases modal and facility detail view now that there are four columns:
image (2)

This actually could happen in prod with a really small window/screen size, but usually didn't:
image (3)

One alternative that we explored was splitting the placeholder text into multiple lines, but it didn't seem feasible/worthwhile after discussing with @macfarlandian so we did this instead. 

CC: @sychang

neeharmb:
CC: @justkunz

src/database/index.tsx
```diff
@@ -754,7 +754,7 @@ export const saveFacility = async (
             })
           ) {
             throw new Error(
-              "Failed to save. Case and death counts are cumulative, and can only increase over time.",
+              "Failed to save. Cumulative cases, recovered cases, and deaths can only increase over time.",
```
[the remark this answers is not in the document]
neeharmb: I believe the changes in `src/infection-model/validators.tsx` should be ensuring that recovered cases/deaths are monotonically increasing along with cumulative cases? Otherwise I can change the message and/or add the validation logic elsewhere.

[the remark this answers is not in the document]
macfarlandian: oh yes I see that now, you're right ... I didn't quite follow the chain of function calls far enough. never mind!

src/design-system/InputTextNumeric.tsx
jessex: Got it, I think this makes sense. Thanks!

Issue #588, opened 2020-06-17
[Report automation] Section 1.2: Intervention impact to-date projection (excalibur v2)


sychang:
**What needs to be done? Why does it need to be done?**
Excalibur MVP was limited because
- We did not have trustworthy historical case counts 
- We generalized the original projection to be one mega facility because we did not have specific facility attributes.

Thanks to shadow data, we will have facility attributes and by-facility historical case counts, which will enable us to make more accurate impact projections that show how actual case counts are tracking against original projections. This is the most hotly requested feature as states “wrap up” their covid response and want to see how they are tracking against their original projections.

**Actual cumulative cases line**
A discrete line will show the total cumulative case counts over time from the date of first case until today, summed across all facilities for each day.

**Original projection line** 

- For each modelled facility, find the date of the first covid case. If the date of the first case was 90 days ago
- Take the facility information at that date, if available, otherwise pull the current facility information. If no facility attributes are available, use generic assumptions (in PRD)
- Use a moderate generic r(t) (3.0) for the facility and project forward to today. The upper bound of cases for each facility should be the total population of the facility. (Alternate, harder consideration in PRD)
- Sum up the projections of the individual facilities at each day to get the system-wide projection line. 

![image](https://user-images.githubusercontent.com/6633887/86651679-57b3c980-bfa9-11ea-9608-b942ba0060e2.png)

Example: If MO has 3 facilities and today is June 17
- Facility 1 got its first case on June 4. Use 1 case on June 4 and project 13 days forward
- Facility 2 got its first case on March 1. Use the case count on March 19 (90 days ago) and project forward to today
- Facility 3 got its first 2 cases on June 1. Use 2 cases on June 1 and project 17 days forward

**Additional context**
Full details in PRD section: https://docs.google.com/document/d/1WwT52cF1I-Wmn7FomkUkMxFT93yTtvP8XGCRYuKKRQs/edit#heading=h.1nmkz8z4hlkk

Subtask of #170
Related to: #589

PR #621, opened 2020-07-01
582 - Fetch underlying data for weekly report 

### LINKED ISSUE #582: [Report automation] Determine underlying data for weekly snapshot
Thanks to shadow baseline work, weekly reports can be generated for all states, even if no one from the state is using the web model. The more complete a state’s baseline or shadow baseline data, the better the report so if states want the best report, we want to bring them back to the model.

If a state has a particular baseline model that they want to base a report off of (usually the DOC researcher), they need to 1) sync it with shadow baseline, 2) share their model with covid@recidiviz.org, and a report will be generated using the data in their baseline scenario.
Otherwise, state reports will be general using a state’s shadow baseline.

Pseudocode user flow:
When an admin selects a state from the dropdown on admin page, model.recidiviz.org/weekly-snapshot (#531), for that selected state, perform two checks:
1. If a state has shared its baseline scenario with covid@recidiviz.org (“State” field of facilities matches the selected state):
2. If the user has shadow “data sync” on (this check can be added in later once shadow baseline is complete 
--Generate a PDF report for the state based on the user’s baseline scenario.

If the user has “data sync” off or no user from the state has shared a baseline scenario with covid@recidiviz.org, generate the same PDF report based on the state’s shadow baseline scenario.

More context in the [PRD](https://docs.google.com/document/d/1WwT52cF1I-Wmn7FomkUkMxFT93yTtvP8XGCRYuKKRQs/edit#)

**Additional context**
Subtask of #170

daschi:
## Description of the change

I'm opening this in draft state to ask for some feedback on the approach. I ended up creating a new `WeeklyReportContext` to keep track of the state, because I wasn't sure how best to re-use the facilities & scenario contexts given these use cases. The state looks like this:

```
{
  loading: boolean;
  stateName: string | null;  -- selected by the user from the dropdown
  scenario: Scenario | null; -- set when stateName changes, if scenario is not shared with user then it will be null
  sharedScenarios: Scenario[]; -- set when component mount, fetches all shared baseline scenarios
  facilities: Facilities | ReferenceFacility[]; -- if there is a scenario with userReferenceData turned on, 
  this will be an array of composite facilities, otherwise it will be an array of referenceFacilities with some added 
  "facility" properties. 
}
```

I still have some questions about whether there is other data we need to add to the reference facilities if there is not a shared scenario or if the shared scenario does not have "data sync" turned on. 

## Type of change

## Related issues

Closes #582

### Development

### Code review


src/page-weekly-snapshot/WeeklyReportContext.tsx
```diff
@@ -0,0 +1,192 @@
...
+      getSharedScenarioByStateName(state.sharedScenarios, state.stateName).then(
+        (scenario) => {
+          dispatch({ type: "RECEIVE_SCENARIO", payload: scenario || null });
+        },
+      );
+    }
+  }, [state.stateName, state.sharedScenarios]);
+
+  useEffect(() => {
+    const systemType = "State Prison";
```
sprad, line 129: Are we only concerned about State Prisons and not County Jails for this feature?

src/page-weekly-snapshot/WeeklyReportContext.tsx
```diff
@@ -0,0 +1,192 @@
...
+        dispatch({
+          type: "RECEIVE_FACILITIES",
+          payload: Object.values(facilities),
+        });
+      } else {
+        if (stateName) {
+          const referenceFacilities = await fetchReferenceFacilities(
+            stateName,
+            systemType,
+          );
```
sprad, line 156: You can probably move this `fetchReferenceFacilities` call outside of the `if/else` statement since we need to call it either way if `stateName` is present.  You may have to assign `referenceFacilities` to an empty array by default so that it can be in scope inside of the `if/else` statement branches

src/page-weekly-snapshot/WeeklyReportContext.tsx
daschi: yes I think so, just confirmed with @sychang !

daschi:
I'm putting this back as a WIP PR because I'm realizing there's going to be a lot of changes to this context for #588. I think it might make sense to open one PR for adding the data and using it in the impact chart.

daschi:
I'm closing this PR because I'm realizing there's going to be a lot of changes to this context for #588. I think it might make sense to open one PR for adding the data and using it in the impact chart.

PR #623, opened 2020-07-01
Update case distribution with active/recovered/deaths

### LINKED ISSUE #607: Incorporate Recovered Case Counts into Covid Model Projection
The FE is adding functionality to gather the number of recovered cases per facility. These counts should be incorporated into the covid case projection logic such that they do not count as infected cases, and are not eligible to infect others. 

Update the simulate portion of the Javascript, where the cases are distributed across the SEIR buckets. If recovered cases are provided:
* Get the active cases by subtracting the recovered cases from the total cumulative cases.
* Distribute active cases across active buckets, distribute recovered cases across recovered cases. 
If recovered cases are not provided do not alter the existing logic.

Compute the estimated recovered case total from the historical case counts. This would be more accurate than our current approach, but less accurate than the desired approach.

**Additional context**
This ticket is intended to contain part of the DS work from #561, and will make the Covid dashboard more realistic and will build more trust from users. This work cannot be fully merged until the FE supports the recovered case input.

justkunz:
## Description of the change

Use the recovered & death counts collected by the frontend to distribute cases for the SEIR model projection. I've updated the percentages based off Rt values that are more realistic for the facilities Zak has been working with (1.4 instead of 3.0). This shifts more of the active cases from the infectious group to the recovering groups. I've also split the "recovered" cases between our two recovered groups, and now initialize the fatalities set using the death count. I have not tested any of these changes!

## Type of change

- [X] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

Closes #607 

### Development

### Code review

### DIFF
--- a/src/infection-model/seir.tsx
+++ b/src/infection-model/seir.tsx
@@ -315,7 +315,7 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
   ).forEach(([pop, cases, recovered, deaths], index) => {
     // distribute cases across compartments proportionally
 
-    if ((typeof recovered !== 'undefined') & (typeof recovered !== 'undefined')) {
+    if ((typeof recovered !== 'undefined') && (typeof recovered !== 'undefined')) {
         const activeCases = Math.max(cases - recovered - deaths, 0);
 
         const infectious = activeCases * pInitiallyInfectious;
--- a/src/infection-model/seir.tsx
+++ b/src/infection-model/seir.tsx
@@ -315,7 +315,7 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
   ).forEach(([pop, cases, recovered, deaths], index) => {
     // distribute cases across compartments proportionally
 
-    if ((typeof recovered !== 'undefined') && (typeof recovered !== 'undefined')) {
+    if ((typeof recovered !== 'undefined') && (typeof deaths !== 'undefined')) {
         const activeCases = Math.max(cases - recovered - deaths, 0);
 
         const infectious = activeCases * pInitiallyInfectious;
--- a/src/infection-model/seir.tsx
+++ b/src/infection-model/seir.tsx
@@ -315,61 +315,79 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
   ).forEach(([pop, cases, recovered, deaths], index) => {
     // distribute cases across compartments proportionally
 
-    if ((typeof recovered !== 'undefined') && (typeof deaths !== 'undefined')) {
-        const activeCases = Math.max(cases - recovered - deaths, 0);
+    if (typeof recovered !== "undefined" && typeof deaths !== "undefined") {
+      const activeCases = Math.max(cases - recovered - deaths, 0);
 
-        const infectious = activeCases * pInitiallyInfectious;
-        // exposed is related to the number of infectious
-        // but it can't be more than the total uninfected population
-        const exposed = Math.min(infectious * ratioExposedToInfected, pop - cases);
+      const infectious = activeCases * pInitiallyInfectious;
+      // exposed is related to the number of infectious
+      // but it can't be more than the total uninfected population
+      const exposed = Math.min(
+        infectious * ratioExposedToInfected,
+        pop - cases,
+      );
 
-        singleDayState.set(index, seirIndex.susceptible, pop - cases - exposed);
-        singleDayState.set(index, seirIndex.exposed, exposed);
-        singleDayState.set(index, seirIndex.infectious, infectious);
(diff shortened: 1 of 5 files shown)


src/infection-model/seir.tsx
```diff
@@ -307,31 +308,34 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
   ).forEach(([pop, cases, recovered, deaths], index) => {
     // distribute cases across compartments proportionally
 
-    const infectious = cases * pInitiallyInfectious;
+    const activeCases = cases - recovered - deaths;
```
justkunz, line 311: Do I need any error checking here to make sure these are all valid numbers and/or that active cases is non-negative?

src/infection-model/seir.tsx
```diff
@@ -307,31 +308,34 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
...
       index,
       seirIndex.hospitalized,
-      cases * pInitiallyHospitalized,
+      activeCases * pInitiallyHospitalized,
     );
     singleDayState.set(
       index,
       seirIndex.mildRecovered,
-      cases * pInitiallyMildRecovered,
+      recovered * pInitiallyMildRecovered,
```
justkunz, line 331: If users haven't entered in any recovered cases yet the two recovered groups will be 0, which may be a weird UX when this change is initially deployed. Should I support both situations when users have zero recovered (use the current logic) and non-zero recovered (use the updated logic)?

src/infection-model/seir.tsx
macfarlandian: yeah, for safety I would probably clamp it at zero with a call to `Math.max` or something like that ... validation in the UI is still pretty haphazard so we can't necessarily trust that inputs will be sane

src/infection-model/seir.tsx
macfarlandian: i agree, this does seem wrong ... I think maybe the condition to detect is when cases are defined but recovered is `undefined` ... that way if someone explicitly enters `0` for recovered we don't just ignore that value.

src/infection-model/seir.tsx
```diff
@@ -307,31 +308,34 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
...
     );
     singleDayState.set(
       index,
       seirIndex.severeRecovered,
-      cases * pInitiallySevereRecovered,
+      recovered * pInitiallySevereRecovered,
     );
+    singleDayState.set(index, seirIndex.fatalities, deaths);
```
macfarlandian, line 338: should there be a similar condition for deaths as for recovered above? (i.e. previously we were assigning a proportion of cases to be fatalities, now we're not doing that anymore so they get artificially zeroed out when we release this feature)

macfarlandian:
I am curious what @sychang thinks about what should happen for users who are encountering this breaking change in our data model for the first time .... do we explicitly warn them, do we try to preserve the old version of the model seamlessly for them, do we do something else?

sychang:
Is the breaking change that their projection will look different if they haven't touched their model? Or something else? If it's the former, one straightforward warning we could give users is to update the copy on the add cases modal to say something along the lines of "we've updated the model to project based on active/recovered cases, so it may look a lil different."

If something actually breaks until they take some action, like no longer being able to see any projection at all, we will probably need to do something to make it more seamless.

macfarlandian:
@sychang it is the former

src/infection-model/seir.tsx
justkunz: Cool, clamped the active cases at a minimum of 0 to be safe.

src/infection-model/seir.tsx
justkunz: I tried to add a check for "if recovered & deaths have been filled out vs if they are not filled out" to fall back to the old distribution logic if the user hasn't added recovered/death counts.

src/infection-model/seir.tsx
```diff
@@ -307,31 +315,62 @@ export function getAllBracketCurves(inputs: CurveProjectionInputs) {
...
-      index,
-      seirIndex.mildRecovered,
-      cases * pInitiallyMildRecovered,
-    );
-    singleDayState.set(
-      index,
-      seirIndex.severeRecovered,
-      cases * pInitiallySevereRecovered,
-    );
+    if ((typeof recovered !== 'undefined') && (typeof recovered !== 'undefined')) {
```
macfarlandian, line 318: was one of these `recovered`s supposed to be `deaths`?

justkunz: Yes! Just updated that, thanks for catching it

macfarlandian:
@justkunz i think you need to run `yarn lint-all-fix` on your local to resolve this lint failure FYI

PR #634, opened 2020-07-06
Add Promo Status for New Model Inputs

### LINKED ISSUE #561: Separate recovered/deceased cases from case count for projection

neeharmb:
## Add Promo Status for New Model Inputs

> Adding recovered cases and deaths is a breaking change that will impact how projection values appear. This promo notifies the user so they are aware of this change.

Note: The changes in `src/infection-model/seir.tsx` are coming from Prettier; once that is run in `jkunz/initialize-recovered-seir` that file should no longer appear in this PR. 

## Type of change

## Related issues

Closes #561

### Development

### Code review


src/page-multi-facility/ScenarioSidebar.tsx
```diff
@@ -83,6 +85,12 @@ const promoTexts: { [promoType: string]: string } = {
     for that facility. To get an accurate Rt, enter case counts for previous
     days by clicking the number of cases on the right (in red) to update case
     numbers.`,
+  newModelInputs: `New! We’ve updated our model to factor in recoveries and 
+    deaths to enable more accurate projections based on active cases. This 
+    may have caused some of your projection values to change. You can input 
+    recoveries and deaths when you “Add Historical Data” for a facility.
```
jessex, line 91: Nit but the double quotes in here are not the "standard" double quotes and may render strangely in some browsers, if I recall. Usually you get these when they are copied and pasted from somewhere, but we should replace them.

src/page-multi-facility/CreateBaselineScenarioPage.tsx
```diff
@@ -47,6 +47,7 @@ const CreateBaselineScenarioPage: React.FC = () => {
         dataSharing: true,
         dailyReports: true,
         addFacilities: true,
+        newModelInputs: false,
```
jessex, line 50: Not sure I immediately follow: why is this one set to false?

src/page-multi-facility/ScenarioSidebar.tsx
neeharmb: good catch, thanks!

src/page-multi-facility/CreateBaselineScenarioPage.tsx
neeharmb: Essentially, we only want this promo to be shown for existing scenarios (newly created ones should already be using recovered cases/deaths), so we set it to false here. Existing scenarios won't have this promo status yet, so we set it to true in `src/database/type-transforms.tsx` only if it is undefined, if that makes sense

PR #643, opened 2020-07-11
[Report Automation Section 1.2] Impact to date chart 

### LINKED ISSUE #588: [Report automation] Section 1.2: Intervention impact to-date projection (excalibur v2)

daschi:
## Description of the change

This adds a chart to the weekly snapshot page that displays the projected and actual cases and fatalities for the past 90 days up to today. There's still a few more tasks for this PR, but would appreciate any feedback!

Todo: 

- [x] Chart annotations (i.e. # Cases - Today's cumulative staff and inmate cases)
- [x] Remove tooltip (currently using it in development to check the data)
- [x] Add a legend at the bottom of the chart and a chart title
- [x] Add the missing properties from the reference facilities for fatalities

From the design:
image

with example arizona data:

image

## Type of change

## Related issues

Closes #588 and #582 

### Development

### Code review


src/database/index.tsx
```diff
@@ -273,6 +273,51 @@ export const getScenarios = async (): Promise<Scenario[]> => {
...
+    console.error(
+      `Encountered error while attempting to retrieve shared scenarios:`,
+    );
+    console.error(error);
+
+    return [];
+  }
+};
+
+export const getScenarioByStateName = async (
```
jessex, line 301: What if there are multiple scenarios that have facilities matching the state name? I didn't double check the API but doesn't the `find` call below just return the first result found in the `scenarios` array?

src/page-weekly-snapshot/projectionChartUtils.tsx
```diff
@@ -0,0 +1,187 @@
...
+export function getCaseCount(curveData: ndarray<number>) {
+  const cases = [];
+  for (let day = 0; day < curveData.shape[0]; ++day) {
+    const caseCountForDay = countCasesForDay(curveData, day);
+    cases.push(caseCountForDay);
+  }
+  return cases;
+}
+
+export const today = dateFns.endOfToday();
```
jessex, line 29: I might be way wrong about this, but will these lines only be executed and these constants be defined and exported only once when the app first starts and the export graph is initialized? That would mean that these exports would become outdated when we passed into the next day after the first deploy that includes this code. Let me know if I just misunderstand and if these are actually being lazily initialized or something!

src/page-weekly-snapshot/projectionChartUtils.tsx
```diff
@@ -0,0 +1,187 @@
...
+      );
+      return dateFns.isSameDay(version.observedAt, date);
+    });
+    versionForProjection = closestVersionWithCases;
+  }
+  return versionForProjection;
+}
+
+export function getProjectedData(epidemicModelInputs: EpidemicModelInputs) {
+  if (!epidemicModelInputs || !epidemicModelInputs?.observedAt) {
```
jessex, line 64: There are no expected cases where this conditional would be true and _not_ be exceptional? e.g. for a brand new baseline scenario?

src/page-weekly-snapshot/projectionChartUtils.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  return {
+    projectedFatalities: addProjectionPadding(
+      getAllValues(getColView(summedProjectionData, seirIndex.fatalities)),
+    ),
+    projectedCases: addProjectionPadding(getCaseCount(summedProjectionData)),
+  };
+}
+
+export function getActualDataForFacility(modelVersions: ModelInputs[]) {
```
jessex, line 101: A comment: logic in generic files like `*Utils` should tend to have docstrings since 1) it tends to be where we put dense logic to spare callers, and 2) it tends to be general-purpose for all kinds of varied consumers, which makes it important for future developers to easily understand the functionality.

src/page-weekly-snapshot/projectionChartUtils.tsx
daschi: Yeah, that makes sense...I think it could cause an issue if the app isn't reloaded, and I would actually want a function then to return endOfToday() whenever we it's being used, or just use the `dateFns.endOfToday()` directly.

src/database/index.tsx
daschi: That's a good point, at the moment this is very closely tied to the context of receiving the `scenarios` param from the return of `getSharedBaselineScenarios`, so I was expecting it to be a collection of baseline scenarios shared with the `covid@recidiviz` user, and from convos with @sychang it seemed like we could assume that only 1 baseline scenario per US State would be available. I could update this to actually return all the results (and rename it to `getScenariosByStateName`) so that it could work in other contexts, and use the first (and presumably only) result in the weekly report context.

src/page-weekly-snapshot/projectionChartUtils.tsx
daschi: Yes that makes sense...ideally I would want to add tests for these too. I added comments, and it's making me see that maybe some of these functions are doing too much. Let me know what you think, and if there is a different format that would be better for the comments.

src/page-weekly-snapshot/projectionChartUtils.tsx
daschi: Yeah I think it is possible for a new baseline scenario to be shared without any cases/version history, I wasn't sure yet at this point what to do in that case, so I updated this to return zero for each day of the projected line. If there's another strategy that would make more sense in this case let me know, happy to talk it through too.

daschi:
I'm seeing an issue on the multi-facility page, and I'm not sure if it's from something I've changed or if my scenario data got into a weird state. Here's a screenshot of the error, looking at the data it looks like that scenario somehow has `null` for the `updatedAt` property:
image

I can look into this more closely tomorrow, but might reach out for some data support!

src/page-weekly-snapshot/projectionChartUtils.tsx
jessex: These look perfect. Thanks!

src/database/index.tsx
jessex: Looks great, thanks!

jessex:
@daschi, does that happen in production for you, or only on your branch? I can't reproduce it myself but if you give your scenario you're using an easily identifiable name, I can go look it up and let you know what I find.

daschi:
Strange, the issue seems to be gone now and everything is loading fine!

src/page-weekly-snapshot/ImpactProjectionChart.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const {
+    state: { stateName, loading: scenarioLoading },
+  } = useWeeklyReport();
+
+  if (!stateName) return null;
+
+  const facilities = Object.values(facilitiesState.facilities);
+  const modelVersions = facilities.map((f) => f.modelVersions);
+
+  if (!scenarioLoading && !facilities.length) {
```
jessex, line 106: Just clarifying: we don't check `localeLoading` here because it's conceptually prior to the `scenarioLoading` check, and if the local data is still loading it doesn't mean anything about the presence of scenario data. Is that right?

daschi: Yeah, actually I don't think it makes sense to check `localeLoading` on line 214, since this component isn't going to render until locale state is loaded. I'll update this to remove that extra check

PR #645, opened 2020-07-13
Add Per Capita and State Stats to Weekly Snapshot

### LINKED ISSUE #587: [Report automation] Section 1.1: Per capita state and system stats
Expanding on the local stats pulled from #531, calculate

- Per capita cases in the state and per 100k, state rank out of all 50 states (source NYT, can adapt script from #531)
- Per capita deaths in the state and per 100k, state rank out of all 50 states (source NYT, can adapt script from #531)
- Per capita cumulative incarcerated cases per 100k incarcerated = # cumulative cases / # incarcerated individuals x 100k 
- Per capita cumulative incarcerated deaths per 100k incarcerated = # cumulative deaths / # incarcerated individuals x 100k

Mock is not final:
image

**Additional context**
PRD section here: https://docs.google.com/document/d/1WwT52cF1I-Wmn7FomkUkMxFT93yTtvP8XGCRYuKKRQs/edit#heading=h.9rilufs1xgg8

Subtask of #170

neeharmb:
## Add Per Capita and State Stats to Weekly Snapshot

> Add incarcerated/state cases and fatalities per capita to weekly snapshot.

test:

image

## Type of change

## Related issues

Closes #587

### Development

### Code review

### DIFF
--- a/src/page-weekly-snapshot/LocaleSummaryTable.tsx
+++ b/src/page-weekly-snapshot/LocaleSummaryTable.tsx
@@ -1,0 +1,389 @@
+import { sum } from "d3-array";
+import { orderBy } from "lodash";
+import React, { useState } from "react";
+import styled from "styled-components";
+
+import Colors, { MarkColors as markColors } from "../design-system/Colors";
+import { Column, PageContainer } from "../design-system/PageColumn";
+import { useFacilities } from "../facilities-context";
+import {
+  formatThousands,
+  TableRow,
+} from "../impact-dashboard/ImpactProjectionTable";
+import {
+  LocaleData,
+  LocaleDataProvider,
+  LocaleRecord,
+  useLocaleDataState,
+} from "../locale-data-context";
+import { Facility } from "../page-multi-facility/types";
+
+const HorizontalRule = styled.hr`
+  border-color: ${Colors.opacityGray};
+  margin: 10px 0;
+`;
+
+export const Table = styled.table`
+  color: ${Colors.black};
+  text-align: left;
+  width: 100%;
+  margin-top: 10px;
+`;
+
+const TableHeadingCell = styled.td`
+  font-family: "Poppins", sans serif;
+  line-height: 24px;
+  margin-right: 5px;
+  vertical-align: middle;
+`;
+
+const BorderDiv = styled.div`
+  border-top: 2px solid ${Colors.darkGray};
+  margin-right: 5px;
+`;
+
+const ProjectionSection = styled.div`
+  margin-top: 10px;
+  padding: 5px 0;
+`;
+
+const ProjectionContainer = styled.div`
+  .axis-title text,
+  .axis-label {
+    fill: ${Colors.black};
+    font-family: "Libre Franklin";
+  }
+`;
+
+const TableCell = styled.td<{ label?: boolean }>`
+    font-size: 13px;
+    line-height: 200%;
+    text-align: "left"};
+    opacity: 0.7;
+    border-top: 1px solid  ${Colors.darkGray};
+    vertical-align: middle;
+    width: ${(props) => (props.label ? "200px" : "auto")};
+    `;
+
+type StateMetrics = {
+  stateName: string;
+  casesPerCapita: number | undefined;
+  deathsPerCapita: number | undefined;
+};
+
+const POPULATION_SIZE = 100000;
+
+function getPerCapita(
+  numerator: number | undefined,
+  denominator: number | undefined,
+) {
+  let result = 0;
+  if (numerator && denominator) {
+    result = Math.round((numerator / denominator) * POPULATION_SIZE);
+  }
+  return result;
+}
+
+function getCurrentStateData(stateMetrics: StateMetrics[], stateName: string) {
+  for (let i = 0; i < stateMetrics.length; i++) {
+    if (stateMetrics[i].stateName === stateName) {
+      return stateMetrics[i];
+    }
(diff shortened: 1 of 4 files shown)


src/page-weekly-snapshot/LocaleSummaryTable.tsx
```diff
@@ -0,0 +1,368 @@
...
+function getPerCapita(
+  numerator: number | undefined,
+  denominator: number | undefined,
+) {
+  let result = 0;
+  if (numerator && denominator) {
+    result = Math.round((numerator / denominator) * POPULATION_SIZE);
+  }
+  return result;
+}
```
macfarlandian, line 115: sorry to be pedantic but some of the naming here is a bit confusing to me ... _per capita_ means "per person" but that is not what you are calculating. You could call it `getPer100K` and then `POPULATION_SIZE` is no longer a magic number and doesn't need a constant. Or you could call it `getRate` or something like that and call the constant something a little less vague like `RATE_UNIT` or `RATE_MULTIPLIER`?

(same general pedantry applies to other uses of "per capita" in this PR)

src/page-weekly-snapshot/LocaleSummaryTable.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+function getAllStateData(localeData: LocaleData, stateNames: string[]) {
+  const stateMetrics: StateMetrics[] = [];
+  // D.C. isn't a state!
+  stateNames = stateNames.filter((item) => item !== "District of Columbia");
```
macfarlandian, line 120: this is true .... do we definitely want to exclude it from these metrics? Does it break anything to leave it in?

src/page-weekly-snapshot/LocaleSummaryTable.tsx
```diff
@@ -0,0 +1,368 @@
...
+    const currStateDataPerCapita = get(currState, filterValue);
+    if (
+      selectedStateName &&
+      currState.stateName === selectedStateName &&
+      currStateDataPerCapita
+    ) {
+      return [currStateDataPerCapita, i + 1];
+    }
+  }
+  return [-1, -1];
```
macfarlandian, line 162: this return value is pretty obscure ... it is not immediately obvious why it takes two numbers to represent the state's rank. could this be an object instead with more explicit keys?

src/page-weekly-snapshot/LocaleSummaryTable.tsx
```diff
@@ -0,0 +1,368 @@
...
+  width: ${(props) => (props.label ? "200px" : "auto")};
+`;
+
+type StateMetrics = {
+  stateName: string;
+  casesPerCapita: number | undefined;
+  deathsPerCapita: number | undefined;
+};
+
+type incarceratedData = {
```
macfarlandian, line 86: nit: type names should be `PascalCase`

src/page-weekly-snapshot/LocaleSummaryTable.tsx
```diff
@@ -0,0 +1,368 @@
...
+  let incarceratedCasesPerCapita = 0;
+  let incarceratedDeathsPerCapita = 0;
+
+  let hasCaseData = true;
+  let hasDeathData = true;
+
+  const allStateMetrics = getAllStateData(localeData, stateNames);
+
+  // TODO: currently getting state name from the drop down
+  // will need to update whenever that changes
```
macfarlandian, line 287: if this isn't a TODO for this PR it should have a ticket number associated with it

src/page-weekly-snapshot/LocaleSummaryTable.tsx
```diff
@@ -0,0 +1,368 @@
...
+    incarceratedDeathsPerCapita = getPerCapita(
+      totalIncarceratedDeaths.incarceratedData,
+      totalIncarceratedPopulation.incarceratedData,
+    );
+    hasDeathData = totalIncarceratedDeaths.hasData;
+  }
+
+  return (
+    <>
+      <HorizontalRule width={"93%"} marginLeft={"20px"} />
```
macfarlandian, line 333: i see what you mean now about the magic numbers ... the margin value at least is something that could be turned into a constant (since it is used both here and by the `Column` component). Why do you even need the width value though? will it not just fill its container on auto width?

src/page-weekly-snapshot/LocaleSummaryTable.tsx
jessex: Ideally we would leave it in but if we can't because it breaks something, then let's let @sychang know so she can prioritize future support.

neeharmb: I actually did ask @sychang about this, and it could just potentially make the rankings weird because including D.C. means we are ranking out of 51 "states." More specifically, leaving D.C. in might mean someone sees a "51st out of 50" ranking

src/page-weekly-snapshot/LocaleSummaryTable.tsx
neeharmb: After playing around with some different CSS ideas, I spoke to Juan directly about this and ended up deciding on a design closer to this, which was much more straightforward to implement: 
image

src/page-weekly-snapshot/LocaleSummaryTable.tsx
jessex: Got it. Thanks for the context, @neeharmb

Issue #657, opened 2020-07-24
Improve performance for large scenarios


macfarlandian:
**What needs to be done? Why does it need to be done?**
As shadow data features have started to come online, making it easier to bulk-add a large number of facilities to a scenario, we are seeing performance start to suffer on the multi-facility dashboard. The task here is to investigate this and either solve it or file tickets for the proposed solution if it is more involved. 

"Large" here means somewhere in the neighborhood of 10+ (can vary since it is likely an issue of memory consumption).

**Additional context**
We do know some potentially expensive operations that increase as a function of how many facilities there are, but we don't know exactly where the current bottleneck lies. 

- Does the main culprit seem to be number of facilities, or the amount of history those facilities have? (e.g. does a large number of facilities with little history trigger it?) Does the shadow data feature itself have anything to do with it, or is it mainly a function of how many facilities there are regardless of whether that feature is enabled?
- There is already a feature in place where we attempt to defer the rendering of `FacilityRow` components until they scroll into view. Is this working as expected? (i.e. are these components still doing some expensive work in the render loop before they are actually visible)
- The curve functions (i.e. whatever hooks ultimately call the projection functions in `infection-model/` are known to be expensive. If this is a bottleneck, there are some incremental steps we have already thought about:
    - make these hooks async (i.e. have them return Promises instead of blocking the React render loop)
    - if that doesn't improve things enough, move the expensive SEIR projections to a Cloud Function and replace the in-browser computations with an additional HTTP request
- There are at least two Firestore requests that we do per facility now (one for the facility data and a second one for its history), plus a Cloud Function request for the Rt history. None of these requests individually should have a particularly high data volume, but is the number of requests itself causing some kind of problem? (Doubtful but worth mentioning as a last resort)
- Is there some other expensive or redundant work that is getting repeated per `FacilityRow` that was not previously a known issue?

macfarlandian:
accidentally closed by a partial PR; reopening

neeharmb:
Marking this closed per #664; outstanding performance issues can be filed as separate tickets

PR #663, opened 2020-07-30
Performance Improvements - Part One

neeharmb:
## Performance Improvements - Part One

> "Hey kids, ~~spelling~~ memoizing is fun!" --Taylor Swift at some point, probably

Changes in this PR:
1. Switch `useState` + `useEffect` patterns to `useMemo`: The expensive function `buildCurves` has been called every time the main page re-rendered, and the result has been thrown away. Switching to `useMemo` caches the results of the last call and prevents the function from being called again if the dependencies haven't changed.

2. Make tooltips optional for projection graphs on the main page: I was noticing a spike in browser memory usage when hovering over projection graphs and decided to make adding tooltips an optional parameter. This seems to have reduced memory consumption and I synced with @sychang that they are not necessary to have on the main page. This parameter can also be used in other places (i.e. report automation) where tooltips are not needed. 

3. Moving `ProjectionsPanel` into a separate function and memoize: I'm actually not sure if this memoization here helps, but in the worst case it still is recalculated on every re-render (as it currently is) and this makes the code more readable. 

None of these changes appear to have impacted functionality from my testing, but please let me know if you notice any strangeness!

Next Steps:
While these changes seem to get larger scenarios loaded without the browser freezing or crashing, navigation to facilities/back to the main page is still very slow. Looking into that next!

## Type of change

## Related issues

Related to #657 

### Development

### Code review

### DIFF
--- a/src/page-multi-facility/MultiFacilityImpactDashboard.tsx
+++ b/src/page-multi-facility/MultiFacilityImpactDashboard.tsx
@@ -7,14 +7,14 @@ import Colors from "../design-system/Colors";
 import iconAddSrc from "../design-system/icons/ic_add.svg";
 import Loading from "../design-system/Loading";
 import TextLabel from "../design-system/TextLabel";
-import { useFacilities } from "../facilities-context";
+import { FacilitiesState, useFacilities } from "../facilities-context";
 import { useFlag } from "../feature-flags";
 import useReadOnlyMode from "../hooks/useReadOnlyMode";
 import useReferenceFacilitiesEligible from "../hooks/useReferenceFacilitiesEligible";
 import useRejectionToast from "../hooks/useRejectionToast";
 import { EpidemicModelProvider } from "../impact-dashboard/EpidemicModelContext";
 import { getFacilitiesRtDataById } from "../infection-model/rt";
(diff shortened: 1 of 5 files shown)


src/page-multi-facility/MultiFacilityImpactDashboard.tsx
```diff
@@ -198,7 +210,15 @@ const MultiFacilityImpactDashboard: React.FC = () => {
...
           </ScenarioTabs>
         </div>
-        {selectedTab === 0 && projectionsPanel}
+        {selectedTab == 0 && (
+          <ProjectionsPanel
+            facilitiesState={facilitiesState}
+            facilities={facilities}
+            localeDataSource={localeDataSource}
+            rtData={rtData}
+            handleFacilitySave={handleFacilitySave}
```
macfarlandian, line 219: `handleFacilitySave` is still getting re-created on every render, so memoizing probably isn't going to do anything with this component as-is; however, I think you should be able to wrap `handleFacilitySave` in a `useCallback` hook and stabilize it somewhat; I think its dependencies should be pretty stable.

sychang:
+1 to add more email spam, this is incredible. Taylor Swift would be proud.

PR #671, opened 2020-08-04
Make selectFacility async 

### LINKED ISSUE #657: Improve performance for large scenarios

neeharmb:
## Make selectFacility async 

> This makes the logic behind selecting and navigating to a facility more tightly coupled, which may prevent some of the sporadic navigation issues I was seeing. 

## Type of change

## Related issues

Closes #657 

(Note: there are probably still existing performance issues, but I think it makes sense to open new tickets when those are identified) 

### Development

### Code review

### DIFF
--- a/src/facilities-context/FacilitiesContext.tsx
+++ b/src/facilities-context/FacilitiesContext.tsx
@@ -43,7 +43,7 @@ export type ExportedActions = {
   ) => Promise<void>;
   duplicateFacility: (facility: Facility) => Promise<Facility | void>;
   deselectFacility: () => void;
-  selectFacility: (facilityId: Facility["id"]) => void;
+  selectFacility: (facilityId: Facility["id"]) => Promise<void>;
   fetchReferenceFacilities: (
     stateName: string,
     systemType: string,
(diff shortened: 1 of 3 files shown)


neeharmb:
This code doesn't do anything useful; if there are issues with navigation, a new ticket can be opened

PR #679, opened 2020-08-10
Fetch locale data CSV via cloud function

### LINKED ISSUE #652: [Shadow data] Scenarios with shadow data have null updatedAt values sometimes
After populating a scenario with reference facilities, a bug happens where a scenario's `updatedAt` field gets set to `null`, which causes a breaking error in the function that reads the scenario data from the database (all data on the homepage fails to load and you see infinite spinners). The strange thing is that the value is not actually `null` in the database; the stored scenario data appears to be unaffected. The error can affect not only your current scenario but all scenarios you have access to (including those shared with you).

It's possible that the shadow data aspect is a red herring because there is no obvious connection between that feature and this bug, but that is the condition under which it was initially observed. (This does not happen in prod, but it does in latest master with the shadow data feature flag active.)

### LINKED ISSUE #678: Replace use of CORS-anywhere proxy with a cloud function
This service, while useful for testing, was not intended to be used in production. It is not reliable and occasionally rate-limits us. We do need some kind of intermediary to read locale data directly from a published Google Sheets CSV but we should switch to something self-hosted instead.

macfarlandian:
## Description of the change

This new cloud function essentially acts as a proxy for the CSV fetch that adds a CORS header and passes the data through otherwise untouched (the same thing we were doing by routing the request through `cors-anywhere.herokuapp.com`).

Also fixed the null timestamp issue as part of this because it was blocking my ability to test this end-to-end. (It is still kind of mysterious to me what triggers it but I was able to figure out that it is at least expected behavior by Firestore that we were not handling properly.)

I have not yet deployed this new cloud function; if you want to test it locally there are instructions in the backend folder for running the functions locally; set the new environment variable introduced in this PR to point your local Gatsby app at the local function emulator.

## Type of change

## Related issues

Closes #678 , closes #652 

### Development

### Code review


package.json
```diff
@@ -107,7 +107,7 @@
     "!(*.d).ts": [
       "false -- Please use .tsx instead of .ts!"
     ],
-    "*/**/*.js": [
+    "src/**/*.js": [
```
macfarlandian, line 110: I don't think we've made any backend JS changes since introducing this pre-commit hook; it was objecting to the JS functions not being in Typescript. This changes it to only run that check for client app source files.

.env-example
```diff
@@ -4,3 +4,5 @@
 
 GATSBY_SHOW_IMPACT_BUTTON="false"
 GATSBY_ENABLE_SHADOW_DATA="false"
+# you probably only need to change this if you are developing cloud functions
+GATSBY_USE_LOCAL_CLOUD_FUNCTIONS="false"
```
jessex, line 8: Nit: newline

macfarlandian:
backend deployed and frontend verified against deployment, FWIW

Issue #642, opened 2020-07-09
[Shadow Data] Toggle the App to Use the Production Covid Case Data


sprad:
**Is your feature request related to a problem? Please describe.**

At the moment, we’ve siloed our app to read Shadow Reference Data from a Firestore collection called `reference_facilities_test`.  We also store the mapping of user facilities to reference facilities in an attribute on the user's Baseline scenario called `testReferenceFacilities`.  Both of these "pointers" will need to be updated to `reference_facilities` and `referenceFacilities` respectively when we launch.  The specific changes will need to be made [here](https://github.com/Recidiviz/covid19-dashboard/blob/3dc6fc9ede152798ee96f5de5b2bb5d3d18192e0/src/database/index.tsx#L57) and [here](https://github.com/Recidiviz/covid19-dashboard/blob/3dc6fc9ede152798ee96f5de5b2bb5d3d18192e0/src/database/index.tsx#L60).

**WARNING:**  This should not be done until all other Shadow Reference Data Issues have been completed and QA'ed as noted in this parent issue: #521

macfarlandian:
closed by #698

PR #681, opened 2020-08-12
Ingest facility metadata from CSV

### LINKED ISSUE #640: [Shadow Data] Import Facility Reference Metadata into Firestore

In order to run calculations with Shadow Reference Facilities, we will need to import Facility Metadata into Firestore.  Over the past month, the public data team has compiled this metadata into a BigQuery table called `covid_public_data.facility_data`.  The task here will be to import the data from this BigQuery table and merge it in with our existing `reference_facilities` documents.  At the moment, these documents only contain the following fields:

- canonicalName
- facilityType
- stateName
- createdAt
- covidCases

At a minimum, we will additionally need to import the following information from the BigQuery table:

- capacity
- population

Ideally, we would also import:

- county (as countyName)
- securityClass
- yearOpened

Note that the `capacity` and `population` fields can change over time so they should be modeled in Firestore as an array data type that holds objects with the following fields:

- date
- value

`date` will represent the date at which we know about a population or capacity and `value` will store the actual integer number of a facility's population or capacity on that date.

Implementation details are a little TBD, but at the end of the day, the data needs to move from the mainline Recidiviz Google Cloud Project's BigQuery table into the Covid-19 Dashboard Google Cloud Project's Firestore instance.

A seemingly reasonable approach would be to write a Google Cloud Function that moves the data.  The challenge here will be setting up the permissions such that one Google Cloud Project can talk to another Google Cloud project.  

In terms of matching Reference Data to the data we store in Firestore, we have been using the combination of a facility's `stateName` and `canonicalName` as a unique identifier/key.

If the reference facility doesn't already exist within the `reference_facilities` Firestore collection, then we should create a new record.

An example of the working Google Cloud Function that imports COVID case data and stores it in our `reference_facilities` collection can be found [here](https://github.com/Recidiviz/covid19-dashboard/blob/master/backend/python-functions/data_ingest.py).  Note: this function just parses a CSV file and stores it in Firestore vs. integrating with BigQuery.

Finally, in terms of import cadence, for the first version of this function, we can just perform the import on a set schedule of once a week or once a month.  In the future, we may want to do something more "real-time" and have imports kick off any time BigQuery is updated.

**Additional context**

In order to implement this issue, the person who does this work will need to have access to both:

1.  The Google Cloud Project that runs the Recidiviz Covid-19 Dashboard.  In particular, this person will need to be able to deploy Cloud Functions and access Firestore.
2.  The Google Cloud Project that houses the `covid_public_data.facility_data` BiqQuery table.

macfarlandian:
## Description of the change

Creates a new Python cloud function that can consume a CSV export of our facility metadata from BigQuery and put it into Firestore. (The schema I tested this with is just a straight dump of the `facility_data` view, though we don't necessarily consume all the fields in it.)

This is a different solution than the one described in the ticket. Some notable features:
- Data coming out of BigQuery has the benefit of unique facility ids; we can also use these as IDs in Firestore for easier lookups. (The daily data ingest function can/should also be updated to take advantage of this)
- This new ID scheme is essentially a **breaking change** on the existing Firestore collection, which gets populated from the existing daily case data ingest. We can't really deploy this without breaking stuff until we also update that function (or disable it), though I did temporarily deploy this and point it at a separate test collection in Firestore just to verify that it works.
- This doesn't necessarily require any automated exports from BigQuery (since the data isn't expected to change very frequently); I tested it by manually uploading files to the specified bucket. It does allow for that automation to happen, though, as long as the end result of said automation is the creation of a file in this bucket. That work would happen outside this codebase anyway.
- Another **breaking change** that bears mentioning: in the final metadata schema, facility capacity did not wind up being an explicitly versioned data point (unlike facility population); rather than implement our own wonky versioning scheme as part of this ingest logic, I think we should just leave it unversioned and change the way we handle it on the frontend.

## Type of change

## Related issues

Closes #640 

### Development

### Code review


backend/.gitignore
```diff
@@ -1,4 +1,6 @@
 firebase-debug.log
+service-account.json
```
jessex, line 2: We should document in the readme that this is required in order to deploy the Cloud Functions backend.

backend/python-functions/data_ingest.py
```diff
@@ -1,12 +1,102 @@
...
+            row_ops_count = 1
+            id = row['facility_id']
+
+            facility_doc_ref = facilities_collection.document(id)
+            facility_dict = {}
+
+            # are we updating or creating?
+            try:
+                fdoc_snapshot = facility_doc_ref.get()
+                assert fdoc_snapshot.exists
```
jessex, line 42: Personally, I have an aversion to using exceptions to conduct normal, accepted control flows. I would just do a standard if-else check on `.exists` here instead of asserting.

backend/python-functions/data_ingest.py
```diff
@@ -1,12 +1,102 @@
...
+def create_or_update_facilities(file_location):
+    batch = fs_client.batch()
+    with open(file_location, newline='') as f:
+        reader = csv.DictReader(f)
+
+        batch_counter = 0
+        for row in reader:
+            # what appears to be a single "write" can contain multiple operations;
+            # we need to keep track of them to stay below the Firestore batch limit
+            row_ops_count = 1
```
jessex, line 33: This is annoying! But okay.

backend/python-functions/data_ingest.py
```diff
@@ -1,12 +1,102 @@
...
+            id = row['facility_id']
+
+            facility_doc_ref = facilities_collection.document(id)
+            facility_dict = {}
```
jessex, line 37: Let's call this `facility_metadata` instead of `facility_dict`. We should try to avoid Hungarian notation.

backend/python-functions/data_ingest.py
```diff
@@ -1,12 +1,102 @@
...
+                        matching_record['value'] = latest_population['value']
+                else:
+                    facility_dict['population'] = [latest_population]
+
+            batch_counter += row_ops_count
+            # if we have hit our limit, start a new batch before proceeding
+            if batch_counter >= BATCH_SIZE:
+                batch.commit()
+                batch = fs_client.batch()
+                batch_counter = row_ops_count
```
jessex, line 92: I don't immediately understand why we are resetting batch_counter to the ever-increasing row_ops_count at this point, but my brain may just be friend. Walk me through what's happening here?

backend/python-functions/main.py
```diff
@@ -34,3 +34,12 @@ def ingest_covid_case_data(event, _context):
...
     ingest_daily_covid_case_data(event['bucket'], event['name'])
+
+
+def ingest_facility_metadata(event, _context):
+    """
+        Ingests facility metadata. This function is triggered when a new
+        CSV file containing facility metadata is placed into the
+        c19-backend-facility-metadata bucket.
+    """
+    ingest_facility_metadata_file(event['bucket'], event['name'])
```
jessex, line 45: Nit: newline

backend/.gitignore
macfarlandian: oh right, I meant to do that but got sidetracked. But FWIW this is not actually required for deployment; it was only used for local testing (it let me write to Firestore from my local env)

backend/python-functions/data_ingest.py
macfarlandian: `row_ops_count` is not ever-increasing, it gets reset to `1` when we start a new row (on L33).

Because we are not necessarily incrementing by 1 every time, it's possible to overshoot the limit. So the logic here is, if processing this row _would_ put us at or over the limit, we start a new batch first and then handle this row as part of that new batch. Thus the initial number of ops in our new `batch_counter` is the number we will need to process the current row.

It's not the most intuitive thing, I agree ... maybe it would be a little clearer if we only did this once we had exceeded the batch size? And/or I could just write a longer comment about all this nonsense

macfarlandian:
@jessex so far I have tested it by temporarily changing `REFERENCE_FACILITIES_COLLECTION_ID` to something else and deploying it as a "development" version. We can continue doing that as needed, but the good news is that nothing is actually using the `reference_facilities` collection yet so we can pretty much wipe it out at will. 

There are a few moving parts that we have to coordinate to actually deploy this, though: 
- the `ingest_covid_case_data` function should be updated to handle the breaking schema changes (it is currently consuming the old spreadsheet format; I think we should stop that and start consuming that from a scheduled BigQuery export also per #682, but I am still figuring out the details of that)
- the existing `reference_facilities` collection should just be deleted right before we deploy both of these functions, which will repopulate it
- there is a little bit of frontend work to handle the change in the `capacity` field; I think the least disruptive way to fix that would be to do it at the same time as when we cut over from the test collection to the real collection in #642. (the changes are all in the guts of the `FacilitiesContext`, there shouldn't be any changes required in any of the UI components that consume that data)

backend/python-functions/data_ingest.py
jessex: No it makes more sense now that I look at it again. I think it's fine as-is!

macfarlandian:
no, I don't think that's actually necessary because all of the development is still happening against the `reference_facilities_test` collection anyway. The one thing that will cause problems here in the meantime is the ongoing ingestion of the old spreadsheet files for daily case data, but I am working on getting that stopped. Once that is resolved we should be able to deploy the cloud functions freely as they are updated.

Issue #641, opened 2020-07-09
[Shadow Sync Config ] Chart / Visualizations Updates


sprad:
**Describe the bug**

At the moment, when a user maps a Reference Facility, there are no corresponding changes within the app.

**To Reproduce**

1. Ensure that `GATSBY_ENABLE_SHADOW_DATA` is set to true in your local `.env.development` file.
2. Go to your Baseline Scenario and remove any existing facilities
2. Refresh the app
3. You should now be presented with a modal that asks you to select a state and system type.  Select "Arizona" and "State Prison" respectively.
4. You should now be presented with a dialog that allows you to select reference facilities to add.  Note that be default all of the facilities should be selected.
5.  Click Save.
6.  Note that the facilities will be added, but none of the charts will be updated.  See the Screenshots below.

**Expected behavior**

The expected behavior is that the charts and cases counts should be updated to reflect the data stores in the `covidCases` found field of the corresponding Shadow reference data.

**Screenshots**

**Users without any Facilities are Presented with the new Reference Data Modal**

<img width="1680" alt="Screen Shot 2020-07-09 at 5 36 29 PM" src="https://user-images.githubusercontent.com/25503/87094111-56360b80-c20c-11ea-90e9-d2c7315bf669.png">

**Click Save to Map the Reference Facilities to Your Baseline Scenario**

<img width="1680" alt="Screen Shot 2020-07-09 at 5 36 40 PM" src="https://user-images.githubusercontent.com/25503/87094130-60f0a080-c20c-11ea-885f-1260531ee020.png">

**Note that None of the Case Counts or Charts are Updated**

This is the bug this issue seeks to address

<img width="1680" alt="Screen Shot 2020-07-09 at 5 37 04 PM" src="https://user-images.githubusercontent.com/25503/87094153-677f1800-c20c-11ea-920f-46ab51d3ff6b.png">

PR #691, opened 2020-08-18
Don't create facilities from reference data with empty model inputs

### LINKED ISSUE #651: [Shadow data] New facilities added from reference data have incorrect modelInputs
When adding a facility from the modal that lists existing reference facilities for a particular state and system type, the resulting Facility objects have a modelInputs and most recent modelVersions object that has the locale data but case and population counts that are "undefined". These objects should be discarded by the facility merge process and the most recent day with actual (reference) data should be used in their place. There is a bug either in the facility creation process or the facility merge function that is preventing this from being done properly.

macfarlandian:
## Description of the change

This fixes some broken tests against `FacilitiesContext` and also resolves the issue where a facility's `modelInputs` could be blank after creating it from reference data via the `SyncNoUserFacilities` modal.

## Type of change

## Related issues

Closes #651 

### Development

### Code review

### DIFF
--- a/src/facilities-context/transforms.tsx
+++ b/src/facilities-context/transforms.tsx
@@ -106,7 +106,7 @@ function mergeModelVersions({
   userVersions,
   referenceFacility,
 }: Optional<MergedHistoryInputs, "referenceFacility">): ModelInputs[] {
-  const combinedVersions: ModelInputs[] = [...userVersions];
+  const combinedVersions: ModelInputs[] = [...userVersions.filter(hasCases)];
   let { stateName, countyName } = combinedVersions.length
     ? (last(combinedVersions) as ModelInputs)
     : ({} as ModelInputs);
--- a/src/facilities-context/__fixtures__/mergedFacilities.tsx
+++ b/src/facilities-context/__fixtures__/mergedFacilities.tsx
@@ -103,6 +103,7 @@ export const compositeFacility: Facility = {
     {
       observedAt: new Date(2020, 4, 28),
       ageUnknownCases: 5,
+      ageUnknownDeaths: 0,
       ageUnknownPopulation: referenceFacility.population[0].value,
       isReference: true,
       stateName,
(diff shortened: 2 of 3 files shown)


daschi:
nice! this is similar to what I was about to do for #641. After this merges I'll rebase from here.

src/facilities-context/transforms.tsx
```diff
@@ -106,7 +106,7 @@ function mergeModelVersions({
   userVersions,
   referenceFacility,
 }: Optional<MergedHistoryInputs, "referenceFacility">): ModelInputs[] {
-  const combinedVersions: ModelInputs[] = [...userVersions];
+  const combinedVersions: ModelInputs[] = [...userVersions.filter(hasCases)];
```
daschi, line 109: I was also considering saving the reference facility's transformed covidCases data directly to the newly mapped userFacility (from `SyncNoUserFacilities` line 258. That way we wouldn't have to filter out the "fake" modelInputs and the userFacility created from the referenceFacility would already have the correct modelInputs and versions. Do you think that would make sense?

macfarlandian: hm yeah, I thought about doing that ... I am not dead-set against it necessarily but there were a couple of issues I saw that made this seem like an easier fix, at least within the scope of the bug I was looking at: 

1. by default it sets the current day as `observedAt`, which is likely to be wrong. it should be set to whatever the most recent observed at date is in reference data. not sure how much work that would be to fix
2. i was a little leery of putting data into the user's facilities collection that they didn't actually enter themselves. In practice it's probably not a huge deal but it did seem slightly unclean somehow. We could do it if it's really necessary for some reason, but as far as I could tell for this issue anyway, it's not?

what do you think? is there something I'm missing or something relevant to your ticket that you think might indicate a different fix here?

daschi: I need to do some more testing but my initial hunch was that the rtData calculation fetches were firing first with the incorrect modelInputs, and they were being saved as errors after the second round of rtData calculations were requested and returned. I think this fix makes sense for this issue, I can keep digging on the rtData calculations for the referenceFacilities from this modal tomorrow morning.

macfarlandian: sounds plausible ... FWIW i didn't really see difference in that behavior with this change

macfarlandian:
@daschi ........ dare I suggest you approve this PR, if you are OK with it? or is that conditional on you digging into your bug a bit more?

PR #692, opened 2020-08-19
Data quality checks for reference data

### LINKED ISSUE #689: [Shadow data] data quality trigger logic for activating feature
the shadow data feature should not be available on a user's baseline scenario if:
- not all facilities in the scenario have the same state and system type; OR
- we do not have at least three reference facilities matching the state and sytem type, ~with at least half of those facilities having 7+ days of data~ (UPDATE: not worth implementing, nearly all states pass this check now)

this only applies to scenarios that do not yet have "useReferenceData" toggled on; once they have enabled the feature, we don't need to check these conditions anymore (and it should remain available even if it stops complying with them)

macfarlandian:
## Description of the change

There are a couple of data quality thresholds that have to be met (one on user data, one on reference data) for the reference data feature to be enabled. Because they are data-dependent, they have to be done within the `FacilitiesContext`, and rather than the hook we were using to expose an eligibility flag (for, e.g., showing a promo modal), the context itself now owns that flag. 

Note that, per the ticket, if the user has already managed to turn on the feature for their scenario, these eligibility checks should have no effect. (We do this to avoid confusing churn from scenarios drifting in and out of eligibility; that was more of a concern when there was also a "data freshness" check that was in scope for this feature, but it's still something that could result from user error and shouldn't need to cause that level of disruption.)

## Type of change

## Related issues

Closes #689 

### Development

### Code review


src/facilities-context/reducer.tsx
```diff
@@ -60,11 +63,14 @@ export function facilitiesReducer(
       return { ...state, referenceFacilities: action.payload };
 
     case actions.CLEAR_REFERENCE_FACILITIES:
-      return { ...state, referenceFacilities: {} };
+      return { ...state, referenceFacilities: {}, canUseReferenceData: false };
 
     case actions.CLEAR_FACILITIES:
       return { ...state, facilities: {}, loading: false };
```
daschi, line 69: Would we need to dispatch `type: SET_CAN_USE_REFERENCE_DATA, payload: false` from the `initializeFacilities` else branch, so that it is set to false whenever facilities or referenceFacilities are cleared?

macfarlandian: yeah that would probably be better than doing it this way ... more explicit, probably more reliable

PR #693, opened 2020-08-20
641 Update charts & rtdata after adding reference facilities from SyncNoUserFacilities modal

### LINKED ISSUE #641: [Shadow Sync Config ] Chart / Visualizations Updates

daschi:
## Description of the change

This fixes the issue where a user added reference facilities from the "SyncNoUserFacilities" modal and the chart and rt data were not populating. I think the issue came from the `EpidemicModelContext` not being updated with the new facility model for each row, and a timing issue with requesting rt data using the incorrect model inputs before the user facilities became merged with the reference model inputs. I used the `hasCases` validation to only request rt data for the facilities context when the model inputs has case data, this prevented the calculation requests with the incorrect inputs to fire. I'm not sure if this will cause any issues for other facilities that might have undefined case data in the inputs, but I assume having no rt data for those facilities would be expected behavior?

I also needed to remove the check for `canUseReferenceData` in the logic for showing the SyncNoUserFacilities modal, since this will be `false` for scenarios that have no facilities. I moved the check down to the next modal's conditions for `showSyncNewReferenceData`.

## Type of change

## Related issues

Closes #641 

### Development

### Code review

### DIFF
--- a/src/facilities-context/FacilitiesContext.tsx
+++ b/src/facilities-context/FacilitiesContext.tsx
@@ -142,6 +142,36 @@ export const FacilitiesProvider: React.FC<{ children: React.ReactNode }> = ({
     },
   };
 
+  async function prepareReferenceFacilities(facilities: FacilityMapping) {
+    // fetch reference facilities based on user facilities
+    // first facility is the reference; assume they're all the same
+    const userFacility = Object.values(facilities)[0];
+    if (!userFacility) return facilities;
+    const {
+      modelInputs: { stateName },
+      systemType,
+    } = userFacility && userFacility;
+
+    if (stateName && systemType) {
+      const referenceFacilities = await facilitiesActions.fetchReferenceFacilities(
+        stateName,
+        systemType,
+      );
+
+      dispatch({
+        type: facilitiesActions.RECEIVE_REFERENCE_FACILITIES,
+        payload: referenceFacilities,
+      });
+
+      return facilitiesActions.buildCompositeFacilities(
+        facilities,
+        referenceFacilities,
+        facilityToReference,
+      );
+    }
+    return facilities;
+  }
+
   // when a new scenario is loaded, facility data must be initialized
   // in a specific order to avoid unwanted side effects:
   // we fetch user facility data, then fetch reference data (when applicable);
@@ -239,14 +269,24 @@ export const FacilitiesProvider: React.FC<{ children: React.ReactNode }> = ({
   useEffect(
     () => {
       if (state.loading || isEmpty(facilityToReference)) return;
-      facilitiesActions.receiveFacilities(
-        dispatch,
-        facilitiesActions.buildCompositeFacilities(
-          state.facilities,
-          state.referenceFacilities,
-          facilityToReference,
-        ),
-      );
+      async function updateFacilities() {
+        let facilities = state.facilities;
+        // If user just synced new reference facilities and is updating
+        // facilityToReference for the first time after initialization, then
+        // referenceFacilities will not be loaded yet.
+        if (isEmpty(state.referenceFacilities)) {
+          facilities = await prepareReferenceFacilities(facilities);
+        } else {
+          facilities = facilitiesActions.buildCompositeFacilities(
+            facilities,
+            state.referenceFacilities,
+            facilityToReference,
+          );
+        }
+
+        facilitiesActions.receiveFacilities(dispatch, facilities);
+      }
+      updateFacilities();
     },
     // we are controlling changes to the state imperatively via the action functions,
     // so it should be safe to exclude. When facilityToReference changes (externally
--- a/src/facilities-context/FacilitiesContext.tsx
+++ b/src/facilities-context/FacilitiesContext.tsx
@@ -3,6 +3,7 @@ import React, { useEffect } from "react";
 
 import { referenceFacilitiesProp } from "../database";
 import { useFlag } from "../feature-flags";
+import { hasCases } from "../impact-dashboard/EpidemicModelContext";
 import {
   Facility,
   RtDataMapping,
@@ -299,7 +300,10 @@ export const FacilitiesProvider: React.FC<{ children: React.ReactNode }> = ({
     const facilities = Object.values({ ...state.facilities });
     if (facilities.length) {
       facilities.forEach((facility) => {
-        if (!state.rtData.hasOwnProperty(facility.id)) {
+        if (
+          !state.rtData.hasOwnProperty(facility.id) &&
+          hasCases(facility.modelInputs)
+        ) {
           facilitiesActions.fetchFacilityRtData(dispatch)(facility);
         }
       });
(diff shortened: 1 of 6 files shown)


src/facilities-context/FacilitiesContext.tsx
```diff
@@ -259,7 +283,10 @@ export const FacilitiesProvider: React.FC<{ children: React.ReactNode }> = ({
     const facilities = Object.values({ ...state.facilities });
     if (facilities.length) {
       facilities.forEach((facility) => {
-        if (!state.rtData.hasOwnProperty(facility.id)) {
+        if (
+          !state.rtData.hasOwnProperty(facility.id) &&
+          hasCases(facility.modelInputs)
```
macfarlandian, line 288: i'm not so sure about this ... I see why it works here but I am a bit concerned that it really depends on a side effect of the way we are creating these facilities. I think maybe if we can find a way to make those updates imperatively in such a way that we prevent the "intermediate" facility objects from showing up in the context (as we purport to do on L275) then we wouldn't have to do this anymore. 

I had an idea about that .... 

what if we made a new action function that combined most of the [handleSave](https://github.com/Recidiviz/covid19-dashboard/blob/master/src/page-multi-facility/ReferenceDataModal/SyncNoUserFacilities.tsx#L249) function in `SyncNoUserFacilities` with the new `updateFacilities` function you added above, in such a way that it waits to update the context until after the reference facilities are fetched? (we are going to have to factor that out anyway because there will eventually be two modals from which you can create new facilities from reference data, Neehar and I are working on the second one now)?

We can defer that if you think it's too much scope increase for this PR (I wouldn't argue with that) but wonder what you think of the idea anyway, or whether you have a better one

src/facilities-context/FacilitiesContext.tsx
```diff
@@ -238,15 +239,38 @@ export const FacilitiesProvider: React.FC<{ children: React.ReactNode }> = ({
...
+        isEmpty(state.facilities)
+      )
+        return;
+
+      async function updateFacilities() {
+        let referenceFacilities = state.referenceFacilities;
+        // If user just synced new reference facilities and is updating
+        // facilityToReference for the first time after initialization, then
+        // referenceFacilities will not be loaded yet.
+        if (isEmpty(referenceFacilities)) {
```
macfarlandian, line 254: since we have to fetch the reference facilites in order to display the `SyncNoUserFacilities` modal anyway, what if we just added them to the context at that point so that this object won't be empty when the facilities get saved? (I guess we would have to export an additional action from the `FacilitiesContext` to support that)

daschi: Oh yeah, that makes sense...we are already doing the fetching in the modal, it just gets set to an internal state. That's a better flow, I'll update this.

src/facilities-context/FacilitiesContext.tsx
daschi: I like the idea of creating the composite facilities right away in the handleSave of the modal, I think that would remove the need for this `hasCases` check here and then like you said we wouldn't have these "intermediate" facility objects. I had originally thought about doing that, but this seemed like the quicker fix at the time, though I can see that it's hidden in here and might cause other side effects. I'll work on updating the handleSave of the modal so that it creates the new facilities and updates the context with the composite facilities. Let me know if that sounds right to you or if I misunderstood anything!

macfarlandian: yeah that sounds good! thanks

src/facilities-context/validators.tsx
```diff
@@ -25,6 +25,14 @@ const removeNonsenseCases: SingleDayValidator = (modelInputs) => {
       )
         return null;
     }
+  } else if (
+    modelInputs?.ageUnknownCases === undefined &&
+    modelInputs?.staffCases === undefined
```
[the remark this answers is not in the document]
daschi: Ah yes, I think with the change to `handleSave` in the modal, this shouldn't be an issue anymore. I'll remove this once I update that.

src/page-multi-facility/MultiFacilityImpactDashboard.tsx
```diff
@@ -132,7 +132,6 @@ const MultiFacilityImpactDashboard: React.FC = () => {
   const systemType = facilities[0]?.systemType;
   const stateName = facilities[0]?.modelInputs.stateName;
   const showSyncReferenceFacilitiesBaseConditions =
-    facilitiesState.canUseReferenceData &&
```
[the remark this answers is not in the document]
daschi: hmm yeah the logic for the reference data seems to be a bit complicated, here's what I understand are the different cases:

1. *Allow user to select whether or not to sync reference data*
2. *Scenario is eligible for syncing new reference data*
  - `enableShadowData` feature flag 
  - scenario.baseline

3. *Scenario is eligible for syncing reference data to existing facilities*
  - user facilities > 0
  - all facilities from the same system type
  - scenario.useReferenceData (user has selected whether or not to use reference data)

Kind of thinking aloud here, but maybe we could add 2 conditions to the context? Using what you already have defined as `referenceDataEligible` for the first two cases, and `referenceDataActive` for the third case for differentiating using reference data after the user has selected that option for the scenario. I'm not sure what happens currently if the user is on the SyncNoUserFacilities modal and they don't select the `useReferenceData` toggle...I'll need to test that out, does it just take them to the dashboard with zero facilities?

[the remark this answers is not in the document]
daschi: Now I'm rereading it, I think this is pretty much exactly the logic you have already and maybe it's just a matter of making both of those variables available in the context?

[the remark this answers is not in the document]
macfarlandian: yeah .... I need to think about this some more. I think you can just unblock this ticket by adding that to the context and if we need to iterate more later then that's fine

PR #696, opened 2020-08-25
Prevent repeat Rt requests while first one is loading

### LINKED ISSUE #688: Too many Rt requests when scenario first loads
Because Rt fetching is centralized in the FacilitiesContext, we should only be issuing one request per facility to initialize the scenario data. In practice, with 7 facilities in my scenario, I am observing anywhere between 30 and 90 requests! In larger scenarios the performance hit from all these requests can be enough to lock up or crash my browser (the number appears to scale by a large multiplier, possibly even exponentially with the number of facilities, and it seems to be exacerbated by requests failing rapidly as happens when you first create new facilities in bulk from reference data)

Something must be wrong in our state management around  Rt requests (we should not be issuing Rt requests for a facility while one is pending and awaiting response, nor after one has succeeded or failed for that facility, unless new data has been input by the user. Which is not what happens on initial page load).

macfarlandian:
## Description of the change

This fixes another of the performance issues that was exacerbated by shadow data, by setting a loading state for pending Rt requests. 

We had this (admittedly obscure) convention in the initial Rt implementation of explicitly mapping a facility ID to `undefined` while a request to the `calculate_rt` cloud function was pending, which let us only issue requests for facility IDs that were completely unmapped when initializing a facility. All the code that supported and built on that assumption was still in place but at some point the actual mapping step must have gotten lost in translation when we moved over to the `FacilitiesContext`, which was causing large numbers of duplicate requests in [the useEffect function that initializes Rt](https://github.com/Recidiviz/covid19-dashboard/blob/ian/688-rt-requests/src/facilities-context/FacilitiesContext.tsx#L320) (scaling in multiples of facility count, because each time one facility returned, an additional request would be issued for each pending facility). This PR restores that convention.

## Type of change

## Related issues

Closes #688 

### Development

### Code review

### DIFF
--- a/src/facilities-context/actions/index.tsx
+++ b/src/facilities-context/actions/index.tsx
@@ -47,10 +47,15 @@ export type RECEIVE_FACILITIES_ACTION = {
   payload: FacilityMapping;
 };
 
-export type RT_DATA_ACTION = {
-  type: typeof UPDATE_FACILITY_RT_DATA;
-  payload: RtDataMapping;
-};
+export type RT_DATA_ACTION =
+  | {
+      type: typeof UPDATE_FACILITY_RT_DATA;
+      payload: RtDataMapping;
+    }
+  | {
+      type: typeof REQUEST_RT_DATA;
+      payload: Facility["id"];
+    };
 
 export type FACILITY_ID_ACTION = {
   type: typeof SELECT_FACILITY | typeof REMOVE_FACILITY;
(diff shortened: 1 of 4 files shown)


macfarlandian:
@jessex I spot-checked the scenarios I have access to, to make sure they were being initialized with the expected Rt data, as well as verifying that they weren't issuing these extra requests.

PR #702, opened 2020-09-04
Duplication fails for large scenarios

### LINKED ISSUE #701: Duplication fails for large scenarios
**Describe the bug**
Two issues here, closely related. The first is that duplicating a large scenario may fail for a known reason (the 500-operation limit on Firestore write batches). The second is that, upon any failure in the `duplicateScenario` function, the error is not handled and no message is displayed. Instead the application erroneously assumes the user has no scenarios, which leads to the creation of an additional baseline scenario. 

**To Reproduce**
1. You need access to a large scenario (e.g. including several hundred `modelVersions` documents across multiple facilities), either owned or shared
2. Duplicate that scenario via the scenario library or the green read-only banner
3. You should now see a new empty baseline scenario, and probably an error in the browser console about the batch size being exceeded

**Expected behavior**
The scenario and all its facilities are duplicated successfully. If an error does happen, it should be displayed in an Error Toast and no new baseline scenario should be created.

**Additional context**
The new scenario creation results from `duplicateScenario` returning `undefined` when an error occurs, and that value being fed into the `ScenarioContext`, which the application interprets as "this user has no scenarios". We should throw that error instead of catching it and then handle it with `useRejectionToast`. 

We encountered the same problem in our data ingest functions on the backend and had to create a [wrapper class](https://github.com/Recidiviz/covid19-dashboard/blob/master/backend/python-functions/data_ingest.py#L24) to manage them; unfortunately that is in Python so it can't be directly reused here. It should be possible to either reimplement this in JavaScript or find a library on npm that has solved this problem, though. Moving this operation to a python cloud function would also be possible but is probably more work and less desirable then fixing it on the client. (other client-side bulk operations might benefit from this functionality also)

macfarlandian:
## Description of the change

Adds a wrapper around Firestore batched writes to manage the batch size limit. This is similar to but more extensive than the Python version implemented in #686 ... there were some existing libraries on npm that partially implemented this but none of them had a solution for the server-side transformations contributing to the batch size, so they did not prevent the errors we were seeing. All of our unbounded client-side batch writes (duplication and deletion of scenarios and facilities) use this new wrapper and I verified that all still work (and where applicable, now succeed where they used to fail). 

Also added some additional error handling for duplication to prevent the additional baseline scenario creation that accompanied batch write failures.

## Type of change

## Related issues

Closes #701 

### Development

### Code review

### DIFF
--- a/src/database/type-transforms.tsx
+++ b/src/database/type-transforms.tsx
@@ -15,8 +15,16 @@ import {
 import { referenceFacilitiesProp } from ".";
 import { FacilityDocUpdate } from "./types";
 
-const timestampToDate = (timestamp: firebase.firestore.Timestamp): Date => {
-  return timestamp.toDate();
+const timestampToDate = (
+  timestamp: firebase.firestore.Timestamp | null,
+): Date => {
+  if (timestamp) {
+    return timestamp.toDate();
+  } else {
+    // when Firestore writes are pending, server timestamps may be null;
+    // we can substitute the local now in the meantime
+    return new Date();
+  }
 };
 
 const buildPlannedRelease = (plannedReleaseData: any): PlannedRelease => {
@@ -98,11 +106,7 @@ export const buildScenario = (
   let scenario: Scenario = documentData;
   scenario.id = document.id;
   scenario.createdAt = timestampToDate(documentData.createdAt);
-  scenario.updatedAt = timestampToDate(
-    // when Firestore writes are pending, updatedAt may be null since it's a server timestamp;
-    // we can substitute the local now in the meantime
-    documentData.updatedAt || firebase.firestore.Timestamp.now(),
-  );
+  scenario.updatedAt = timestampToDate(documentData.updatedAt);
   scenario.referenceDataObservedAt = documentData.referenceDataObservedAt
     ? timestampToDate(documentData.referenceDataObservedAt)
     : undefined;
--- a/src/database/type-transforms.tsx
+++ b/src/database/type-transforms.tsx
@@ -15,16 +15,8 @@ import {
 import { referenceFacilitiesProp } from ".";
 import { FacilityDocUpdate } from "./types";
 
-const timestampToDate = (
-  timestamp: firebase.firestore.Timestamp | null,
-): Date => {
-  if (timestamp) {
-    return timestamp.toDate();
-  } else {
-    // when Firestore writes are pending, server timestamps may be null;
-    // we can substitute the local now in the meantime
-    return new Date();
-  }
+const timestampToDate = (timestamp: firebase.firestore.Timestamp): Date => {
+  return timestamp.toDate();
 };
 
 const buildPlannedRelease = (plannedReleaseData: any): PlannedRelease => {
@@ -106,7 +98,11 @@ export const buildScenario = (
   let scenario: Scenario = documentData;
   scenario.id = document.id;
   scenario.createdAt = timestampToDate(documentData.createdAt);
-  scenario.updatedAt = timestampToDate(documentData.updatedAt);
+  scenario.updatedAt = timestampToDate(
+    // when Firestore writes are pending, updatedAt may be null since it's a server timestamp;
+    // we can substitute the local now in the meantime
+    documentData.updatedAt || firebase.firestore.Timestamp.now(),
+  );
   scenario.referenceDataObservedAt = documentData.referenceDataObservedAt
     ? timestampToDate(documentData.referenceDataObservedAt)
     : undefined;
--- a/src/database/utils.tsx
+++ b/src/database/utils.tsx
@@ -157,6 +157,16 @@ export class BatchWriter {
             case "UPDATE":
               writeBatch.update(operation.params.ref, operation.params.data);
               break;
+            default:
+              console.error(
+                // type assertion here because operation is technically a never
+                // in the present implementation; this is mainly intended as a guardrail
+                // for future implementations of additional write methods (e.g. CREATE)
+                `Operation type ${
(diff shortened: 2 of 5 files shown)


src/database/type-transforms.tsx
```diff
@@ -15,8 +15,16 @@ import {
...
-const timestampToDate = (timestamp: firebase.firestore.Timestamp): Date => {
-  return timestamp.toDate();
+const timestampToDate = (
+  timestamp: firebase.firestore.Timestamp | null,
+): Date => {
+  if (timestamp) {
+    return timestamp.toDate();
+  } else {
+    // when Firestore writes are pending, server timestamps may be null;
+    // we can substitute the local now in the meantime
```
jessex, line 25: How do we know that this isn't going to be persisted?

src/database/type-transforms.tsx
```diff
@@ -15,8 +15,16 @@ import {
...
+    return new Date();
```
jessex, line 26: I see that in the call to this below for `scenario.updatedAt`, it used to substitute `firebase.firestore.Timestamp.now()` which _looks_ like a different type than just `new Date()`. Am I mistaken? If not, why is this type change needed?

src/database/utils.tsx
```diff
@@ -25,3 +26,141 @@ export const prepareFromStorage = (state: object): EpidemicModelPersistent => {
...
+  }
+
+  private startBatch() {
+    const newBatch: BatchOperation[] = [];
+    this.currentBatch = newBatch;
+    this.operationBatches.push(newBatch);
+  }
+
+  private trackWrite() {
+    this.currentBatchSize += 1 + this.pendingOpsCount;
```
jessex, line 81: Could this not lead to double counting pending operations? For example, if I have 2 pending ops, and I pass through some code flow that leads to consecutive calls to `trackWrite`, wouldn't those pending ops get counted for each invocation of `trackWrite`? What's preventing that from happening?

I could also envision the answer being, "Yes, this is possible, but it's okay because all it would do is lead to us being more conservative in chunking up operations."

src/database/utils.tsx
```diff
@@ -25,3 +26,141 @@ export const prepareFromStorage = (state: object): EpidemicModelPersistent => {
...
+    // mimic Firestore interface, allows for chained method calls
+    return this;
+  }
+
+  async commit() {
+    await Promise.all(
+      this.operationBatches.map((operationBatch) => {
+        const writeBatch = this.client.batch();
+        operationBatch.forEach((operation) => {
+          switch (operation.type) {
```
jessex, line 146: I wonder if we should have a default that logs the operation type we attempted as an error, just so we know if a future developer tries to use CREATE or some other unsupported operation?

jessex (commented):
Just some questions from me. Great work, @macfarlandian!

src/database/type-transforms.tsx
macfarlandian: rather than requiring all callers to catch nulls on their own and replace them with the local now in `Timestamp` form so that we can immediately convert it to a `Date`, we are now catching the nulls inside this function. When that happens there is nothing to convert, so we can just return a `Date` directly

src/database/utils.tsx
macfarlandian: they wouldn't get double-counted because `trackWrite` resets that counter to 0. But they could get counted out of order in that situation ... there is an expectation here that you will use this class by adding whatever server transforms you need and then immediately calling the "write" method that will trigger them. If you didn't follow that order — e.g. you queued up 500 objects that each rack up one or two `pendingOps` and then called `BatchWrite.update` 500 times — then `commit` would throw an error. 

It's an imperfect heuristic ... the truly officially Recommended way to do this is to make a Cloud Function and do more parallel writes instead of serial writes (I think the batching functionality is somewhat intentionally limited to encourage you to do this instead), but that seemed like it would be a bigger architectural change. I was kind of hoping we could get away with not doing it? But if you think this approach is too brittle I couldn't exactly argue with you ...

src/database/utils.tsx
macfarlandian: i left that out because there's not really any way to do an unsupported operation currently (since it involves only private members), but I guess if someone added a partial implementation (added the method without updating `commit`) that would alert them that they had made a mistake .... seems worth it then

src/database/type-transforms.tsx
macfarlandian: I will admit that the provenance of these nulls is still mysterious to me ... I haven't really been able figure out why we are getting them. It wasn't such a big deal when they were only happening in `updatedAt` (because we always update that when saving so the previous value wouldn't get persisted anyway) but if it can happen in `createdAt` (it happened to me during one of these duplication operations, that's why I made this change) then it could potentially be persisted. I guess I didn't really think this through all the way? I will try to discern the cause of the null `createdAt` and see if there is a less broad solution

src/database/utils.tsx
jessex: Given prioritization and the relatively lowish likelihood of this drifting away from us, I think we can keep this as-is. But we should comment on `trackWrite` in big bold letters what you said here about the expected pattern.

src/database/type-transforms.tsx
jessex: Got it. Let me know if you plan on making another change here or if you're ready for another review towards approval.

macfarlandian: yeah i am still poking around at the null issue, I expect there will be another change here

jessex (approved):
Thanks for the thorough debugging and testing here, @macfarlandian!

PR #707, opened 2020-10-30
Handle current population less than historical cases in projections

### LINKED ISSUE #706: curve model makes negative projections when all cases are recovered
Not posting the screenshots since it is user data, but to describe the conditions: 

- there are cases in various age brackets; for each of these brackets the number of recovered cases is equal to the total number of cases (i.e. there are no active cases)
- in the staff bracket there are also cases, but the number recovered is less than the total 
- the population turnover factor is less than 1.0; I suspect this may be involved as well 

the curve function should show curves going negative. when seen in the weekly snapshot report you should also see some negative numbers in the weekly projections

*Expected behavior*: 
Numbers in the curve function should never go below zero

macfarlandian:
## Description of the change

Addresses an issue where the projection would predict negative populations in some compartments if the current population for a bracket is less than its cumulative number of cases. This is not necessarily a user error, since people can exit a facility after being infected. It was observed in the wild in facilities with very small numbers (single digits) in individual age brackets.

Clamping negative numbers to zero in the initial distribution of current cases and population across departments prevents any subsequent projections from dipping below zero. However, we also have a competing assumption that the sum of all compartments will always equal the total population. Both conditions can't be true in this case; they will be off by the difference between cumulative cases and current population. My solution to this was to adjust the model's total population figure by this same amount. There is some precedent for this, since the model's total population already includes some people who are presumably not physically in the facility anymore (e.g., fatalities, people moved to hospitals), but I definitely need confirmation that @justkunz has no objections to this!

The test data replicates (with falsified values) the conditions observed in the wild that led to this bug. I also used the `/inspect-model` UI to both reproduce the bug and verify the fix. I also don't see any obvious regressions in any of my available test data.

## Type of change

## Related issues

Closes #706 

### Development

### Code review

### DIFF
--- a/src/infection-model/__tests__/curves.test.tsx
+++ b/src/infection-model/__tests__/curves.test.tsx
@@ -108,6 +108,30 @@ describe("curveInputsWithRt function", () => {
   });
 });
 
+/**
+ * Verifies that the sum of all compartments equals the total population
+ */
+function validateCompartmentsChecksum({
+  totalPopulationByDay,
+  projectionGrid,
+}: Pick<
+  ReturnType<typeof calculateAllCurves>,
+  "projectionGrid" | "totalPopulationByDay"
+>) {
+  totalPopulationByDay.forEach((totalPop, day) => {
+    let sumOfCompartments = 0;
+
+    range(projectionGrid.shape[0]).map((compartment) =>
+      range(projectionGrid.shape[2]).map(
+        (bracket) =>
+          (sumOfCompartments += projectionGrid.get(compartment, day, bracket)),
+      ),
+    );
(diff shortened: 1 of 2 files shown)


jessex (approved):
I think this makes sense and I appreciate the thorough approach to testing! If Justine approves, let's get it into prod!

macfarlandian:
also worth mentioning that I asked Zak to test this in the preview deployment and he confirmed that it fixed the bug he reported

macfarlandian:
ping @justkunz

justkunz (approved):
The changes look great! The comments make it really easy to follow the logic and the new test is great 🎉 🚀

PR #714, opened 2021-04-01
Resolve security vulnerability in is-svg

### LINKED ISSUE #713: Resolve 2 security vulnerabilities
As of this writing there are two Dependabot alerts on this repo. I believe that both can be safely dismissed. 

The [alert for `y18n`](https://github.com/Recidiviz/covid19-dashboard/security/dependabot/yarn.lock/y18n/open) has evidently been updated to reflect that the versions we have installed are no longer considered vulnerable. The alert has not been dismissed automatically but there seems to be no reason not to close it at this point.

The [alert for `is-svg`](https://github.com/Recidiviz/covid19-dashboard/security/dependabot/yarn.lock/is-svg/open) concerns a transitive Gatsby dependency; it [doesn't sound like](https://github.com/cssnano/cssnano/issues/1019#issuecomment-802182732) there will be a security patch for the line that Gatsby is on (even if we were to upgrade to v3.x). However, according to the maintainers this package is only part of the build tooling and is not actually in use in the bundled application; since there is no actual vulnerability in our deployment I believe this is safe to dismiss as well.

paging @jessex to ratify

macfarlandian:
## Description of the change

Versions of `is-svg` (a transitive dependency of Gatsby via Webpack) below 4.2.2 have [a vulnerability](https://github.com/advisories/GHSA-7r28-3m3f-r2pr); because this package is only part of our build tooling, the actual risk here is negligible, but forcing an upgrade is trivial so we might as well do it. (This generates a yarn warning, but the only breaking change introduced by this version bump is a lack of support for Node 4.x, which we already don't support.)

Because the vulnerable package is not included in any runtime bundle, no deployment will be necessary to complete this patch.

## Type of change

## Related issues

Closes #713 

### Development

### Code review

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,5 +168,8 @@
     "jest-styled-components": "^7.0.2",
     "react-test-renderer": "^16.13.1",
     "ts-jest": "^25.4.0"
+  },
+  "resolutions": {
+    "is-svg": "^4.2.2"
   }
 }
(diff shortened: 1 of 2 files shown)


jessex (approved):
Sounds great to me. Thanks, @macfarlandian!

PR #730, opened 2021-06-15
force upgrade three low level dependencies

### LINKED ISSUE #728: Resolve ws vulnerability
The `ws` package, a transitive dependency of Gatsby and Jest, has a moderate severity vulnerability as documented [here](https://github.com/Recidiviz/covid19-dashboard/security/dependabot/yarn.lock/ws/open). The package should be updated to a non-vulnerable version. (The patch has been back-ported to both the 6.x and 7.x lines, both of which are in our dependency graph, so this shouldn't require any unusual measures.) 

**Additional context**
FYI @jessex I don't believe there is any live vulnerability, because we don't use any websockets in the application, so I don't think it's urgent that we deploy anything to resolve this.

### LINKED ISSUE #729: Upgrade normalize-url and glob-parent dependencies
There is an active security vulnerability in normalize-url relied upon by the app. It is a HIGH severity and was first noted 1 business day ago, so we want this resolved by Friday of this week. There is also an active vulnerability for glob-parent which is a MODERATE serverity, but if we can resolve both at the same time then all the better.

**Additional context**
[Read more here](https://github.com/Recidiviz/covid19-dashboard/security/dependabot)

macfarlandian:
## Description of the change

Resolves three transitive Gatsby dependencies to non-vulnerable versions. Necessary again due to our older version of Gatsby, which still does not really seem to be worth updating. 

Mostly these are just part of dev tooling so there's not much risk associated with them anyway, but that's not necessarily the case for `normalize-url`, so we probably do want to do a release to complete this fix. Smoke tests reveal no ill effects.

## Type of change

## Related issues

Closes #729, closes #728 

### Development

### Code review

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,6 +171,9 @@
   },
   "resolutions": {
     "@firebase/util": "^0.3.4",
-    "is-svg": "^4.2.2"
+    "glob-parent": "^5.1.2",
+    "is-svg": "^4.2.2",
+    "normalize-url": "^4.5.1",
+    "ws": "^7.4.6"
   }
 }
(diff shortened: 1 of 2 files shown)


jessex (approved):
Thanks so much for this, as always, @macfarlandian 💓

